PR #11, opened 2025-09-23
Python agent

Coykto:


mkf-simpson:
> it's obvious that this particular agent is supposed to be an expert only in web development which looks to me a little bit too restrictive or specific. do we have plans to introduce other "roles" like data scientist, devops, etc?

The short answer: we don't know. The long answer: there is a trade-off between how narrow we want to have our agents and their overall number. In the current state, LLMs have limited "attention" and if we overload them with too many tools, they will mess everything up. Where this threshold is we don't know yet, but we will figure out once we hit the ceiling.

> ps: not sure what are the use cases for that agent but I assumed that it should act as a "copilot" rather than an independent "employee" so my points above are based on this assumption.

Wrong assumption ¯\_(ツ)_/¯ The ultimate goal is to create a system of agents that actually can do things

ppavlov-provectus:
Thanks for clarification @mkf-simpson. Feel free to ignore my comments then but I still want you guys to take into consideration "standardization" aspect of the agent's output. Since you're planning to make those agents to be full-time developers (which is great and I really like your ambition) I think it's going to be a good thing to outline some of the coding/tooling standards for the sake of maintainability and readability/navigation for meat-based developers :)

Probably I'll be more helpful after initial tests and performance benchmarking because rn I'm kinda out of the track and don't even know what to expect from such independent agents.

PR #53, opened 2025-10-08
Use technology experts when drafting tech specification

Coykto:
I noticed that even though we have technology expert agents, tech spec creation does not utilize them and produces less than ideal code for future implementation.


src/utils/logger.js
```diff
@@ -70,29 +70,50 @@ function showHeader(asciiArt, subtitle) {
...
+    console.log('');
+    console.group(style.bold('  Summary:'));
+  }
 
   summaryItems.forEach(([label, value]) => {
-    console.log(
-      `${style.success('•')} ${label}: ${style.bold(value.toString())}`
-    );
+    const prefix = options?.dryRun ? style.dim('○') : style.success('•');
+    const labelText = options?.dryRun ? `Would ${label.toLowerCase()}` : label;
```
workshur:
```suggestion
    const labelText = label;
```

`What would happen:` as a header already ok.   

```
  ○ Would directories created: 0
  ○ Would files installed: 19
  ○ Would existing items preserved: 12
  ○ Would migrations applied: 1
```
Sounds incorrect

Coykto: Fixed phrasing: 
```
  - ○ Would create directories: 5
  - ○ Would install files: 26
  - ○ Would preserve existing items: 0
  - ○ Would apply migrations: 1
```

PR #55, opened 2025-10-23
feat: Multi-Agent Installer + GitHub Copilot Support

AndreyNenashev:
This PR introduces full multi-agent support to the AWOS framework, transforming the installer into an agent-agnostic system capable of configuring multiple AI assistants (starting with Claude and GitHub Copilot).

The installer now uses an explicit --agent flag to determine which agent to set up, with centralized validation and configuration logic.

1. GitHub Copilot Integration: Added full Copilot support to the installer, expanding AWOS into the Copilot ecosystem while maintaining backward compatibility.
2. Scalable Foundation: Established a clear separation between core files and agent-specific directories (claude/, copilot/, etc.), paving the way for additional agent integrations.

agent flag example:
```
npx @provectusinc/awos

  ✗ Error: --agent flag is required
  • Usage: npx @provectusinc/awos --agent <agent-name>
  • Example: npx @provectusinc/awos --agent claude
```

🧠 **Ecosystem Note — GitHub Copilot**
Unlike Claude Code, GitHub Copilot cannot read or reference external files directly from prompts.
Because of this limitation, all AWOS commands and guidelines must be fully copied into Copilot prompts rather than referenced by path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,38 @@ To get started with **`awos`**, you'll need a couple of things set up first:
 
 ### Step 1: Install `awos`
 
-First, open your terminal, create a new directory for your project, and run this single command. It will set up everything you need.
+First, open your terminal, create a new directory for your project, and run the installer with the required `--agent` flag. This will set up everything you need.
+
+**For Claude Code:**
 
 ```sh
-npx @provectusinc/awos
+npx @provectusinc/awos --agent claude
 ```
+
+**For GitHub Copilot:**
+
+```sh
+npx @provectusinc/awos --agent copilot
+```
+
+**Note:** The `--agent` flag is mandatory and specifies which AI agent you'll be using. Currently supported agents are `claude` and `copilot`.
+
+### CLI Flags Reference
+
+**Required Flags:**
+
+- `--agent <agent-name>` - Specifies which AI agent to configure (required)
+  - Supported values: `claude`, `copilot`
+  - Example: `npx @provectusinc/awos --agent claude`
+
+**Optional Flags:**
+
+- `--force-overwrite` - Overwrites all files, including customizations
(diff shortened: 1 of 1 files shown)


AndreyNenashev:
I didn't properly test it yet so let's keep in draft state for now

AndreyNenashev:
> @AndreyNenashev can we reuse existing prompts?

@workshur Those prompts are actually full copies of original commands. Copilot cannot operate by `referencing` like claude but only by what we have in prompt content. The only prompt I changed is `implement` command.
Main difference is in the `implement` - switching instructions and role from "delegate to agent" to "prepare/formulate instructions for expert" that will require user input and confirmation.

```
### Step 3: Prepare Implementation Instructions

- **CRITICAL RULE:** You are **strictly prohibited** from writing, editing, or modifying any production code, configuration files, or database schemas yourself. Your only role is to prepare instructions for the appropriate expert.

1.  **Formulate Implementation Instructions:** Construct a clear and detailed instruction for the specialized implementation expert (Python Expert or React Expert). This instruction MUST include:
    - The full context from the three files you just loaded.
    - The specific task description that needs to be implemented.
    - Clear instructions on what code to write or what files to modify.
    - A definition of success (e.g., "The task is done when the new migration file is created and passes linting.").

2.  **Prepare Expert Invocation:** Based on the technology stack identified, prepare the appropriate expert invocation:
    - For Python/FastAPI backend tasks: Prepare invocation for `@awos.python-expert`
    - For React/Frontend tasks: Prepare invocation for `@awos.react-expert`

3.  **Output Implementation-Ready Prompt:** Present a complete, copy-paste ready prompt for the user to invoke the appropriate expert. Format it as follows:
```

PR #109, opened 2026-04-09
feat: Feature Testing & Regression slice, QA audit command, and verification hardening

FlySpot:
## Summary

- **`/awos:tasks`** — adds a mandatory **Feature Testing & Regression** final slice to every task list. The slice runs `testing-expert` to generate acceptance-level tests against `functional-spec.md` acceptance criteria after all implementation slices are complete. Also adds a **Cleanup** sub-task after each implementation slice to delete ephemeral artifacts (screenshots, videos, e2e scripts) produced during verification. Artifact cleanup is skipped for the Feature Testing & Regression slice — those are retained for the regression suite.
- **`/awos:hire`** — adds a **QA Complement Rule**: after hiring a primary tech agent, always propose `testing-expert` alongside it. Frontend stacks also get `playwright` MCP suggested.
- **`/awos:verify`** — adds a **CONSTRAINTS** section enforcing that Step 3 verification is mandatory. Defines a fallback chain: browser MCP → curl/shell → logs/database → explicit user approval via `AskUserQuestion`. Agents can no longer silently skip verification.
- **`templates/qa-context-template.md`** — starter template for the test registry at `context/qa/list-of-tests.md`.
- **Audit dimensions** — `spec-driven-development.md` (SDD-07) updated to recognize the Feature Testing & Regression final-slice model alongside the legacy per-slice model.
- **Installer** — `src/config/setup-config.js` now copies `plugins/awos/agents/` → `.claude/agents/` during `npx @provectusinc/awos`, matching the documentation in `src/CLAUDE.md`.

## What is NOT in this PR

- **`/awos:regression`** command — lives in `feat/regression` (PR #114). The regression sub-task inside the Feature Testing & Regression slice is commented out pending that merge.
- **`testing-expert` agent** — hired via `awos-recruitment` registry, not bundled in the core package.
- **`/awos:qa`** slash command — the Claude Code wrapper was removed; `commands/qa.md` is preserved as a future plugin reference only.

## File map

```
commands/tasks.md       ← Feature Testing & Regression slice + Cleanup sub-task
commands/hire.md        ← QA Complement Rule
commands/verify.md      ← mandatory Step 3 + fallback chain constraints
templates/
  qa-context-template.md              ← test registry scaffold

plugins/awos/skills/ai-readiness-audit/dimensions/
  spec-driven-development.md          ← SDD-07: dual-model QA coverage
  end-to-end-delivery.md              ← updated references
  software-best-practices.md          ← updated references

src/config/setup-config.js            ← agents copy operation added
src/CLAUDE.md                         ← installer docs
```

## Test plan

- [ ] Run `/awos:tasks` on a spec — confirm the Feature Testing & Regression slice appears as the last slice with correct wording and agent assignments
- [ ] Confirm Cleanup sub-task is present after each implementation slice's verify step
- [ ] Run `/awos:hire` with a frontend tech — confirm `testing-expert` and `playwright` MCP appear in proposals
- [ ] Run `/awos:verify` without browser MCP — confirm fallback chain is followed and agent cannot silently skip

## Summary by CodeRabbit

* **New Features**
  * Complete QA audit workflow for end-to-end testing verification
  * Mandatory verification step with fallback validation methods
  * Automatic cleanup of verification artifacts

* **Documentation**
  * New test registry template
  * Enhanced QA agent discovery and pairing guidelines
  * Updated spec-driven development QA coverage model

[Review Change Stack](https://app.coderabbit.ai/change-stack/provectus/awos/pull/109?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


commands/tasks.md
```diff
@@ -67,6 +67,23 @@ Follow this process precisely.
         - Tech stack identified in technical-considerations.md
       - Append the subagent assignment using format: `**[Agent: agent-name]**` at the end of the sub-task description
       - Use `general-purpose` agent when no specialist clearly matches the task — but **track these assignments** for the Recommendations table
+  3b. **Generate paired test tasks for each slice (REQUIRED):**
```
BulatKha: Weird numbering mixing thought process syncing

workshur: What if I don't want to generate any tests?

FlySpot: > Weird numbering mixing thought process syncing

Fixed.  Thanks.

FlySpot: > What if I don't want to generate any tests?

Added as an option to skip test creation if the user asks about it.

commands/tasks.md
```diff
@@ -58,19 +58,21 @@ Follow this process precisely.
...
+  5.  **Generate paired test tasks for each slice (skip only if user explicitly opts out):** - After defining the implementation sub-tasks for a slice, invoke the `testing-expert` agent in **planning mode**. - Pass it: the current slice's implementation sub-task description, `functional-spec.md`, and `technical-considerations.md`. - The `testing-expert` will return a list of test sub-tasks covering the applicable pyramid layers (unit, integration, e2e, contract) with both positive and negative cases. - Insert those test sub-tasks as children of the same slice, after the implementation sub-tasks. - If the user has said they do not want tests generated, skip this step entirely and omit all test sub-tasks. - **CRITICAL — Slice Completion Rule:** A slice parent task is only `[x]` when ALL its sub-tasks — both implementation AND all test sub-tasks — are `[x]`. This is enforced by `/awos:implement`'s existing sub-task completion logic. - Example result for a slice:
```
kmakarychev-dev: You mention here a `testing-expert` agent, but it's not registered as an agent (the md file is just in `commands` directory for some reason). Also, I doubt that "invoke in planning mode" can actually work

commands/tasks.md
```diff
@@ -58,19 +58,21 @@ Follow this process precisely.
...
+      ` - [ ] **Slice 1: User authentication**
+      - [ ] Implement JWT token generation **[Agent: python-expert]**
```
kmakarychev-dev: This is a bad example, actually. It is a general purpose command that must produce a comprehensive task list. The example should not contain only qa-related tasks, otherwise it might be biased towards only testing tasks.

commands/testing-expert.md
```diff
@@ -0,0 +1,151 @@
+---
+description: Plans and executes layered test suites (unit, integration, e2e, contract) for vertical slices. Enforces negative tests and RED validation.
```
kmakarychev-dev: It doesn't sound like a command

kmakarychev-dev (changes requested):

commands/testing-expert.md
kmakarychev-dev: Also, it's kinda weird to have a prescriptive testing agent: I would assume that it should have specific skills and instructions for each projects. Moreover, in monorepos, there can be bunch of different testing agents: for UI, for different backends, etc.

We have to find a way of how to provide a clear process, maybe, help with generating those agents (e.g. through the recruitment service).

commands/tasks.md
FlySpot: You're right on both counts. 
- 'testing-expert' belongs in plugins/awos/agents/, not commands/ — that's where dimension-auditor lives and where the Agent tool can resolve subagent_type. 
- The "planning mode" instruction is also broken as written: it relies on the agent's internal mode detection (no implementation code in context = planning mode), but the invocation mechanism isn't specified. 

**Plan**: move the file to pugins/awos/agents/testing-expert.md and rewrite step 5 to use an explicit Agent tool call.

commands/tasks.md
FlySpot: Agreed, the example is misleading. A slice example should show implementation sub-tasks alongside test sub-tasks to reflect what a real balanced slice looks like. Will replace it with one that includes both.

commands/testing-expert.md
FlySpot: Both points are valid. 

1. It's not a command — it's an agent persona definition and should live in plugins/awos/agents/. 
2. On the prescriptive agent concern: fully agree. A hardcoded generic testing-expert breaks down in monorepos and projects with diverse stacks. 

Thoughts: 
- plugins/awos/agents/testing-expert.md stays as a generic fallback, 
- /awos:hire — installs a project-specific testing agent into .claude/agents/ via the recruitment service. Project agents take priority over plugin agents in Claude Code's resolution order, so the override is automatic. tasks.md and qa.md would then call whichever testing agent is available.

FlySpot:
Just to sum-up what was changed in the current flow:

## Added

- Added a new optional `/awos:qa` command for full QA audits.
- Added `plugins/awos/agents/testing-expert.md` — the QA subagent definition.
- Added `templates/qa-context-template.md`.
- Added `templates/regression-suite-template.md`.

---

## Changed

### `/awos:tasks` — each slice can now include explicit test tasks, not only implementation tasks.

**Before**

```text
Slice 1
  [ ] impl sub-task
  [ ] verify: open browser
```

**After**

```text
Slice 1
  [ ] impl sub-task
  [ ] Unit: positive case
  [ ] Unit: negative case
  [ ] Integration: positive case
  [ ] Integration: negative case
  [ ] verify: open browser
```

### `/awos:qa` — delegates to `testing-expert` instead of writing tests inline

**Before**

```text
Step 6: for each gap — write the missing test inline,
        run RED validation,
        update list-of-tests.md
```

**After**

```text
Step 6: for each gap — invoke Task tool with subagent_type: "testing-expert",
        passing gap description, spec files, and implementation code.
        testing-expert handles RED validation and registry update.
```

### Installer — `testing-expert` is now deployed to user projects

`src/config/setup-config.js` now copies `plugins/awos/agents/testing-expert.md` → `.claude/agents/testing-expert.md` during `npx @provectusinc/awos`.

---

## Workflow

The full development flow after this change is:

```text
/awos:spec → /awos:tech → /awos:hire
                         │
                         ▼
                    /awos:tasks
          ┌──────────────────────────────────┐
          │ For each slice:                  │
          │                                  │
          │ 1. Create impl sub-tasks         │
          │                                  │
          │ 2. Invoke testing-expert         │
          │    passing: slice description,   │
          │    functional-spec.md,           │
          │    technical-considerations.md,  │
          │    NO implementation code        │
          │                                  │
          │ 3. testing-expert sees no code   │
          │    → returns test task list      │
          │      (unit/integration/e2e/      │
          │       contract, +/- cases)       │
          │                                  │
          │ 4. Test tasks inserted into      │
          │    the slice, before Verify      │
          └──────────────────────────────────┘
                         │
                         ▼
                 /awos:implement
          ┌──────────────────────────────────┐
          │ impl task  → domain agent        │
          │                                  │
          │ test task  → testing-expert      │
          │    passing: task description,    │
          │    spec files,                   │
          │    implementation code  ← key    │
          │                                  │
          │ testing-expert sees real code    │
          │ → writes tests + RED validation  │
          │ → returns completion signal or   │
          │   incomplete/blocked on gap      │
          └──────────────────────────────────┘
                         │
                         ▼
                    /awos:qa (optional)
          ┌──────────────────────────────────┐
          │ audit existing test coverage     │
          │ detect gaps by pyramid layer     │
          │ delegate gap tests to            │
          │   testing-expert via Task tool   │
          │ sync regression suite            │
          │ produce qa-report-YYYY-MM-DD.md  │
          └──────────────────────────────────┘
```

---

## Clarified

`testing-expert` is a single agent definition with context-driven behavior, not separate invocation modes.

| Context received | Behavior |
|---|---|
| spec files, no implementation code | returns a list of test task descriptions — no code written |
| spec files + real implementation code | writes actual test code, runs RED validation, confirms GREEN |

- `testing-expert` is not a slash command — it lives in `plugins/awos/agents/` and is invoked exclusively via the `Task` tool with `subagent_type: "testing-expert"`.
- `testing-expert` returns a **completion signal** (no gaps) or **incomplete/blocked status** (gap found) — the caller decides what to do with it, not the agent.

---

## Test layer selection

`testing-expert` selects test layers per slice based on the slice description and acceptance criteria from `functional-spec.md`, plus `technical-considerations.md`.

```text
Slice implementation sub-task description
+ acceptance criteria from functional-spec.md
       │
       ├──► pure logic, no external deps?  → Unit
       ├──► service-to-service / DB calls? → Integration
       ├──► full user flow via UI or API?  → E2E
       └──► API schema / interface check?  → Contract
```

> Not every slice needs all four layers. Apply judgment.

---

## Notes

- Test layer selection is a judgment call, not a fixed algorithm.
- If `/awos:hire` has installed a project-specific testing agent (e.g. `react-testing`, `python-testing`), it takes precedence over `testing-expert`.
- The previous `tasks.md` example was misleading because it implied every slice always needs all four layers. In practice, a slice may only require unit + integration coverage.

commands/hire.md
```diff
@@ -89,6 +89,29 @@ Follow this process precisely.
...
+
+Complementary pairs reference:
+
+- React / Vue / Angular → `testing-expert` + `playwright` MCP
+- FastAPI / Django / Flask → `testing-expert` + `pytest-best-practices` skill
+  _(if `pytest-best-practices` not found in registry — `testing-expert` covers this)_
+- Any Python backend → `testing-expert` + `pytest-best-practices` skill
+  _(if `pytest-best-practices` not found in registry — `testing-expert` covers this)_
+- Any TypeScript/Node backend → `testing-expert`
+- Terraform / IaC → `testing-expert` (for infra validation)
```
dustyo-O: any proofs that is going to run better in comparison with devops-infra?

commands/hire.md
```diff
@@ -89,6 +89,29 @@ Follow this process precisely.
...
+primary tech agent — unless a more specific testing agent (e.g. `react-testing`,
+`rust-tester`) is also found, in which case prefer the specific one.
+
+Also: if the tech stack includes any frontend framework (React, Vue, Angular,
+Svelte, etc.), always include the `playwright` MCP in the proposal — it enables
+browser-based E2E testing via the testing-expert agent.
+
+Complementary pairs reference:
+
+- React / Vue / Angular → `testing-expert` + `playwright` MCP
```
dustyo-O: Im confused that we reference testing-expert for every example, looks that claude could be biased to testing-expert even if more focused testing agents would be available.

I suggest providing testing expert as a fallback in each example, even though there is no react-testing expert yet available

commands/hire.md
```diff
@@ -89,6 +89,29 @@ Follow this process precisely.
...
+- FastAPI / Django / Flask → `testing-expert` + `pytest-best-practices` skill
+  _(if `pytest-best-practices` not found in registry — `testing-expert` covers this)_
+- Any Python backend → `testing-expert` + `pytest-best-practices` skill
```
dustyo-O: isn't it completely the same with previous?

commands/qa.md
```diff
@@ -0,0 +1,188 @@
+---
+description: Optional full-audit QA command — coverage analysis, gap detection, regression suite management.
```
dustyo-O: did you tested this against developer agent with the same tasks?

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+  - New `context/qa/audit-reports/qa-report-YYYY-MM-DD.md`
+
+---
+
+# PROCESS
+
+## Step 1: Identify scope
+
+1. Read `<user_prompt>`. If it names a spec, target that directory only.
+2. If empty, target all spec directories under `context/spec/`.
```
dustyo-O: that seems to be dangerious, both for spending enormous amount of tokens and making unexpected impact for the project

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+3. Announce: "Running QA audit for: [scope]."
+
+## Step 2: Discover frameworks
+
+1. Read `context/product/architecture.md` for declared testing stack per layer.
+2. Fall back to auto-detection via dependency files: `package.json`, `requirements.txt`, `go.mod`, `Gemfile`, `pyproject.toml`, `pom.xml`.
```
dustyo-O: I would rather define architecture as a must (for project context, decisions about where to store tests, ...), or at least asked user to run architecture before qa. If user insists, we can continue with fallbacks

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Step 3: Load test registry
+
+Read `context/qa/list-of-tests.md`. If it does not exist, create it from the template at `.awos/templates/qa-context-template.md` (or create an empty registry if template is absent).
```
dustyo-O: this file could be easily out-of-date. there should be some process about exploring real codebase for existing tests, and syncing it with this file

also this file will be the pain in back because of conflicts if it's suggested to be updated with every test

I am not sure that this registry is better than grep. for any tests metadata, it is much better to store it IN the test file (some jsdoc/docstrings-like format) or at least near that file in md format (each feature.test.js file has feature.test.js.md file in the same folder)

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+**Existence check:**
+
+- Does the test file still exist in the codebase? If not → flag `MISSING`.
+
+**Spec linkage check:**
+
+- Does the `@spec` annotation reference a spec directory that still exists?
+  - YES → OK.
+  - Spec directory deleted →
```
dustyo-O: in long-living projects specs are deleted, that is normal practice, because they can easily overwhelm claude context. codebase and existing tests are the source of truth

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+**Regression tag:**
+
+- If tagged `@regression` → carry forward to the regression-suite.md sync in Step 8.
+
+## Step 5: Gap analysis
+
+For each acceptance criterion in the functional spec(s) in scope:
+
+1. Check which pyramid layers have coverage (unit / integration / e2e / contract).
```
dustyo-O: for frontend, there could be component tests, which can be called unit, but they are more comprehensive than units, so consider splitting them to separate class

dustyo-O: in small projects, excluding e2e and integration tests could be intentional decisions to avoid infrastructure overhead. this architectural decisions should be taken into account here

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+2. Check whether each covered layer has a negative test counterpart.
+3. Record gaps:
+   - Layer with no tests at all → `MISSING LAYER`
+   - Layer with only positive tests → `MISSING NEGATIVE`
+
+## Step 6: Generate missing tests
+
+For each gap identified in Step 5, invoke the `Task` tool with `subagent_type: "testing-expert"`. Pass:
```
dustyo-O: why testing expert? what if we have more focused experts in lineup?

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+- The gap description (layer, spec, positive/negative scope)
+- The contents of `functional-spec.md` and `technical-considerations.md` for the target spec
```
dustyo-O: what of there is no spec description?

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+2. Ask user to choose:
+   - **A) Full suite** — run all N tests
+   - **B) Regression suite only** — run M tests tagged @regression
+   - **C) Skip** — do not run tests this session
+3. Wait for user confirmation before executing.
+4. Run the selected suite. Collect pass/fail results per layer.
+
+## Step 8: Update regression suite
+
+If `context/qa/regression-suite.md` does not exist, create it from the template at `.awos/templates/regression-suite-template.md` (or create an empty file with a `# Regression Suite` header if the template is absent).
```
dustyo-O: file is absent

dustyo-O: but I also suggest tagging real tests instead of creating separate registry, which could be difficult to keep up-to-date

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+Report summary to user and list any flags requiring human attention.
+
+---
+
+# TODO
+
+- **E2E tests are ephemeral — no CI artifact.** Step 6 delegates E2E gap test generation to `testing-expert`, but the output is an in-session test run, not a committed rerunnable artifact. Future work: have `testing-expert` generate playwright-cli script files (e.g., `tests/e2e/*.sh` or `.ts`) as output for E2E gaps, so they can be committed and executed in CI without agent interaction.
+
+- **QA audit is coverage-by-inspection, not coverage-by-measurement.** Step 5 (gap analysis) reads source files and infers coverage from test file contents. It does not invoke actual coverage tooling (`vitest --coverage`, Istanbul, c8, pytest-cov, etc.). This means untested branches and dead-path gaps are invisible to the audit. Future work: in Step 2, detect available coverage reporters and, in Step 7, run the suite with coverage flags; parse the output to feed real line/branch metrics into the Coverage Summary table.
```
dustyo-O: coverage is a tool which actually measures something strange which cannot be used as-is.

100% coverage does not mean we are actually covered everything.
<100% coverage does not mean we have a problem

the only thing that matters --> coverage dropped from previous run

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+- **Audit reports are snapshots with no regression baseline or enforcement.** Each `/awos:qa` run produces a dated report, but there is no mechanism to diff successive reports, track coverage trend over time, or fail a build when coverage drops below a threshold. Future work: add a `context/qa/coverage-baseline.md` file that stores the last known layer coverage counts; at the end of Step 9, compare current counts against the baseline and surface regressions explicitly in the Flags section.
+
+---
+
+# CONSTRAINTS
+
+- Never modify production code.
+- Never auto-deprecate tests whose spec was deleted — always flag for human review.
```
dustyo-O: tests which are able to run successfully ALWAYS useful.

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+# ROLE
+
+You are a senior QA Architect running a full audit of the test suite for one or all specifications. You analyze existing tests, identify coverage gaps, generate missing tests, manage the regression suite, and produce a structured audit report. You do not modify production code.
+
+---
+
+# TASK
+
+Perform a full QA audit for the target spec(s). Check existing tests for health (missing, stale, deprecated), identify gaps against spec acceptance criteria, generate missing tests, offer to run the suite, and produce an audit report in `context/qa/audit-reports/`.
```
AlexanderMakarov: > Check existing tests for health (missing, stale, deprecated)

For big codebase and after AWOS update it could trigger quite heavy task. Better to scope it somehow.

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+
+---
+
+# INPUTS & OUTPUTS
+
+- **User Prompt (Optional):** <user_prompt>$ARGUMENTS</user_prompt>
+  - Empty = audit all specs
```
AlexanderMakarov: 1. After AWOS update it could trigger quite heavy task.
2. All specs != all code base (if AWOS wasn't used for all tasks or was added to project midterm), so it could make gaps in tests coverage anyway.

I would suggest to always scope to one spec, if not provided - ask user for the scope. Only if user explicitly asked for the whole project - run for the whole project.

commands/qa.md
AlexanderMakarov: I agree with concern but let me add few facts here as well:

> `package.json`, `requirements.txt`, `go.mod`, `Gemfile`, `pyproject.toml`, `pom.xml`

won't provide the full architecture - they usually provide information only about the main language or repo. Also they may be placed in folders/subfolders and CC may miss them. Or it could be one ["BUILD" file orchestrating big monorepo](https://bazel.build/concepts/build-files).

We can't rely on CI files as well, because repos may don't have CI at all, have it partial or have CI configured in a separate CI server.

Therefore I would rather fail if `context/product/architecture.md` is absent and don't try to provide unclear quality of tests for projects without required AWOS artifacts.

commands/qa.md
AlexanderMakarov: Let me note that

> codebase and existing tests are the source of truth

is not the optimal solution because of tokens usage and quality overheads. We [may mitigate](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS_tuWz1ceN/r35) this but fundamentally index of tests is more efficient. If it is up-to-date.

While we should handle absence of specs gently, because it could happen with:
- AWOS was added to repo midterm;
- Specs were archived/outdated (AWOS doesn't have such feature yet, but competitors do);
- Not all tests could be added in scope of spec of with AWOS participation at all.

TL;DR: managing of tests is a separate feature which requires careful design. I would rather split it into separate PR.

commands/qa.md
AlexanderMakarov: "React component test" it is unit test (when component doesn't interact with other services directly, which is a bad design). Please look at https://martinfowler.com/articles/practical-test-pyramid.html

So I would agree with the current layers.

AlexanderMakarov: This is debatable.

If project is so tiny that even integration tests are rejected then why we need to use AWOS in it at all? It is just one-prompt project.

If e2e and integration tests were rejected because quality is not a priority at all (aka prototype/throwaway projects) then we should use "no tests" shortcut/backdoor which was asked for this PR.

commands/qa.md
```diff
@@ -0,0 +1,188 @@
...
+## Step 5: Gap analysis
+
+For each acceptance criterion in the functional spec(s) in scope:
+
+1. Check which pyramid layers have coverage (unit / integration / e2e / contract).
+2. Check whether each covered layer has a negative test counterpart.
+3. Record gaps:
+   - Layer with no tests at all → `MISSING LAYER`
+   - Layer with only positive tests → `MISSING NEGATIVE`
+
```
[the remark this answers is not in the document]
AlexanderMakarov: This classification happens in 

>For each acceptance criterion in the functional spec(s) in scope:

so scope is well defined and intent is "check tests coverage". It should not over-report.

commands/qa.md
AlexanderMakarov: I guess we should fail for such "without prerequisites" project.

commands/qa.md
AlexanderMakarov: @dustyo-O please read https://sqlite.org/testing.html and https://news.ycombinator.com/item?id=18685748 - SQLite is the stable-est app in the world. Because it has 100% coverage. While it had some issues since archiving 100% coverage but taking into account how widespread it is and how much errors it has I may conclude - bigger coverage == quality.

commands/hire.md
FlySpot: Restructured "Complementary pairs" to follow the search-first rule already stated at lines 96-99: for each stack, search the registry for a technology-specific testing agent first and fall back to `testing-expert` only if none is found. In practice `testing-expert` will still be the outcome in most projects — specialized testing agents are rare in the registry — but this framing prevents hardcoding it when a better option is available. Also replaced `playwright` MCP with `playwright` CLI.

commands/hire.md
FlySpot: You're right, it's a duplicate of the FastAPI/Django/Flask entry above. Removed.

commands/qa.md
AlexanderMakarov: It is not true. `assert true`-like tests are quite harmful because:
1. Need to maintain them (I am talking about more exploded versions of "tests" than `assert true`).
2. They provide feel of safety while check nothing.

commands/hire.md
FlySpot: No proof — `testing-expert` was the safe fallback because its presence in the registry is guaranteed after `/awos:hire`. That said, the updated search-first rule (see line 107 fix) now handles this correctly: a dedicated `devops-infra` or `terraform-validator` agent will take priority if found in the registry. We can't assume it exists, but we no longer hardcode `testing-expert` either.

commands/qa.md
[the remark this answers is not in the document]
FlySpot: Fixed with a hybrid two-pass approach. Step 5 now gates gaps before checking any AC:

1. **Project-level gate** — reads `architecture.md` (or auto-detected stack from Step 2) to determine which layers the project actively supports. Unsupported layers are excluded from all gap checks (e.g., "no e2e infrastructure" → never report `MISSING LAYER: e2e`).
2. **Per-AC gate** — for each acceptance criterion, determines applicable layers based on its nature (UI flow → e2e; service boundary → integration; pure logic → unit; public API → contract).

`MISSING LAYER` and `MISSING NEGATIVE` are only reported for layers that pass both gates. This prevents false positives for layers the project intentionally doesn't support.

commands/qa.md
FlySpot: Fair point. Added a confirmation step to Step 1: when no spec argument is given, the command lists all specs found and asks the user to confirm before proceeding. This makes the scope, token cost, and potential test generation visible upfront.

commands/qa.md
FlySpot: Agreed. Updated Step 2: if `architecture.md` is absent, the command warns and asks for explicit confirmation before falling back to auto-detection. Silent fallback was too risky — without architecture context, gap analysis can falsely flag intentionally excluded layers as missing. Consider running `/awos:arch` first if the file doesn't exist yet.

commands/qa.md
FlySpot: Valid long-term concern. The central registry is a practical choice — one file for the agent to read vs. scanning every test file on each run. Staleness is mitigated by the existence check in Step 4. Added to TODO: future work is to generate the registry automatically from inline `@spec`/`@regression` annotations in test files, making the annotations the source of truth.

commands/qa.md
FlySpot: Already handled in Step 4 "Spec linkage check": when a spec directory is deleted, the command searches all active `functional-spec.md` files for matching behavior/acceptance criterion. If found → re-links. If not found → flags `HUMAN REVIEW`, never auto-deprecates (the test may be intentional regression coverage). No change needed.

commands/qa.md
FlySpot: Covered by the updated Step 5. Applicable layers are now determined in two passes: first by what `architecture.md` declares the project supports, then by the nature of each AC. If `architecture.md` declares "component" as a distinct test layer (common in React projects) → it's included in gap analysis. No separate change needed beyond the Step 5 fix.

FlySpot: Also covered by the updated Step 5. The project-level gate reads `architecture.md` before checking any AC: if e2e or integration infrastructure is not declared as supported, those layers are excluded from all gap checks entirely. Intentional architectural decisions to skip certain layers are respected.

commands/qa.md
FlySpot: Implemented a coordinator pattern in Step 6:
1. Check `.claude/agents/` for a technology-specific testing agent suited to the gap.
2. If found — specialist agent writes the test code only; then `testing-expert` acts as coordinator: validates and applies annotations (`@layer`, `@spec`, `@regression`), runs RED validation, and updates `list-of-tests.md`.
3. If no specialist found — `testing-expert` handles everything directly.

This way specialist agents contribute their domain knowledge while `testing-expert` ensures protocol consistency and registry integrity regardless of who wrote the test.

commands/qa.md
FlySpot: Added a guard at the start of Step 6: before invoking any agent, the command verifies that `functional-spec.md` exists for the target spec. If missing — skips gap generation and logs `SPEC INCOMPLETE: [spec-dir] — functional-spec.md missing, gap analysis skipped` in the audit report.

commands/qa.md
FlySpot: `list-of-tests.md` is absent by design until the first spec completes the Feature Testing & Regression slice of `/awos:tasks` — it's auto-maintained by `testing-expert` at that point (see `templates/qa-context-template.md`). `/awos:qa` was incorrectly claiming to create it. Fixed Step 3: the command now reads the registry if it exists and proceeds by scanning live test files directly if it doesn't.

FlySpot: See reply on the comment above — the registry stays as the current approach. The inline annotation idea is tracked as future work in the TODO section.

commands/qa.md
FlySpot: Good point. Updated the TODO to clarify the goal: delta detection (did coverage drop from the previous run?) rather than enforcing absolute thresholds — absolute numbers are meaningless in isolation.

commands/qa.md
FlySpot: `testing-expert` has been validated on a live project throughout a real SDLC cycle — not a synthetic benchmark, but actual feature delivery from spec to merge. The QA Architect role is orthogonal to the developer flow: it inspects existing tests and specs rather than implementing features, so direct overlap is minimal.

Also updated the command description to better reflect its actual scope after the changes in this PR.

commands/qa.md
```diff
@@ -0,0 +1,224 @@
...
+  - YES → OK.
+  - Spec directory deleted →
+    - Search all active `functional-spec.md` files for matching behavior/acceptance criterion.
+    - Match found → re-link `@spec` to the new spec, mark OK.
+    - No match → flag `HUMAN REVIEW` (do NOT auto-deprecate; may be intentional regression coverage).
+
+**Staleness check:**
+
+- Does the test logic match the current implementation?
+  - Read the test file and the relevant implementation code.
```
AlexanderMakarov: Weak metrics. Consider to use some [code coverage metrics](https://en.wikipedia.org/wiki/Code_coverage) instead.

commands/qa.md
```diff
@@ -0,0 +1,224 @@
...
+4. If no specialist found — invoke `testing-expert` directly to handle everything. Pass:
+   - The gap description (layer, spec, positive/negative scope)
+   - The contents of `functional-spec.md` and `technical-considerations.md` for the target spec
+   - The relevant implementation source code
+
+Do not write tests inline here.
+
+## Step 7: Run tests (with user confirmation)
+
+1. Count all tests in scope. Use `AskUserQuestion` with the following question and options:
```
AlexanderMakarov: (nit) I would rather ask LLM to decide (basing on current spec/code changes) - interrupting user is much longer and usually provides the same quality of decision nowadays.

commands/qa.md
```diff
@@ -0,0 +1,224 @@
...
+- [ ] Needs attention — [N] gaps or failures require action
+```
+
+## Step 10: Announce
+
+Report summary to user and list any flags requiring human attention.
+
+---
+
+# TODO
```
AlexanderMakarov: LLM would read this section as context like any other section. So mentioning future work here just dilutes context. Please remove or put into supported by CC `<!-- TODO: ... -->` comments.

commands/qa.md
```diff
@@ -0,0 +1,224 @@
...
+
+- **Audit reports are snapshots with no regression baseline or enforcement.** Each `/awos:qa` run produces a dated report, but there is no mechanism to diff successive reports, track coverage trend over time, or fail a build when coverage drops below a threshold. Future work: add a `context/qa/coverage-baseline.md` file that stores the last known layer coverage counts; at the end of Step 9, compare current counts against the baseline and surface drops explicitly in the Flags section. Goal is delta detection (did coverage drop?), not enforcing absolute thresholds — absolute numbers are meaningless in isolation.
+
+---
+
+# CONSTRAINTS
+
+- Never modify production code.
+- Never auto-deprecate tests whose spec was deleted — always flag for human review.
+- Always ask for user confirmation before running tests (use `AskUserQuestion`).
```
AlexanderMakarov: Why we need to bother user for a few seconds background task?

commands/qa.md
```diff
@@ -0,0 +1,224 @@
...
+
+- **Audit reports are snapshots with no regression baseline or enforcement.** Each `/awos:qa` run produces a dated report, but there is no mechanism to diff successive reports, track coverage trend over time, or fail a build when coverage drops below a threshold. Future work: add a `context/qa/coverage-baseline.md` file that stores the last known layer coverage counts; at the end of Step 9, compare current counts against the baseline and surface drops explicitly in the Flags section. Goal is delta detection (did coverage drop?), not enforcing absolute thresholds — absolute numbers are meaningless in isolation.
+
+---
+
+# CONSTRAINTS
+
+- Never modify production code.
```
AlexanderMakarov: Not agree. Tests allows to find bad design patterns and I time to time change production code after trying to cover it with tests without good results.

FlySpot:
**Update: `commands/qa.md` has been removed** (commit 5245b66)

After reviewing all the feedback on this file and mapping it against the broader command set, we concluded that `commands/qa.md` had no clear place in the current architecture:

- It had no active Claude Code wrapper (the `/awos:qa` slash command was already removed earlier in this PR)
- Its **proactive test writing** functionality is fully covered by `testing-expert` (called from the Feature Testing & Regression slice in `/awos:tasks`)
- Its **regression suite management** (Step 8) is covered by `/awos:regression` (PR #114)
- The remaining unique value (retroactive staleness/gap auditing across all specs) is a valid use case but belongs in a dedicated plugin, not in the core command set

This makes most of the open comments on `commands/qa.md` moot. Thank you @dustyo-O and @AlexanderMakarov for the thorough review — the feedback directly informed this decision.

commands/hire.md
```diff
@@ -96,6 +96,29 @@ Follow this process precisely.
...
+Svelte, etc.), always include `playwright` CLI in the proposal — it enables
+browser-based E2E testing via the testing-expert agent.
+
+Complementary pairs reference:
+For each stack, search the registry for a technology-specific testing agent.
+If none found, fall back to `testing-expert`.
+
+- React / Vue / Angular → + `playwright` CLI for E2E
+- FastAPI / Django / Flask → + `pytest-best-practices` skill _(if not in registry, `testing-expert` covers it)_
+- Other Python backends → + `pytest-best-practices` skill _(if not in registry, `testing-expert` covers it)_
```
AlexanderMakarov: (nit) We may merge two lines above together. Not fixed from previous review.

commands/hire.md
```diff
@@ -96,6 +96,29 @@ Follow this process precisely.
...
+- TypeScript / Node backend → (no additional skill needed)
+- Terraform / IaC → search for an infra-validation agent _(if not in registry, `testing-expert` covers plan/output validation)_
```
AlexanderMakarov: > for an infra-validation agent

Why "infra-validation" here without backticks formatting as other agents above? Actually I can't find such agent in AWOS-recruitment, so probably better to explain agent in some common words without using [kebab case](https://en.wiktionary.org/wiki/kebab_case) (to match future agents), i.e. "infra-validation" -> "technology specific or dedicated infrastructure validation agent".

commands/tasks.md
```diff
@@ -36,8 +36,9 @@ Follow this process precisely.
 
 ## Step 1: Identify the Target Specification
 
-1.  Analyze `<user_prompt>`. If it clearly references a spec by name or index, identify the corresponding directory in `context/spec/`.
-2.  If the prompt is empty or ambiguous, list the spec directories that contain both `functional-spec.md` and `technical-considerations.md` and ask the user to choose. Do not proceed until a valid spec is selected.
+1.  **Detect `--no-tests` flag:** Before anything else, check whether the `<user_prompt>` contains `skip tests` or `--no-tests` (case-insensitive). If found, set an internal flag `SKIP_TESTS = true`. This flag suppresses all verification sub-tasks inside slices and omits the Feature Testing & Regression slice entirely. Strip the flag from the prompt before spec identification.
```
AlexanderMakarov: >**Detect `--no-tests` flag:** Before anything else, check whether the `<user_prompt>` contains `skip tests` or `--no-tests` (case-insensitive). If found, set an internal flag `SKIP_TESTS = true`.

I would propose to don't shape it as a flag. Because it may fail on human's wording - i.e. with "don't generate qa stuff now" phrase would probably be not detected here as "--no-tests flag" while I may bet that user wanted exactly this.

Like:

```suggestion
1.  If `<user_prompt>` explicitly mentions that tests are not expected then skip all verification sub-tasks inside slices and omits the Feature Testing & Regression slice entirely.
```

commands/tasks.md
```diff
@@ -55,6 +56,7 @@ Follow this process precisely.
   - Check that the project has the MCPs, services, and dependencies needed for testing each slice. If something is missing, instruct the user to install it.
   - If a slice cannot be tested, explain why and get user approval before proceeding.
   - A slice is not complete unless it is tested or the user has explicitly approved skipping the test.
+  - After a slice is verified and marked complete, **delete all temporary artifacts** generated during that slice's verification — screenshots, recorded videos, generated e2e test scripts, and any other ephemeral files produced by the e2e-tester or browser MCP. Exception: do **not** delete artifacts from the **Feature Testing & Regression** slice — those are intentionally kept for the regression suite.
```
AlexanderMakarov: >e2e-tester

What it is?

src/config/setup-config.js
```diff
@@ -66,6 +66,12 @@ const copyOperations = [
     manualUpdateUrl:
       'https://github.com/provectus/awos/tree/main/claude/commands',
   },
+  {
+    source: 'plugins/awos/agents',
+    destination: '.claude/agents',
+    patterns: ['*'],
+    description: 'AWOS agents',
+  },
```
AlexanderMakarov: `.claude/agents/` is the user's customization area. This copy operation has no `preserveOnUpdate: true`, so every `npx @provectusinc/awos` will silently clobber any user agent file whose name collides with an AWOS-shipped one. Same UX bug PR #121 just fixed for `.claude/commands/awos/`.

So please don't ship agents through the installer at all — per https://provectus.slack.com/archives/C09GCR80NC8/p1779723615299239?thread_ts=1778491156.726109&cid=C09GCR80NC8 `testing-expert` is to be hired from `awos-recruitment`, which contradicts this copy op.

commands/verify.md
```diff
@@ -32,6 +32,19 @@ Verify a specification's implementation against its acceptance criteria. For eac
 
 ---
 
+# CONSTRAINTS
+
+- **Step 3 is MANDATORY.** Never skip verification. Never write "the user can verify this manually"
```
AlexanderMakarov: The new `--no-tests` / `skip tests` flag in `commands/tasks.md` (line 39, also see relevant comment) lets the user opt out of test generation. But this CONSTRAINTS block then says "Step 3 is MANDATORY. Never skip verification." A user who ran `/awos:tasks ... skip tests` will hit a wall here.

Either:
- This block needs to acknowledge the no-tests path (e.g. if the spec was generated with `SKIP_TESTS=true`, Step 3 falls back to look-and-feel / spec-freshness only), OR
- `/awos:verify` itself should detect skip-tests intent (from prompt or a persisted flag) and adjust.

Aslo see https://provectus.slack.com/archives/C09GCR80NC8/p1779454777693989?thread_ts=1778491156.726109&cid=C09GCR80NC8 - `/awos:verify` was originally for spec-freshness and it is OK to add "look-and-feel" checks, but not running all test suites.

commands/verify.md
```diff
@@ -32,6 +32,19 @@ Verify a specification's implementation against its acceptance criteria. For eac
...
+  unless you have: (a) tried all available verification tools (browser MCP, curl, shell), AND
+  (b) received explicit user approval to skip via AskUserQuestion.
+- **Verification tool fallback order:** browser MCP → curl/shell → AskUserQuestion to skip.
```
AlexanderMakarov: > `browser MCP → curl/shell → AskUserQuestion`

still hardcodes a tool priority. Let CC pick by efficiency/quality first and wall-clock time second, rather than dictating browser MCP first. Cypress/WebdriverIO/playwright-cli projects, headless-only stacks, and curl-only API specs are all penalised by this fixed order.

Reframe as "use any available verification tool — prefer reliable/faster/cheaper for the given AC," or wrap the ladder as "typical example, not prescribed."

commands/hire.md
```diff
@@ -96,6 +96,29 @@ Follow this process precisely.
...
+testing agent. Use the query: "[primary technology] testing QA acceptance".
+
+If `testing-expert` is returned from the registry, always include it in the
+proposals table with label "(Recommended for QA coverage)" alongside the
+primary tech agent — unless a more specific testing agent (e.g. `react-testing`,
+`rust-tester`) is also found, in which case prefer the specific one.
+
+Also: if the tech stack includes any frontend framework (React, Vue, Angular,
+Svelte, etc.), always include `playwright` CLI in the proposal — it enables
+browser-based E2E testing via the testing-expert agent.
```
AlexanderMakarov: Hardcoding `playwright` CLI for every frontend stack contradicts the search-first principle stated a few lines below (line 113: "search the registry for a technology-specific testing agent"). Cypress, WebdriverIO, Vitest-browser, Storybook test-runner all exist and may be a better fit per project.

Same anti-hardcoding concern as `commands/verify.md` line 40. The bullet at line 116 (`React / Vue / Angular → + playwright CLI for E2E`) repeats the same hardcoding.

commands/hire.md
```diff
@@ -96,6 +96,29 @@ Follow this process precisely.
...
+
+Complementary pairs reference:
+For each stack, search the registry for a technology-specific testing agent.
+If none found, fall back to `testing-expert`.
```
AlexanderMakarov: Two paragraphs cover the same testing-expert fallback logic with conflicting defaults:

- Lines 99-106 (QA Complement Rule): "always include `testing-expert` in proposals … unless a more specific testing agent is also found."
- Lines 112-114 (Complementary pairs reference): "search the registry for a technology-specific testing agent. If none found, fall back to `testing-expert`."

First frames testing-expert as `always-include-with-caveat`, the second as `search-first-with-fallback`. Both rules currently apply → ambiguous behavior. Pick one framing and remove the other.

commands/tasks.md
```diff
@@ -83,11 +128,23 @@ Follow this process precisely.
...
       - `[ ] Sub-task: Update the user API endpoint to return the avatar_url. **[Agent: python-expert]**`
       - `[ ] Sub-task: Update the 'ProfileAvatar' component to fetch and display the user's avatar_url, falling back to the placeholder if null. **[Agent: react-expert]**`
-      - `[ ] Sub-task: Run the application. Use chrome MCP to connect the page in Browser. Verify that the profile page shows the correct avatar or placeholder. **[Agent: manual-qa-expert]**`
+      - `[ ] Verify: Run the application, use playwright-cli to open the profile page in a browser, confirm correct avatar or placeholder is shown. **[Agent: manual-qa-expert]**`
+      - `[ ] Cleanup: Delete any screenshots, videos, or e2e scripts generated during this slice's verification. **[Agent: general-purpose]**`
+    - `[ ] **Slice 3: Feature Testing & Regression**`
+      - `> Verifies the complete feature works end-to-end as described in functional-spec.md.`
+      - `> Run AFTER all implementation slices are complete.`
+      - `> **Requires \`testing-expert\` agent.\*\* If it is not present in \`.claude/agents/\`, stop and run \`/awos:hire\` before executing this slice.`
```
AlexanderMakarov: Markdown escape artifacts. The `\*\*` and backslash-escaped backticks look unintentional — likely from copying the template at line 98 (which renders fine) into this example with extra escaping. Compare to template:

```
> **Requires `testing-expert` agent.** If it is not present in `.claude/agents/`
```

The example line will render with literal `\*\*` instead of bold.

BTW a good place to use XML tags.

commands/tasks.md
```diff
@@ -66,13 +68,56 @@ Follow this process precisely.
...
-  5.  Next, identify the second-smallest piece of value that builds on the first. This is **Slice 2**.
-  6.  Create a high-level checklist item and its sub-tasks with subagent assignments.
-  7.  Repeat this process until all requirements from the specification are covered.
-  8.  For each slice's verification sub-task, identify required MCPs/services (browser MCP, curl, database access, etc.) and note any that may be missing.
+  5.  **If `SKIP_TESTS = true`**, omit the verification sub-task and the cleanup sub-task for this slice entirely. Skip to generating the next slice.
+
+      **If `SKIP_TESTS = false`**, after the verification sub-task, add a cleanup sub-task as the last item of the slice:
+
+      ```md
+      - [ ] Cleanup: Delete any screenshots, videos, or e2e scripts generated during this slice's verification. **[Agent: general-purpose]**
```
AlexanderMakarov: (nit) `[Agent: general-purpose]` for the cleanup step — `rm` of known-named artifacts is a one-line shell command. SDD-07 audit dimension flags "majority of implementation tasks use general-purpose" as a warn condition; cleanup sub-tasks add to that count without adding decision value. Two options:
- Inline the `rm` into the Verify step (no separate sub-task).
- Exempt cleanup sub-tasks from the SDD-07 count, and document that exemption in `plugins/awos/skills/ai-readiness-audit/dimensions/spec-driven-development.md`.

templates/qa-context-template.md
```diff
@@ -0,0 +1,11 @@
+# Test Registry
+
+> Auto-maintained by `testing-expert` agent (called from the Feature Testing & Regression slice).
```
AlexanderMakarov: Future-debt note: this template ships, but the contract for keeping it in sync with the real codebase is informal (just "auto-maintained by `testing-expert`"). It is just not used in this repo. 

The same concerns raised in the now-deleted `commands/qa.md` threads (registry staleness, merge-conflict hotspots) apply here.

Consider to move it into https://github.com/provectus/awos-recruitment/pull/60 entirely or into future PR-s.

AlexanderMakarov (dismissed):

commands/tasks.md
```diff
@@ -66,13 +68,56 @@ Follow this process precisely.
...
+
+      Otherwise, after all implementation slices are defined, append one final slice. This slice is generated automatically — do not ask the user about it.
+
+      ```md
+      - [ ] **Slice N: Feature Testing & Regression**
+
+        > Verifies the complete feature works end-to-end as described in functional-spec.md.
+        > Run AFTER all implementation slices are complete.
+        > **Requires `testing-expert` agent.** If it is not present in `.claude/agents/`,
+        > stop and run `/awos:hire` before executing this slice.
```
AlexanderMakarov: This would cause interruption for `/awos:implement` command while we may catch absence of `testing-expert` just here and form request to user with 3 options:
1. Install agent,
2. Skip slice with tests.
3. Add slice but without guarantees for quality (BTW not sure it is possible at all).

AlexanderMakarov: These `> ` note lines leak meta-instructions into the artifact. Testing this branch on a real spec, the rendered slice carried a `> QA agent for this slice: testing-expert (selected from .claude/agents/ and the Agent tool's description block)` line — the agent-*selection rationale* ended up in the user's file. Recommend collapsing to a single one-line description (matching how every other slice has one `> ` description) and dropping the selection note — the `**[Agent: ...]**` markers already state who runs the slice. The `> **Requires testing-expert** … run /awos:hire` gate is likewise superseded by the search-first QA-agent selection.

commands/tasks.md
```diff
@@ -66,13 +68,56 @@ Follow this process precisely.
...
+              for long-term regression. **[Agent: testing-expert]**
+        - [ ] Run all generated tests. All must pass. Fix any failures before proceeding.
+            **[Agent: testing-expert]**
+        <!-- TODO: enable when feat/regression merges
+        - [ ] Run `/awos:regression [spec-directory-name]` to review candidates for the
+              regression suite, resolve duplicates, and optionally execute the full
+              regression suite. Pass the current spec directory name as the argument
+              (e.g., `/awos:regression 003-user-avatar`). Do not run without an argument —
+              auto-detection requires all tasks to be complete, which is not yet the case.
+              -->
```
AlexanderMakarov: Better to drop this `<!-- TODO: enable when feat/regression merges -->` block. Its trigger lives in a *different* PR (#114) in this repo, so it can't be tracked from a generated `tasks.md` sitting in a user's project — it just ships as dead commentary in every user's file. Let PR #114, which owns `/awos:regression`, add the call to this slice when it lands. The duplicate block in the example below (~line 145) should go too.

commands/verify.md
```diff
@@ -32,6 +32,19 @@ Verify a specification's implementation against its acceptance criteria. For eac
...
+  If browser MCP is unavailable, try curl or shell commands. Only ask to skip as a last resort.
```
AlexanderMakarov: (found this verifying a real UI-heavy spec) Even with browser-MCP listed first, `/awos:verify` produced **no visual verification and no screenshots**. On an already-implemented (Status: Completed) spec the agent took the fastest-evidence path — ran the existing pytest/component suite and read the code — and never rendered the UI. NiceGUI / React-Testing-Library (in my case) test-client tests confirm logic, not look-and-feel, so the run passed with zero rendered-pixel evidence.

Two gaps:
1. Nothing forces *real rendering* for UI/visual acceptance criteria — in-process component tests are accepted as sufficient. For look-and-feel this is not enough - the actual UI must be visible to a human.
2. No screenshots are captured or retained, so a human can't review the look-and-feel afterward.

Suggest aligning with the `testing-expert` agent (awos-recruitment): it already writes E2E captures to `docs/screenshots/` and treats git-ignoring that folder as one-time project setup (it never edits `.gitignore`). `/awos:verify` should (a) require driving the real UI for visual ACs — component tests don't count, (b) save screenshots to that same `docs/screenshots/` folder, named per-spec so they sort by spec, and (c) reference the paths in its report. Reuse testing-expert's folder + no-.gitignore-edit convention rather than inventing a new evidence location.

AlexanderMakarov (approved):

AlexanderMakarov:
@kmakarychev-dev, @workshur, @FlySpot - please note that https://github.com/provectus/awos-qa/pull/10 adds 2 new e2e tests specifically for this functionality (see "INSTRUCTIONS.md" files for the checks they do), could be executed with something like `bun run e2e:run tasks-emits-feature-testing-regression-slice ; bun run e2e:run tasks-honors-skip-tests-intent`.

kmakarychev-dev (approved):

workshur (approved):

PR #113, opened 2026-04-27
feat(spec): add AC Given-When-Then shape, clarify-first, propagate-to-requirement rules

oleksii-shevchuk:
## Summary

Adds three rule anchors to `commands/spec.md` §Step 3.3 Acceptance Criteria.
No template change, no new files, no dependencies.

## What changes

`commands/spec.md` §Step 3.3 — three new bullets between the existing "manual QA test scripts" rule and the Example Statement:

1. **Three-part shape.** Each AC follows Given (precondition) + When (user action) + Then (visible outcome) — the same shape as the example, matching Gherkin/BDD convention.
2. **Clarify-first.** Ask clarifying questions if information is missing, before writing the AC.
3. **Propagate-to-requirement.** If a clarifying answer affects the parent requirement (not just the AC), update §Functional Requirements first.

### Claude Code prompting
- Explicit rules combined with examples produce more stable structured output than examples alone.
- The new rules use the same vocabulary as the existing example (Given/When/Then) — naming the shape strengthens the signal more than paraphrasing it.
- "Ask before guessing" reduces a documented failure mode where the model fills missing inputs with plausible details.
- "Name the destination" prevents the model from appending new facts only at the point of receipt instead of updating upstream sections.

## Risk and limitations
- **Applicability.** Works for behavioral specs and can be adapted for background or scheduled flows.
- **Trade-off.** Slightly longer prompts, but clearer and more testable acceptance criteria.
- **Scope.** Touches only `/awos:spec`. Downstream skills (`/awos:tech`, `/awos:tasks`, `/awos:implement`) are untouched.

## Summary by CodeRabbit

* **Documentation**
  * Updated functional specification guidelines to require explicit marking of missing details with clarification markers in requirements sections.
  * Enhanced acceptance criteria guidance with Given/When/Then format for concrete, testable scenarios.
  * Clarified workflow for resolving clarification markers before finalizing acceptance criteria.

[Review Change Stack](https://app.coderabbit.ai/change-stack/provectus/awos/pull/113?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/commands/spec.md
+++ b/commands/spec.md
@@ -92,6 +92,9 @@ Your first goal is to determine the **topic** - the single, specific feature or 
 3.  **Acceptance Criteria:**
     - After clarifying a requirement, turn it into a concrete, testable acceptance criterion.
     - Acceptance criteria must read as manual QA test scripts that a non-developer could execute. Describe only what is visible on screen and what the user does — never reference internal system behavior.
+    - Each acceptance criterion should follow the same three-part shape as the example below: a precondition (Given), a user action (When), and a visible outcome (Then). State the precondition explicitly even when it is obvious from context.
+    - If information is missing, ask clarifying questions before writing acceptance criteria.
+    - If a clarifying answer reveals a constraint or detail that belongs to the parent requirement (not just the acceptance criterion), update the requirement statement in §Functional Requirements before continuing. The requirement and its acceptance criteria must agree on level of detail.
     - Example Statement: "Okay, I've captured that. So a clear acceptance criterion would be: 'Given the user is on their profile page, when they upload a PNG file smaller than 5MB, then the new picture appears on their profile and a 'Success' message is shown.' Is that correct?"
 
 4.  **Scope and Boundaries:**
--- a/commands/spec.md
+++ b/commands/spec.md
@@ -87,13 +87,13 @@ Your first goal is to determine the **topic** - the single, specific feature or 
     - For every piece of information the user gives you, think like a tester and clarify ambiguities. If the user answers in technical terms, rewrite the information into plain, user-facing language before including it in the spec.
     - If the user says: "The user needs to be able to upload a profile picture."
     - You MUST ask clarifying questions like: "Great. Let's break that down. What file formats should be allowed (e.g., JPG, PNG)? Is there a maximum file size? What should happen after the upload is successful? What specific error message should the user see if it fails?"
-    - Mark every unresolved detail with `[NEEDS CLARIFICATION: your specific question]` directly in the draft. Example: "The user should see an error message. [NEEDS CLARIFICATION: What should the exact text of the error message be?]"
+    - If information is missing, mark every unresolved detail with `[NEEDS CLARIFICATION: your specific question]` directly in the draft. Example: "The user should see an error message. [NEEDS CLARIFICATION: What should the exact text of the error message be?]"
 
 3.  **Acceptance Criteria:**
     - After clarifying a requirement, turn it into a concrete, testable acceptance criterion.
     - Acceptance criteria must read as manual QA test scripts that a non-developer could execute. Describe only what is visible on screen and what the user does — never reference internal system behavior.
-    - Each acceptance criterion should follow the same three-part shape as the example below: a precondition (Given), a user action (When), and a visible outcome (Then). State the precondition explicitly even when it is obvious from context.
-    - If information is missing, ask clarifying questions before writing acceptance criteria.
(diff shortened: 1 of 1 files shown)


commands/spec.md
```diff
@@ -86,6 +86,9 @@ Your first goal is to determine the **topic** - the single, specific feature or
 3.  **Acceptance Criteria:**
     - After clarifying a requirement, turn it into a concrete, testable acceptance criterion.
     - Acceptance criteria must read as manual QA test scripts that a non-developer could execute. Describe only what is visible on screen and what the user does — never reference internal system behavior.
+    - Each acceptance criterion should follow the same three-part shape as the example below: a precondition (Given), a user action (When), and a visible outcome (Then). State the precondition explicitly even when it is obvious from context.
+    - If information is missing, ask clarifying questions before writing acceptance criteria.
+    - If a clarifying answer reveals a constraint or detail that belongs to the parent requirement (not just the acceptance criterion), update the requirement statement in §Functional Requirements before continuing. The requirement and its acceptance criteria must agree on level of detail.
```
AlexanderMakarov, line 97: Thanks for this — the rules are well-motivated and the change is small. Two concerns worth resolving before merge:

**1. Overlap with the §3.2 ambiguity policy.** Step 3.2 already prescribes an ambiguity-handling pattern at line 84:

> **MARK ALL AMBIGUITIES:** If a detail cannot be confirmed by the user, you MUST use the `[NEEDS CLARIFICATION: your specific question]` tag directly in the draft.

The new bullet at line 90 ("If information is missing, ask clarifying questions before writing acceptance criteria") prescribes a different posture — interactive clarification rather than inline tagging — for the same underlying situation. As written, the model now has two competing instructions about what to do when a fact is missing: tag-in-draft (§3.2) vs. ask-then-write (§3.3). Could you either:
- explicitly state how the two relate (e.g., "ask interactively at §3.3 if the §3.2 `[NEEDS CLARIFICATION]` tag is still unresolved when you reach AC"), or
- collapse to a single ambiguity rule at §3.2 and reference it from §3.3?

The propagate-to-requirement bullet has a related issue — it tells the model to update §Functional Requirements *after* it has already moved to §3.3, which is fine, but the §3.2 tag flow doesn't acknowledge that backtracking exists. A one-line forward reference from §3.2 would close the loop.

**2. UI-bias gets tighter, not looser.** The existing §3.3 rule already biases AC to UI flows ("describe only what is visible on screen"). The new "explicit precondition even when obvious" + GWT shape pushes this further — both fit UI specs cleanly but produce stilted AC for background, scheduled, or API-only flows (e.g., "Given the daily ingestion job runs at 02:00 UTC, when…"). The PR description notes the rule "can be adapted" for those flows, but the rule text gives the model no signal to deviate. Options:
- soften "should follow" to "follow GWT for user-facing flows; for background/scheduled flows describe trigger → action → observable outcome", or
- keep GWT but explicitly exempt non-UI flows from the "state the precondition explicitly even when obvious" requirement.

Otherwise looks good — happy to approve once the §3.2/§3.3 interaction is clarified.

oleksii-shevchuk:
Thanks for the review! Both concerns addressed:

  **Concern 1 (clarify-first ↔ §3.2 ambiguity policy)** — went with collapse-and-reference (your second option). §3.2 now explicitly states its trigger ("If information is missing, mark..."), and §3.3 references that mechanism:

  > If any `[NEEDS CLARIFICATION: …]` markers remain on the parent requirement in §Functional Requirements, ask clarifying questions and resolve the markers before writing acceptance criteria.

  One mechanism, two phases, no more parallel tactics.

  **Concern 2 (UI-bias)** — softened from "even when obvious" to a condition:

  > Each acceptance criterion follows the same three-part shape ... Include Given only when the precondition affects the outcome.

  This removes stilted Givens for trivial preconditions.

  I did NOT add a separate branch for background/scheduled/API-only flows. Reason: lines 11-17 (Language Rules), line 75, and especially the bullet immediately above (line 94 — "describe only what is visible on screen ... never
  reference internal system behavior") explicitly scope this command to UI flows. Adding "background/scheduled flows describe trigger → action → outcome" would directly contradict line 94.

  If we want `/awos:spec` to also cover non-UI flows, I think that's a broader scope change worth a separate PR — happy to discuss.

  Minor notes:
  - Rebased onto current main; #116 touched the same file but the diffs didn't overlap.
  - The propagate-to-requirement bullet (line 97) was left unchanged since it wasn't flagged.

  Ready for re-review.

AlexanderMakarov (approved):

AlexanderMakarov:
@oleksii-shevchuk thank you for the contribution!

FYI: added "behavioral" test `spec-uses-gwt-acceptance-criteria` for your change in https://github.com/provectus/awos-qa/pull/4

PR #124, opened 2026-05-22
feat: add tutorial skill

tigra:
## What

Adds an `/awos:tutorial` command that produces a per-increment learning tutorial for a completed spec: a narrative-first walkthrough plus illustrative code snippets and an optional mermaid diagram.

A companion `concepts.md` ledger dedups concepts across increments, so each tutorial teaches only what its increment newly introduced and references prior concepts by slug. When the target spec is not the most recent one, the skill reconstructs the historical code state via **read-only** git archaeology rather than documenting today's code.

## Files

- `commands/tutorial.md` + `claude/commands/tutorial.md` wrapper
- `templates/tutorial-template.md`, `templates/concepts-template.md`
- `scripts/create-tutorial-directory.sh` (mirrors the existing `create-spec-directory.sh`)
- Integration hook: `commands/verify.md` suggests `/awos:tutorial` after a spec is verified (non-blocking suggestion; existing flow unchanged)

## Validation

- `npm test` → 69/69 pass
- `npx prettier --check .` clean

Suggested release-drafter label: **minor** (new feature skill).


dashagaranina:
Closing as part of the philosophy-alignment cleanup (#194, docs/refactoring-roadmap.md): the philosophy's Is-Not list includes "not a document generator," and a tutorial generator sits fully outside the path (intent → confirmed understanding → implementation → verification). Nothing in the flow reads what it would produce, which is the memory regression the direction docs warn about. Thanks for the polished skill — it's a good fit for a personal or project-local skill outside awos.

PR #127, opened 2026-05-25
feat(implement): loop over every incomplete task by default

AlexanderMakarov:
## Summary

- `/awos:implement` now loops through every incomplete task in the selected spec by default, instead of stopping after the first one — eliminating the "run the command N times" friction for multi-task specs.
- The single-task escape hatch is preserved: when the user names a specific task in `<user_prompt>`, scope is restricted to just that task. The orchestrator-only contract is unchanged — the loop is over delegations, not direct edits.
- Per-task `**[Agent: name]**` re-extraction is explicit, so different tasks in one spec can route to different specialists. `tasks.md` is re-read each iteration so the next task is selected from the latest on-disk state. On subagent failure the loop halts so the user can decide before more work is dispatched.

## Test plan

- [x] `bun test tests/` — 69 pass / 0 fail (static prompt linter contracts still hold: `**[Agent: ` reference, `<scope_discipline>`/`<investigate_before_answering>`/`<use_available_skills>` XML blocks, `Agent(subagent_type=` example, `# INTERACTION` + `AskUserQuestion`).
- [x] `bunx prettier . --check` — clean.
- [x] Behavioral end-to-end in [`awos-qa`](https://github.com/provectus/awos-qa): the `implement-orchestrator-only` scenario has been reworked into `implement-orchestrates-and-loops` (two slices, two distinct specialists, three new loop-specific assertions). Land that PR alongside this one; instructions for a manual run are in the scenario's `INSTRUCTIONS.md`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified implementation command to run tasks in an explicit per-task loop, re-read and mark tasks as completed, cascade slice/header completion, and report percent-complete status.
  * Formalized vertical “slices” as runnable end-to-end units with nested atomic tasks and required verification steps.
  * Specified agent assignment workflow (including specialist vs fallback) and verification/testing requirements per slice.
  * Updated guidance to assess agent coverage at the task/slice level.

[Review Change Stack](https://app.coderabbit.ai/change-stack/provectus/awos/pull/127?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/commands/implement.md
+++ b/commands/implement.md
@@ -36,21 +36,20 @@ Your goal is to execute the pending work for a given specification, one task at 
 
 Follow this process precisely. Steps 2–5 form the per-task loop: repeat them for each task in scope, in document order, until the scope is exhausted. Step 6 runs once after the loop.
 
-### Step 1: Identify the Target Specification and Scope
+### Step 1: Identify the Target Specification and Load Static Context
 
 1.  Analyze `<user_prompt>`. If it names a specific task, set scope to that single task in the spec it belongs to. If it names a spec (without a specific task), set the target spec from the prompt and set scope to "every incomplete (`[ ]`) task in that spec".
 2.  Otherwise (no prompt): scan `context/spec/` in order, find the first directory whose `tasks.md` has an incomplete item (`[ ]`), select it as the target spec, and set scope to "every incomplete task in that spec".
 3.  If no target can be determined (ambiguous prompt, or all tasks are done), tell the user and stop.
-
-### Step 2: Load Full Context and Pick the Next Task
-
-1.  On the first iteration, load all three context files in parallel:
+4.  Load the static spec context once, in parallel:
     - `[target-spec-directory]/functional-spec.md`
     - `[target-spec-directory]/technical-considerations.md`
-    - `[target-spec-directory]/tasks.md`
 
-    On later iterations, reload only `tasks.md` — the spec files don't change. Re-reading `tasks.md` each iteration ensures the next task is selected from the latest on-disk state.
+    These files don't change during the run; Step 3 embeds their content into the delegation prompt for every task.
 
+### Step 2: Read `tasks.md` and Pick the Next Task
+
+1.  Read `[target-spec-directory]/tasks.md`. Re-reading it each iteration ensures the next task is selected from the latest on-disk state.
 2.  Pick the next task in scope. If the user named a single task, that's the only task; once it's done the loop ends. Otherwise pick the first remaining `[ ]` task in document order from the freshly-read `tasks.md`. If no incomplete tasks remain, exit the loop and go to Step 6.
 3.  Extract the agent assignment from the selected task line:
     - Look for the `**[Agent: agent-name]**` pattern in the task line (e.g., `python-expert`, `react-expert`, `testing-expert`).
@@ -62,7 +61,7 @@ Follow this process precisely. Steps 2–5 form the per-task loop: repeat them f
 You do not write or edit code, configuration, or database schemas yourself. Your role is to delegate.
 
 1.  Construct a delegation prompt that includes:
-    - The full context from the three files loaded in Step 2.
+    - The full context from the three files loaded in Steps 1–2 (`functional-spec.md`, `technical-considerations.md`, `tasks.md`).
     - The specific task description.
     - Clear instructions on what code to write or files to modify.
     - A `<scope_discipline>` block: "Only make changes the task requires. Don't add features, refactor unrelated code, or add validation for scenarios outside the task. If something is unclear, ask rather than guessing."
--- a/commands/implement.md
+++ b/commands/implement.md
@@ -10,7 +10,7 @@ You are a Lead Implementation Agent, acting as an AI Engineering Manager or a pr
 
 # TASK
 
-Your goal is to execute the next available task for a given specification. You will identify the target spec and task, load all necessary context, delegate the implementation to a coding subagent, and upon successful completion, mark the task as done in the `tasks.md` file.
+Your goal is to execute the pending work for a given specification, one task at a time, until the agreed scope is done. By default you loop through every incomplete task in the selected spec in document order; if the user names a single task, you execute only that one. For each task in scope you load context, re-extract its `**[Agent: name]**` marker, delegate to a coding subagent, and on success mark the task as done in `tasks.md` before moving to the next.
 
 ---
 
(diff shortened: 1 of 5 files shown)


commands/implement.md
```diff
@@ -34,23 +34,28 @@ Your goal is to execute the next available task for a given specification. You w
 
 # PROCESS
 
-Follow this process precisely.
+Follow this process precisely. Steps 2–5 form the per-task loop: repeat them for each task in scope, in document order, until the scope is exhausted. Step 6 runs once after the loop.
 
-### Step 1: Identify the Target Specification and Task
+### Step 1: Identify the Target Specification and Scope
```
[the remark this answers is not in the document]
AlexanderMakarov: `### Step N` is the established pattern across every core command (`architecture.md`, `spec.md`, `product.md`, `roadmap.md`, `tech.md`, `verify.md`) and is unchanged from pre-PR `implement.md`. The repo has no `.markdownlint*` config and `CLAUDE.md` does not reference markdownlint, so MD001 is not enforced here. If we want to switch to `## Step`, it should be a separate PR that flips the heading level across all `commands/*.md` uniformly.

commands/implement.md
```diff
@@ -34,23 +34,28 @@ Your goal is to execute the next available task for a given specification. You w
...
 
-1.  Load the three context files in parallel:
+1.  On the first iteration, load all three context files in parallel:
     - `[target-spec-directory]/functional-spec.md`
     - `[target-spec-directory]/technical-considerations.md`
     - `[target-spec-directory]/tasks.md`
-2.  Extract the agent assignment from the task description:
+
+    On later iterations, reload only `tasks.md` — the spec files don't change. Re-reading `tasks.md` each iteration ensures the next task is selected from the latest on-disk state.
+
```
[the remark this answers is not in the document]
AlexanderMakarov: The orchestrator-only contract in `CLAUDE.md` prohibits *editing code* ("The orchestrator is explicitly prohibited from editing code itself"), not *reading files*. Step 3 of this same prompt requires the full content of all three files to be embedded into the delegation prompt ("The full context from the three files loaded in Step 2") — without those reads, the subagent would receive only a task line with no spec context to act on. The reads are also pre-existing: pre-PR `implement.md` already loaded all three files; this PR only clarifies the per-iteration reload pattern (reload `tasks.md` each iteration; the spec files do not change). Not changing this here.

AlexanderMakarov:
Re: [walkthrough](https://github.com/provectus/awos/pull/127#issuecomment-4533730441) — the summary matches the intent of the change. No action item.

commands/implement.md
workshur: The instruction about loading
- `[target-spec-directory]/functional-spec.md`
- `[target-spec-directory]/technical-considerations.md`

can be moved to Step 1, so you don’t need to mention that these files remain unchanged.

> On later iterations, reload only `tasks.md` — the spec files don't change. Re-reading `tasks.md` each iteration ensures the next task is selected from the latest on-disk state.

AlexanderMakarov: Good call — moved the static spec reads (`functional-spec.md`, `technical-considerations.md`) into Step 1 so they happen once with target-spec resolution, and Step 2 now just reads `tasks.md` each iteration. Drops the "first iter vs later iter" branching. Step 3's delegation prompt still embeds all three files (now sourced from Steps 1–2).

commands/implement.md
```diff
@@ -10,7 +10,7 @@ You are a Lead Implementation Agent, acting as an AI Engineering Manager or a pr
 
 # TASK
 
-Your goal is to execute the next available task for a given specification. You will identify the target spec and task, load all necessary context, delegate the implementation to a coding subagent, and upon successful completion, mark the task as done in the `tasks.md` file.
+Your goal is to execute the pending work for a given specification, one task at a time, until the agreed scope is done. By default you loop through every incomplete task in the selected spec in document order; if the user names a single task, you execute only that one. For each task in scope you load context, re-extract its `**[Agent: name]**` marker, delegate to a coding subagent, and on success mark the task as done in `tasks.md` before moving to the next.
```
AlexanderMakarov: Term "task" is too dim here. "one task at a time" is better to remove.

AlexanderMakarov: Renamed the children of a slice from **sub-task** to **task** across `commands/tasks.md` and `commands/implement.md` (commit f270150). The vocabulary is now consistent end-to-end:

- A **slice** is the top-level grouping header — vertical, end-to-end runnable, composite, never executed directly.
- A **task** is the atomic nested checkbox under a slice — carries `**[Agent: name]**`, executed by exactly one subagent. `/awos:implement` iterates over tasks; when all tasks under a slice are `[x]`, the slice header is ticked too.

Dropped the redundant "one task at a time" from line 13 — "every incomplete task ... in document order" already conveys serial execution. Also took the opportunity to tighten Step 2 (explicit skip-slice-headers rule) and Step 6 (count only nested tasks to avoid double-counting slice headers).

Rejected "sub-slice" as a candidate name because the children are not independently shippable — only the parent slice is end-to-end runnable, so a `sub-slice` term would imply something the model does not guarantee.

Same change also propagated into `docs/commands/{tasks,implement}.md` and the SDD-07 audit dimension (which still recognizes the legacy `Sub-task:` label so existing user specs do not fail the dimension after this rename).

commands/tasks.md
```diff
@@ -4,13 +4,15 @@ description: Breaks the Tech Spec into a task list for engineers.
...
+Your goal is to create a markdown file with a comprehensive list of checkbox slices for a given specification. You will identify the target spec, carefully analyze its functional and technical documents, and generate a list where each slice represents a small, end-to-end, runnable increment of the feature, broken down into the atomic tasks needed to implement it. Every slice should contain test scenarios for subagents to verify that the slice is completed correctly. The final list will be saved to `tasks.md` within the spec's directory.
+
+A **slice** is the top-level grouping checkbox — a vertical, end-to-end runnable increment. It is composite and never executed directly. A **task** is the atomic nested checkbox under a slice — it carries a `**[Agent: agent-name]**` marker and is executed by exactly one subagent. `/awos:implement` iterates over tasks; when all tasks under a slice are `[x]`, the slice header is ticked too.
```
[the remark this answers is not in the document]
AlexanderMakarov: Valid gap, but out of scope here. The Creation/Update Mode contract is missing on `main` too — PR #127 doesn't touch tasks.md's write behavior, only renames `sub-task` → `task` and adds the slice/task distinction. Adding mode handling expands the diff well beyond the implement-loop change.

I'll open a follow-up issue/PR for `commands/tasks.md` (and audit the other `commands/*.md` while I'm there).

workshur (approved):

PR #130, opened 2026-05-29
feat: add scan codebase command for brownfield

dustyo-O:
## Release Notes

* **New Features**
  * Introduced `/awos:scan` command to document existing codebases, generating structure and decision records
  * All workflow commands now automatically detect and incorporate scan results for brownfield projects

* **Documentation**
  * Updated command guidance to differentiate brownfield vs. greenfield workflows
  * Expanded architecture documentation to emphasize decisions, infrastructure, security, and observability

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,11 +96,11 @@ AWOS is **spec-driven** — all project state lives in markdown files under `con
 The canonical flow (each command is a markdown prompt under `commands/`):
 
 ```
-/awos:product → /awos:roadmap → /awos:architecture → /awos:hire
-              → /awos:spec → /awos:tech → /awos:tasks → /awos:implement → /awos:verify
+/awos:scan (brownfield only) → /awos:product → /awos:roadmap → /awos:architecture → /awos:hire
+                              → /awos:spec → /awos:tech → /awos:tasks → /awos:implement → /awos:verify
 ```
 
-The first four are run once at project setup; the last five iterate per feature. Each command reads/writes a specific document under `context/` (e.g. `context/product/product-definition.md`, `context/spec/NNN-feature/tasks.md`). The numeric prefix on spec directories is allocated by `scripts/create-spec-directory.sh`.
+`/awos:scan` runs once before product setup on brownfield projects to populate the knowledgebase. The next four commands are run once at project setup; the last five iterate per feature. Each command reads/writes a specific document under `context/` (e.g. `context/product/product-definition.md`, `context/spec/NNN-feature/tasks.md`). The numeric prefix on spec directories is allocated by `scripts/create-spec-directory.sh`.
 
 **Implementation delegation rule:** `/awos:implement` is an orchestrator only — it reads `tasks.md`, extracts the `**[Agent: name]**` marker from each task, and delegates to a subagent. The orchestrator is explicitly prohibited from editing code itself. Preserve this contract when editing `commands/implement.md`.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,10 +97,10 @@ The canonical flow (each command is a markdown prompt under `commands/`):
 
 ```
 /awos:scan (brownfield only) → /awos:product → /awos:roadmap → /awos:architecture → /awos:hire
-                              → /awos:spec → /awos:tech → /awos:tasks → /awos:implement → /awos:verify
+                              → /awos:spec → /awos:tech → /awos:tasks → /awos:implement → /awos:verify → /awos:archive
 ```
 
-`/awos:scan` runs once before product setup on brownfield projects to populate the knowledgebase. The next four commands are run once at project setup; the last five iterate per feature. Each command reads/writes a specific document under `context/` (e.g. `context/product/product-definition.md`, `context/spec/NNN-feature/tasks.md`). The numeric prefix on spec directories is allocated by `scripts/create-spec-directory.sh`.
+`/awos:scan` runs once before product setup on brownfield projects to populate the knowledgebase. The next four commands are run once at project setup; the last five iterate per feature. `/awos:archive` closes the loop after `/awos:verify` — it extracts learnings into the knowledgebase docs, logs the feature, and removes the spec directory. Each command reads/writes a specific document under `context/` (e.g. `context/product/product-definition.md`, `context/spec/NNN-feature/tasks.md`). The numeric prefix on spec directories is allocated by `scripts/create-spec-directory.sh`.
 
 **Implementation delegation rule:** `/awos:implement` is an orchestrator only — it reads `tasks.md`, extracts the `**[Agent: name]**` marker from each task, and delegates to a subagent. The orchestrator is explicitly prohibited from editing code itself. Preserve this contract when editing `commands/implement.md`.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,7 +89,11 @@ Each file in `claude/commands/{name}.md` is a tiny wrapper that points at `.awos
 
 ## Architecture: Document-Centric Workflow
 
-AWOS is **spec-driven** — all project state lives in markdown files under `context/` in the user's project, not in chat history. An AI agent can rehydrate full context by reading the files alone. The canonical flow (each command is a markdown prompt under `commands/`):
+AWOS is **spec-driven** — all project state lives in markdown files under `context/` in the user's project, not in chat history. An AI agent can rehydrate full context by reading the files alone.
+
+**Brownfield projects** can populate `context/spec/knowledgebase/` with two documents that give downstream commands awareness of the existing codebase: `structure.md` (directory layout, module boundaries, data flow) and `decisions.md` (non-standard project decisions that override or extend default agent behavior). These files are producer-agnostic — they can be populated by `/awos:scan` (codebase scan), `/awos:archive` (post-implementation learnings), or manually.
+
+The canonical flow (each command is a markdown prompt under `commands/`):
 
 ```
 /awos:product → /awos:roadmap → /awos:architecture → /awos:hire
--- a/README.md
+++ b/README.md
@@ -21,18 +21,19 @@ npx @provectusinc/awos
 
 This sets up the `.awos/` directory (commands, templates, scripts), the `.claude/commands/awos/` wrappers, and the `context/` directory where your project documents will live. It also registers the AWOS plugin marketplace in your project settings.
 
-> **Running on an existing codebase?** Start with an AI readiness audit to understand how AI-friendly your project is. Install the plugin with `/plugin install awos@awos-marketplace`, then run `/awos:ai-readiness-audit` to get a scored assessment with actionable recommendations for improvement. [Learn more](plugins/awos/README.md)
+> **Running on an existing codebase?** Run `/awos:scan` first to produce `context/spec/knowledgebase/` documents that give all downstream commands awareness of your existing codebase. Then continue with the foundation setup below. You can also run an AI readiness audit — install the plugin with `/plugin install awos@awos-marketplace`, then run `/awos:ai-readiness-audit`. [Learn more](plugins/awos/README.md)
 
 ### Step 2: Foundation Setup
 
 These commands establish your project's foundation. Run them once at the start, and re-run them as your project evolves — requirements change, architecture decisions get refined, and that's normal.
 
-| Command              | What it does                                                                      | Docs                                     |
-| -------------------- | --------------------------------------------------------------------------------- | ---------------------------------------- |
-| `/awos:product`      | Defines the Product — what, why, and for who.                                     | [Details](docs/commands/product.md)      |
-| `/awos:roadmap`      | Builds the Product Roadmap — features and their order.                            | [Details](docs/commands/roadmap.md)      |
-| `/awos:architecture` | Defines the System Architecture — stack, DBs, infra.                              | [Details](docs/commands/architecture.md) |
-| `/awos:hire`         | Hires specialist agents — finds, installs skills and MCPs, generates agent files. | [Details](docs/commands/hire.md)         |
+| Command              | What it does                                                                                  | Docs                                     |
+| -------------------- | --------------------------------------------------------------------------------------------- | ---------------------------------------- |
+| `/awos:scan`         | Scans existing codebase — produces structure and decisions documents for brownfield projects. | [Details](docs/commands/scan.md)         |
+| `/awos:product`      | Defines the Product — what, why, and for who.                                                 | [Details](docs/commands/product.md)      |
+| `/awos:roadmap`      | Builds the Product Roadmap — features and their order.                                        | [Details](docs/commands/roadmap.md)      |
+| `/awos:architecture` | Defines the System Architecture — stack, DBs, infra.                                          | [Details](docs/commands/architecture.md) |
+| `/awos:hire`         | Hires specialist agents — finds, installs skills and MCPs, generates agent files.             | [Details](docs/commands/hire.md)         |
 
 ### Step 3: Feature Development Cycle
 
--- a/README.md
+++ b/README.md
@@ -39,15 +39,16 @@ These commands establish your project's foundation. Run them once at the start, 
 
 Once your foundation is set, iterate through this cycle for each feature on your roadmap. These commands are designed to be run repeatedly — once per feature.
 
-> **Tip**: Don't hesitate to delete specs after implementation. Completed specs can become outdated and confuse the AI. Your code documentation is the source of truth.
+> **Tip**: Run `/awos:archive` after verifying a feature. It extracts learnings into the knowledgebase, logs the feature, and removes the spec directory so completed specs don't accumulate and confuse the AI.
 
-| Command           | What it does                                                                      | Docs                                  |
-| ----------------- | --------------------------------------------------------------------------------- | ------------------------------------- |
-| `/awos:spec`      | Creates the Functional Spec — what the feature does for the user.                 | [Details](docs/commands/spec.md)      |
-| `/awos:tech`      | Creates the Technical Spec — how the feature will be built.                       | [Details](docs/commands/tech.md)      |
-| `/awos:tasks`     | Breaks the Tech Spec into a task list for engineers.                              | [Details](docs/commands/tasks.md)     |
-| `/awos:implement` | Runs tasks — delegates coding to sub-agents, tracks progress.                     | [Details](docs/commands/implement.md) |
-| `/awos:verify`    | Verifies spec completion — checks acceptance criteria, marks Status as Completed. | [Details](docs/commands/verify.md)    |
+| Command           | What it does                                                                                  | Docs                                  |
+| ----------------- | --------------------------------------------------------------------------------------------- | ------------------------------------- |
+| `/awos:spec`      | Creates the Functional Spec — what the feature does for the user.                             | [Details](docs/commands/spec.md)      |
+| `/awos:tech`      | Creates the Technical Spec — how the feature will be built.                                   | [Details](docs/commands/tech.md)      |
+| `/awos:tasks`     | Breaks the Tech Spec into a task list for engineers.                                          | [Details](docs/commands/tasks.md)     |
+| `/awos:implement` | Runs tasks — delegates coding to sub-agents, tracks progress.                                 | [Details](docs/commands/implement.md) |
+| `/awos:verify`    | Verifies spec completion — checks acceptance criteria, marks Status as Completed.             | [Details](docs/commands/verify.md)    |
+| `/awos:archive`   | Archives a completed spec — extracts learnings, logs the feature, removes the spec directory. | [Details](docs/commands/archive.md)   |
 
 > **When to skip the cycle**: Not every change needs a spec. Hotfixes, simple bugfixes, and small edits don't require the full spec workflow — Claude Code's built-in plan mode handles those just fine.
 
--- a/claude/commands/scan.md
+++ b/claude/commands/scan.md
@@ -1,0 +1,6 @@
+---
+description: Scans existing codebase — produces structure and decisions documents for brownfield projects.
+argument-hint: '[scope path, optional]'
+---
+
+@.awos/commands/scan.md
--- a/commands/architecture.md
+++ b/commands/architecture.md
@@ -20,6 +20,7 @@ Your task is to manage the architecture file located at `context/product/archite
 - **Prerequisite Input 1:** `context/product/product-definition.md` (The "what" and "why").
 - **Prerequisite Input 2:** `context/product/roadmap.md` (The implementation phases).
 - **Primary Input/Output:** `context/product/architecture.md` (The file to create or update).
+- **Knowledgebase (Optional):** `context/spec/knowledgebase/structure.md` and `context/spec/knowledgebase/decisions.md` — if present, provide awareness of the existing codebase layout and non-standard project decisions that override or extend default agent behavior.
 
 ---
 
@@ -36,7 +37,8 @@ Follow this logic precisely.
 ### Step 1: Prerequisite Checks
 
 - If either `context/product/product-definition.md` or `context/product/roadmap.md` is missing, stop and tell the user to run `/awos:product` and `/awos:roadmap` first.
-- Otherwise, proceed to the next step.
+- If `context/spec/knowledgebase/structure.md` exists, read it to understand the existing project layout. The existing structure informs decisions, but architecture.md remains the sole authority on the technology stack.
+- Proceed to the next step.
 
 ### Step 2: Mode Detection
 
--- a/commands/archive.md
+++ b/commands/archive.md
@@ -1,0 +1,147 @@
+---
+description: Archives a completed spec — extracts learnings, logs the feature, removes the spec directory.
+---
+
+# ROLE
+
+You are a Knowledge Curator responsible for closing the spec-implement-verify loop. Your job is to extract lasting learnings from a completed specification into the project's shared context documents, log the feature for future reference, and remove the spec directory so it does not accumulate.
+
+---
+
+# TASK
+
+Archive a completed specification (Status: `Completed`). This involves four things: (1) use git history to discover what the implementation changed in the codebase, (2) update `context/spec/knowledgebase/` documents with new patterns and conventions introduced by the feature, (3) append a summary entry to `context/spec/feature-log.md`, and (4) delete the spec directory. Each knowledgebase doc and the feature log must stay under 500 lines.
+
+---
+
+# INPUTS & OUTPUTS
+
+- **User Prompt (Optional):** <user_prompt>$ARGUMENTS</user_prompt>
+- **Primary Context:** A spec directory under `context/spec/` containing:
+  - `functional-spec.md`
+  - `technical-considerations.md`
+  - `tasks.md`
+- **Knowledgebase (Optional):** `context/spec/knowledgebase/structure.md` and `context/spec/knowledgebase/decisions.md` — updated with learnings from the archived feature. If absent, knowledgebase updates are skipped.
+- **Feature Log:** `context/spec/feature-log.md` — appended with a summary of the archived feature. Created if it does not exist.
+- **Output:** Updated knowledgebase docs (if they existed), updated feature log, deleted spec directory.
+
+---
+
+# INTERACTION
+
+- Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+
+---
+
+# PROCESS
+
+Follow this process precisely.
+
+### Step 1: Identify Target Specification
+
+1. Analyze `<user_prompt>`. If it names a spec by name or index (e.g., "archive 002" or "archive user-auth"), use that spec directory.
+2. Otherwise, find the first spec directory in `context/spec/` where `functional-spec.md` has Status: `Completed`.
+3. If no eligible spec is found, tell the user no completed specs are ready for archiving and stop.
+
+### Step 2: Load Context
+
+Read the following files in parallel:
+
+- `[target-spec-directory]/functional-spec.md`
+- `[target-spec-directory]/technical-considerations.md`
+- `[target-spec-directory]/tasks.md`
+- `context/spec/knowledgebase/structure.md` (if it exists)
+- `context/spec/knowledgebase/decisions.md` (if it exists)
+- `context/spec/feature-log.md` (if it exists)
+
+If neither knowledgebase doc exists, note that knowledgebase updates will be skipped — the feature log and spec cleanup still proceed.
+
+### Step 3: Git Archaeology
+
+Delegate to an `Explore` agent to discover the concrete changes the feature introduced to the codebase. The spec directory name contains the feature's short-name, which narrows the search.
+
+```text
+Agent(subagent_type="Explore", description="Trace feature git history", prompt="
+Find all commits related to the feature in this spec directory: [target-spec-directory].
+
+1. Run `git log --oneline -- [target-spec-directory]/` to find commits that touched the spec itself.
+2. For each commit found, run `git show --stat <hash>` to see which source files were modified in the same commit.
+3. Collect the unique set of source files changed across all commits (exclude context/ files).
+4. Run `git log --oneline -- <file1> <file2> ...` on those source files to find related implementation commits.
+5. For the most significant implementation commits (up to 10), run `git show --stat <hash>` to understand the scope of changes.
+
+Report:
+- The list of commits (hash + one-line message) in chronological order
+- The set of source files added or modified
+- New directories or modules created
+- Any new dependencies added (look for package.json, requirements.txt, build.gradle, go.mod changes)
+- New patterns introduced (new middleware, new test helpers, new utility modules)
+")
+```
+
+### Step 4: Extract and Merge Learnings
+
+Using the spec contents (Step 2) and the git archaeology findings (Step 3), determine what lasting information belongs in the knowledgebase docs. Skip this step entirely if no knowledgebase docs exist.
+
+For each doc, identify what the feature introduced:
+
+1. **`context/spec/knowledgebase/structure.md`** — New directories, modules, or architectural patterns. Changes to data flow or file placement rules.
+2. **`context/spec/knowledgebase/decisions.md`** — New non-standard decisions: coding patterns established by this feature, changes to error handling or testing strategy, new integration points, new constraints or tech debt introduced.
+
+Merge the new information into the appropriate sections of each doc. Do not duplicate what is already present. If the feature used existing patterns without introducing new ones, leave that doc unchanged.
+
+**500-line budget:** After merging, count the lines in each updated doc. If any exceeds 500 lines:
+
+- Combine entries that describe similar patterns into a single entry.
+- Remove redundant examples, keeping one representative per pattern.
+- Summarize older, well-established entries before trimming newly added ones — recent additions capture the latest state of the codebase.
+- Trim overly verbose descriptions to their essential content.
+
+### Step 5: Update Feature Log
+
+Append an entry to `context/spec/feature-log.md`. If the file does not exist, create it with the following header:
+
+```markdown
+# Feature Log
+
+Archived features and the capabilities they introduced. Read by `/awos:spec` and `/awos:tech` to find patterns available for reuse.
+
+---
+```
+
+Each feature entry follows this format:
+
+```markdown
+### [NNN] [Feature Name]
+
+- **Archived:** YYYY-MM-DD
+- **Summary:** [1-2 sentences — what the feature does for the user]
+- **Key additions:** [Components, patterns, modules introduced — e.g., "JWT auth middleware, user sessions table, login/register endpoints"]
+```
+
+Fill in the spec index number, the feature name from the spec title, today's date, and extract the summary and key additions from the spec contents and git archaeology.
+
+**500-line budget:** After appending, if the feature log exceeds 500 lines, collapse the oldest entries (lowest spec numbers) into a single summary section titled `### Legacy Features (001–NNN)` with one bullet per collapsed feature: `- **[NNN] [Name]:** [one-line summary]`. This preserves searchability while reducing line count.
+
+### Step 6: Write Updated Files
+
+Write only the files that changed:
+
+1. `context/spec/knowledgebase/structure.md` (if updated)
+2. `context/spec/knowledgebase/decisions.md` (if updated)
+3. `context/spec/feature-log.md`
+
+### Step 7: Confirm and Delete Spec Directory
+
+Before deleting, confirm with the user. Present the directory name and note that git history preserves the original files.
+
+If the user confirms, delete the entire `context/spec/[NNN-feature-name]/` directory.
+
+If the user declines, skip deletion — the feature log and knowledgebase updates are already saved.
+
+### Step 8: Report
+
+- Which knowledgebase docs were updated (one-line summary of what was added to each)
+- The feature-log entry that was appended
+- Whether the spec directory was deleted
+- Next command: `/awos:spec`
--- a/commands/product.md
+++ b/commands/product.md
@@ -24,6 +24,7 @@ Your primary task is to **fill in** a product definition template using a guided
     ```
 2.  **Template File:** Use `.awos/templates/product-definition-template.md` as a template.
 3.  **Existing Definition (Optional):** The file `context/product/product-definition.md`, which, if present, triggers "Update Mode".
+4.  **Knowledgebase (Optional):** `context/spec/knowledgebase/structure.md` — if present, provides awareness of the existing codebase.
 
 ---
 
@@ -63,13 +64,14 @@ First, check if the file `context/product/product-definition.md` exists.
 ### Step 2B: Creation Mode
 
 1.  If `<user_prompt>` is non-empty, briefly note that you'll use it as a starting point, then refine from there.
-2.  Walk the user through the sections of the template, explaining each one.
+2.  If `context/spec/knowledgebase/structure.md` exists, read it. Summarize what already exists when presenting context to the user — this helps frame features relative to the current system.
+3.  Walk the user through the sections of the template, explaining each one.
     - **Project Name & Vision:** Ask for the project's name and its core purpose.
     - **Target Audience & Personas:** Ask who the product is for and help create one simple persona.
     - **Success Metrics:** Ask how they will measure the product's impact on the user.
     - **Core Features & User Journey:** Ask for the 3-5 most important high-level features and a simple user workflow.
     - **Project Boundaries:** Ask what is essential for the first version (In-Scope) and what can wait (Out-of-Scope).
-3.  Once all sections are complete, proceed to **Step 3: File Generation**.
+4.  Once all sections are complete, proceed to **Step 3: File Generation**.
 
 ---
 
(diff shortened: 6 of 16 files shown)


commands/scan.md
```diff
@@ -0,0 +1,167 @@
+---
```
AlexanderMakarov, line 1: **Blocker — design diverges from roadmap.**

[AWOS Roadmap](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS_tuWz1ceN) items r3 and r4 explicitly call for *modifying* `/awos:product`, `/awos:architecture`, and `/awos:hire` to detect existing source files, pre-fill answers, and ask only deltas — plus a `docs/brownfield.md` walkthrough describing **Update Mode behaviour for those existing commands**.

This PR instead introduces a parallel `/awos:scan` command that materializes three new context documents (`stack.md`, `structure.md`, `conventions.md`) and splits the canonical flow into separate brownfield and greenfield paths. That's a different design from what r3/r4 specify, with knock-on cost: every downstream command now needs explicit "if scan docs exist" branches (see the changes in `architecture.md`, `hire.md`, `implement.md`, `product.md`, `roadmap.md`, `spec.md`, `tasks.md`, `tech.md` — eight files in this PR alone). The existing commands also already have **Update Mode** logic that could host the "read what's already there and pre-fill" behaviour without adding new state on disk.

So I would recommend either:

1. **Refactor to match r4** — move the codebase introspection into `/awos:product` (detect language/framework/structure from `package.json`/`pyproject.toml`/etc., pre-fill the product definition), `/awos:architecture` (Update Mode reads the actual stack rather than asking), and `/awos:hire` (already extracts technologies — extend to read package manifests when architecture is thin). No new command, no new on-disk format. This is what r4 describes.
2. **Keep `/awos:scan` but narrow scope** — ship only `stack.md` (the one document whose downstream consumers actually exist in this PR), defer `structure.md` and `conventions.md` until their consumers are designed. Reduces the new contract surface.
3. **Move to a plugin** — if there's a reason the parallel-command approach is better than r4, document it and create a plugin to support existing AWOS flow, not alter it. See https://provectus.slack.com/archives/C09GCR80NC8/p1779863788301979 for the similar case.

CLAUDE.md
```diff
@@ -92,11 +92,16 @@ Each file in `claude/commands/{name}.md` is a tiny wrapper that points at `.awos
 AWOS is **spec-driven** — all project state lives in markdown files under `context/` in the user's project, not in chat history. An AI agent can rehydrate full context by reading the files alone. The canonical flow (each command is a markdown prompt under `commands/`):
 
 ```
-/awos:product → /awos:roadmap → /awos:architecture → /awos:hire
-              → /awos:spec → /awos:tech → /awos:tasks → /awos:implement → /awos:verify
+Brownfield:  /awos:scan → /awos:product → /awos:roadmap → /awos:architecture → /awos:hire
```
AlexanderMakarov: Two issues on this block:

1. **`/awos:archive` is referenced but doesn't exist in this PR.** Cited here (both flow lines) and on line 100 ("`/awos:archive` closes the loop after verification by archiving the spec and updating scan docs with learnings"). No `commands/archive.md`, no `claude/commands/archive.md`, no `docs/commands/archive.md`. Hanging contract — future contributors will assume the command exists and is documented elsewhere. Drop the archive references from this PR until the archive command ships, or split archive into its own change.
2. **Brownfield and Greenfield flow lines duplicate 90% of each other.** The only difference is the `/awos:scan` prefix on brownfield. Collapse to one canonical flow plus a one-line preamble — e.g.:

```
/awos:product → /awos:roadmap → /awos:architecture → /awos:hire
              → /awos:spec → /awos:tech → /awos:tasks → /awos:implement → /awos:verify
```

> Brownfield projects prepend `/awos:scan` to capture the existing stack/structure/conventions before product setup.

Less noise, same information.

claude/commands/scan.md
```diff
@@ -0,0 +1,5 @@
+---
+description: Scans an existing codebase — stack, structure, conventions.
```
AlexanderMakarov: Frontmatter is too thin compared to the other wrappers:

- **Description names inputs, not outputs.** Compare `hire.md` ("Hires specialist agents — finds, installs skills, MCPs, and agents from registry, generates agent files.") — names what the command produces. Here the description just says what the command looks at.
- **Missing `argument-hint`.** `commands/scan.md` accepts `<user_prompt>$ARGUMENTS</user_prompt>` for scoped scans (lines 19, 64, 218), so users can pass an argument — but the wrapper doesn't advertise it. Every other wrapper with an optional argument declares one (`product.md`, `architecture.md`, `hire.md`).

Suggested:

```yaml
---
description: Scans existing codebase — produces context/codebase/{stack,structure,conventions}.md.
argument-hint: '[scope path, optional]'
---
```

plugins/awos/skills/scan/SKILL.md
```diff
@@ -0,0 +1,167 @@
...
+### Step 1: Mode Detection
+
+Check if all three files exist under `context/codebase/`.
+
+- If **all three exist**, proceed to **Step 2A: Update Mode**.
+- If **any are missing**, proceed to **Step 2B: Creation Mode**.
+
+---
+
+### Step 2A: Update Mode
```
AlexanderMakarov: **"Targeted rescan" merge semantics are underspecified and contradict Step 4.**

This section promises that a targeted rescan will "merge agent findings into the existing documents rather than overwriting them wholesale" and "preserve sections unrelated to the target area" (line 257). But Step 4 (lines 79–146) instructs each Explore agent to return a **full filled-in version of the template** — there's no instruction to constrain output to the target area, no rule for how the orchestrator should merge sections, and no specification of what wins on conflict (last-write? user-confirm? section-by-section diff?).

In practice the agents will return three full documents and Step 5 will overwrite the existing files — exactly what this section says it won't do.

Two options:

1. **Drop targeted rescan from this PR** until it's designed. Update Mode becomes full-rescan-only.
2. **Specify the merge protocol.** Pass each existing doc to its Explore agent, instruct the agent to return *only* sections that need updating, and have the orchestrator merge by section heading (replace matching headings, keep the rest).

commands/scan.md
```diff
@@ -0,0 +1,167 @@
...
+- Known concerns and tech debt (deprecated patterns, TODO/FIXME/HACK comments, fragile areas, performance issues)
+
+For conventions, be prescriptive: say 'use X pattern' not just 'X pattern exists'. New code should follow these conventions.
+
+Format your response as a filled-in version of this template:
+
+[conventions-template content here]
+")
+```
+
```
AlexanderMakarov, line 117: **Template substitution mechanism is fragile.**

The agent prompts above (lines 88–141) use literal placeholders like `[stack-template content here]` and this line tells the orchestrator to interpolate template content there. Two failure modes:

1. **Bracket collision.** The templates themselves already use `[bracketed]` placeholders (e.g. `templates/stack-template.md` line 6: `**[Language]:** [Version] — [Purpose]`). When the orchestrator substitutes them in, an LLM has no syntactic boundary between "this is the substitution marker" and "this is template content".
2. **Fence nesting.** The agent prompts are wrapped in ```` ```text ```` fences. If a future template ever contains a fenced code block of its own (currently they don't, but `structure-template.md` has a placeholder for a directory tree that's likely to attract fences), the outer fence breaks.

Suggested fix: switch to an unambiguous sentinel, e.g. `<template>...</template>` or `<<<STACK_TEMPLATE>>>...<<<END>>>`, and have the orchestrator literally interpolate the file content between them. Reflect that in the example prompts above so the convention is visible.

Minor: the `Agent(subagent_type="Explore", ...)` snippets look like callable code but they're prose examples. A reader might lift one verbatim — consider a leading sentence clarifying these are illustrative.

commands/spec.md
```diff
@@ -30,6 +30,7 @@ Your primary task is to create a new functional specification file. You will det
 - **Template File:** `.awos/templates/functional-spec-template.md`.
 - **Context File 1:** `context/product/product-definition.md`.
 - **Context File 2:** `context/product/roadmap.md`.
+- **Codebase Scan (Optional):** `context/codebase/structure.md` — if present, provides awareness of existing features and capabilities so the spec does not re-specify what already exists. Only `structure.md` is used; `stack.md` and `conventions.md` are technical and not relevant to a functional spec.
```
AlexanderMakarov: The trailing justification — "Only `structure.md` is used; `stack.md` and `conventions.md` are technical and not relevant to a functional spec" — belongs somewhere else.

Inputs blocks describe contracts, not rationale. Every other input row in this file (and across the other commands) names a file and its purpose in one short clause. This one explains *why two sibling files are excluded*, which is information about scan's design, not spec's. A reader of `spec.md` shouldn't have to learn the layout of the codebase-doc family.

Move the rationale to `commands/scan.md` (or `docs/commands/scan.md` once it exists) under a "Which downstream commands read which docs" section, and shorten the input here to:

> **Codebase Scan (Optional):** `context/codebase/structure.md` — if present, provides awareness of existing features and capabilities.

commands/architecture.md
```diff
@@ -19,6 +19,7 @@ Your task is to manage the architecture file located at `context/product/archite
 - **Template File:** `.awos/templates/architecture-template.md` (The required structure).
 - **Prerequisite Input 1:** `context/product/product-definition.md` (The "what" and "why").
 - **Prerequisite Input 2:** `context/product/roadmap.md` (The implementation phases).
+- **Codebase Scan (Optional):** `context/codebase/stack.md` and `context/codebase/structure.md` — if present, those documents own the technology inventory. The architecture document focuses on decisions, infrastructure, security, observability, and service topology instead of re-enumerating the stack.
```
AlexanderMakarov: Good split here — `stack.md` owns the technology inventory, `architecture.md` owns decisions / infra / security / observability / services-diagram. But the boundary is invisible outside this file:

- `claude/commands/architecture.md` wrapper description still reads `Defines the System Architecture — stack, DBs, infra.` — exactly what scan now owns. The wrapper should reflect the new split.
- The README Step-2 table uses the same stale description.
- No user-facing doc explains *why* `architecture.md` looks lighter in brownfield mode — a user reading it for the first time after running scan will wonder why the stack section is a one-line pointer.

Suggested:

1. Update `claude/commands/architecture.md` description to something like `Documents architecture decisions, infra, security, observability — defers stack inventory to scan when present.`
2. Mirror that change in the README architecture row.
3. Add a "Relationship to `/awos:scan`" note to the architecture command doc (`docs/commands/architecture.md`) and the scan doc once it exists.

AlexanderMakarov (changes requested):
PR adds a new `/awos:scan` command, but [the roadmap](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS_tuWz1ceN) (r3/r4) said to teach the existing commands to be brownfield-aware instead. The chosen design splits the workflow into two paths and pushes "if scan docs exist" branches into eight commands — extra surface area the roadmap didn't ask for. On top of that, the user-facing layer is incomplete: README untouched, no `docs/commands/scan.md`, and `CLAUDE.md` references an `/awos:archive` command that doesn't ship in this PR. Worth a design conversation before merging.

commands/scan.md
dustyo-O: yes, this command explicitly adds new files, this is not a bug but a feature ->

1) I do not agree with the idea move this awos:scan tasks to awos:product. That command has completely different scope and goals, it's better keep it focused on what it currently does
2) I already reduced count of produced files comparing to gsd-map-codebase command, why I keep three files -> it is easier to keep each file in the limit of 500 lines, more control on which context we are really need for particular task (save tokens)
3) that is not a plugin, that is different flow, I mean it to be required for brownfield

commands/product.md
```diff
@@ -63,13 +64,14 @@ First, check if the file `context/product/product-definition.md` exists.
 ### Step 2B: Creation Mode
 
 1.  If `<user_prompt>` is non-empty, briefly note that you'll use it as a starting point, then refine from there.
-2.  Walk the user through the sections of the template, explaining each one.
+2.  **Brownfield check:** If `context/codebase/stack.md` and `context/codebase/structure.md` exist, read them in parallel. Summarize what the existing system is (e.g., "This is a React + Express app with a PostgreSQL database") before starting the interview. Frame questions around what the user is building on top of the existing system, not from scratch.
```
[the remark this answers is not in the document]
dustyo-O: nope, if only one file exists, this represents broken state, which can't happen if workflow worked correctly, so no change needed here

commands/architecture.md
```diff
@@ -49,12 +50,19 @@ Follow this logic precisely.
...
+      - **Infrastructure & Deployment** — cloud provider, hosting, CI/CD pipelines, environment topology, container strategy. This information lives outside the codebase and is not covered by scan.
+      - **Network & Security** — authentication model, authorization scheme, network topology, secrets management.
+      - **Observability & Monitoring** — logging, metrics, alerting, and tracing strategy.
+      - **Services Diagram** — how deployed components interact at runtime (service boundaries, data flows between services, external integration points).
+      - Only propose **new** technologies when the roadmap requires capabilities the current stack does not cover. Frame questions as "here is what you have — what needs to change or be added?"
+    - **If scan docs do not exist (greenfield):** The architecture document is the first and only technology record. Work through the template section by section, proposing concrete technology choices with alternatives for each area.
```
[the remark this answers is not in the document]
dustyo-O: nope, it should

plugins/awos/skills/scan/SKILL.md
dustyo-O: ✅

commands/scan.md
dustyo-O: we already use this brackets many times. better if ai reviews would be aware of current coding patterns

image

commands/spec.md
dustyo-O: ✅

commands/architecture.md
dustyo-O: ✅

commands/scan.md
AlexanderMakarov: Adding new artifact files it is OK. Don't need to reduce them. While my concern is about adding a new command.

`gsd-map-codebase` is a command from project which has 62 commands in total - this is quite hard to navigate. AWOS is deliberately wants to stay as simple as possible and it means as fewer commands as possible. Moreover [GSD author these weekends executed crypto scam with this project](https://www.reddit.com/r/ClaudeAI/comments/1tktl4w/if_you_use_the_get_shit_done_gsd_ai_tool_you_need/) so it has quite bad reputation now (BTW it is better to remove original GSD and switch on community-driven fork).

> that is not a plugin, that is different flow

Plugin (BTW it is inside this repo, not some separate codebase) is an exact place for a **different flow**. Please look at https://provectus.slack.com/archives/C09GCR80NC8/p1779863788301979.

commands/scan.md
AlexanderMakarov: I agree that it is used in AWOS, but it is not the best practice/example to follow. There is a [roadmap item to migrate on XML tags](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS_tuWz1ceN/r32) which is recommended by[ Anthropic best practices](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/claude-prompting-best-practices#structure-prompts-with-xml-tags).

commands/scan.md
dustyo-O: nope, I disagree. this is not plugin, this is adjustment of the main flow for brownfield projects which was not yet covered. that's a very basic thing which is not for "different" flow, that's a must-have feature for a quite common field.

@workshur  your thoughts on this?

workshur: I agree that it functions more like a plugin, allowing us to extend the flow by adding a preparation step to gather information before running the `/awos:product`. However, I disagree with including it as part of `/awos:product`.

workshur:
Thanks for putting this together — it's cleanly built. Before going deeper on the implementation, I'd love to step back with you on whether scan is the right shape for the problem, because I'm not fully convinced yet. A few honest questions:

**1. What does scan solve that on-demand `Explore` in `architecture`/`tech` doesn't?**
Stack and structure are cheap to rediscover anytime, always fresh — whereas scan docs go stale and need a manual rescan. Could this be a small adjustment to `architecture` (letting it take "current state" as input) rather than a new command plus three durable docs?

**2. The extraction axes feel a bit misaligned with where the value is.**
The downstream benefits you describe — `roadmap` marking features "already built", `spec` avoiding re-specifying existing functionality — are all about *what the product already does*. But `structure.md` captures directory layout and module boundaries; a folder tree doesn't really tell you which features are implemented. For a platform built over years, that existing behavior is the genuinely hard thing to reconstruct — not the stack (easy) or conventions (often absent). If capturing existing product capabilities is the real brownfield need, should the artifact be a feature/capability inventory instead?

**3. `conventions.md` somewhat inverts how conventions flow in AWOS.**
In greenfield there's no conventions doc and it works fine, because conventions are injected *forward* via the skill layer (`/awos:hire` + specialists). Here we extract them *backward* from existing code and tell `implement`/`tasks` to conform — which, in a brownfield codebase where a lot is exactly what you'd want to refactor away, risks cementing tech debt. Could green and brownfield share one path (skills, forward), with a "propose and let the user choose" step when conventions need to be *established* rather than recorded as-is?

**4. Stack ownership.** Moving the inventory out of `architecture` into `stack.md` creates two sources of truth that will drift. Why not keep `architecture` as the single owner?

None of this is a knock on the execution — just want to make sure we're solving the right problem in the right form before it lands. My gut: `structure.md` as navigation context is defensible, `stack.md` mostly overlaps `architecture`, and `conventions.md` is the one I'd push back on hardest. Keen to hear your thinking!

commands/scan.md
dustyo-O: @AlexanderMakarov okay I moved command to plugin. what confuses me in current setup -> main commands are aware of ./context/codebase warehouse, but it will always be empty unless one of the plugins which uses it is installed (currently only scan, next command will be archive).

but if we remove mentions of that source (./context.codebase) -> those md's will be completely dead end in the project no one is aware of. So currently I left thing as is, comment if you have any new ideas about that

dustyo-O:
@workshur

1) The problem I'm trying to solve: Claude (like any human) often overlooks relevant project context while developing a new feature. This can happen with both smaller details and bigger things — from small shared components to large features whose implementation could be reused. If the driver isn't aware of how we should design the feature, or if there's no deep research of the codebase before implementation, there will likely be a lot of copy-paste, bad abstractions, etc. If the project starts from greenfield, this problem is partially solved by the existing specs. If no spec is available (or the specs have already been deleted), the quality of the generated code drops. I'm trying to create this source of truth.

Could this be a change to `architecture`? I don't rule it out, but in my head the plan was to create the place where specs are archived, and where the initial archive is created after scanning the brownfield.

2) Sorry, I didn't quite get the issue. The generated documents could partly overlap with `architecture.md` (which I mention and actively avoid in my command), but the rest of the content is completely different from what the framework already generates.

3)
> (`/awos:hire` + specialists)

Yes, but we HAVE to add some custom documentation to `CLAUDE.md` whenever we want project-specific instructions. This skill aims to extract that custom knowledge from the codebase.

> risks cementing tech debt

There's no problem with following existing patterns EVEN when they're already legacy that's planned for a refactor:
- if the refactor is planned to change the old pattern in one go, it's better to keep following the old code style, because then all occurrences of the legacy code get fixed at once;
- if the refactor is planned to be step-by-step, and new code should follow the new pattern, it's EVEN BETTER to have this convention documented somewhere, to keep Claude from getting confused about which pattern to follow.

4) This one is the most debatable, indeed. I prefer to think of `architecture` as a view of the planet from space — a description of the basic principles the project was built on. The codebase context is the living thing — a summary of how the specs actually shaped the real implementation of the project.

commands/scan.md
AlexanderMakarov: The dead-end risk you describe is the signal to split this into two PRs rather than to pick a folder.

Right now the PR teaches nine core commands to read `context/codebase/*.md` **and** ships the only thing that writes them (the scan skill, folded into the audit plugin). That bundling is what makes the plugin feel mandatory and creates the "aware of a warehouse that's always empty" smell.

Suggested split:

- **PR A — core only.** Define `context/codebase/{stack,structure,conventions}.md` as a *producer-agnostic contract*: file names, shape (the templates), and which command reads which doc. Wire the nine `if exists` reads, document it in `CLAUDE.md`, and add a nudge when the files are absent on a non-empty repo. Now the `md`'s aren't a dead end — they're a documented input anything can populate (scan, a future `archive`, a manual write, CI).
- **PR B — producer only.** The scan skill as one producer of that contract.

On PR B, I'd also reconsider folding scan into the existing **audit** plugin. Audit and scan share nothing downstream — audit is a read-only scored report in `context/audits/` for a human; scan produces workflow input for core commands. Folding them couples two lifecycles under one version and one vague description ("codebase scanning and code quality audits"), and the `keywords` still only say `audit`/`code-quality`/`dimensions`. Cleaner end-states: scan → core (best), or scan as its own plugin (if it must stay a plugin) with audit left untouched.

Net: core depends on a *format*, not a *plugin*; the plugin stays opt-in by install; and the audit plugin keeps a single purpose.

plugins/awos/skills/scan/SKILL.md
```diff
@@ -0,0 +1,162 @@
...
+
+1. **User Prompt (Optional):** `$ARGUMENTS`. Can scope the scan to a specific area (e.g., `src/api`) or request a full scan.
+2. **Templates:** `templates/stack-template.md`, `templates/structure-template.md`, `templates/conventions-template.md` (relative to this SKILL.md).
+3. **Existing Scan Docs (Optional):** `context/codebase/stack.md`, `context/codebase/structure.md`, `context/codebase/conventions.md` — if present, triggers Update Mode.
+
+## Outputs
+
+1. **`context/codebase/stack.md`:** Technology stack — languages, frameworks, dependencies, external services, build tools.
+2. **`context/codebase/structure.md`:** Project structure — directory layout, module boundaries, architectural patterns, data flow, file placement rules.
+3. **`context/codebase/conventions.md`:** Codebase conventions — naming, code organization, error handling, testing patterns, known concerns.
```
AlexanderMakarov: Since this is where the three artifacts are decided, flagging the summary's question here so it isn't buried: of the three, `structure.md` is the only genuinely new one — `stack.md` overlaps `architecture.md`, and `conventions.md` runs opposite to AWOS's forward convention flow.

And it might be worth a look at `structure.md` too (@workshur's [point](https://github.com/provectus/awos/pull/130#issuecomment-4602221514)): the value claimed for it — `roadmap` knowing what's "already built", `spec` not re-specifying — is about *what the product does*, but a layout tree doesn't show which features exist. A **capability/feature inventory** might serve those consumers better. Open question, not a line-level fix.

tests/lint-prompts.test.js
```diff
@@ -208,8 +208,9 @@ test('all /awos:<name> cross-references resolve', () => {
   const rootCommands = new Set(
     listMarkdown(commandsDir).map((f) => '/awos:' + f.replace(/\.md$/, ''))
   );
-  // Plugin-provided commands (the audit plugin contributes /awos:ai-readiness-audit).
+  // Plugin-provided commands.
   rootCommands.add('/awos:ai-readiness-audit');
```
AlexanderMakarov, line 212: Nice to see `/awos:scan` registered here. If it's easy, a few more assertions would cover the new contracts and match the `CLAUDE.md` test policy ("a structural contract ships its test in the same PR"):

- skill frontmatter on `plugins/awos/skills/scan/SKILL.md` has `name` / `description` / `argument-hint`;
- the three `plugins/awos/skills/scan/templates/*.md` files exist;
- every `context/codebase/*.md` path referenced in a prompt matches one of the real outputs — catches typo drift like `convetions.md` before QA does.

If the restructure above happens, the path-consistency check fits the core PR and the others fit the producer PR.

AlexanderMakarov (commented):
Thanks for the iterations — most of the surface-level feedback has landed. My remaining concern is structural, and it's the thing I'd like resolved before merge.

### The shape: a new command/plugin, where the flow should grow instead

The PR introduces `scan` as a standalone producer plus a new `context/codebase/*` artifact family that nine core commands read. Roadmap [r4](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS_tuWz1ceN) calls for the opposite: fold codebase-awareness into the *existing* `/awos:product` and `/awos:architecture` commands. As shipped, the PR also bundles two separable things — (1) core support for the new `context/codebase/*` artifacts, and (2) a producer that writes them — which is what makes the plugin effectively mandatory even though the reads are nominally optional.

### The artifacts overlap mechanisms AWOS already has

This is @workshur's point in [this comment](https://github.com/provectus/awos/pull/130#issuecomment-4602221514), and I think it's the crux. Two of the three scan artifacts duplicate existing AWOS mechanisms:

- **`stack.md`** overlaps what `architecture.md` already owns — two sources of truth for the technology inventory that will drift. (The author already called stack ownership "the most debatable.")
- **`conventions.md`** inverts how conventions flow. AWOS injects them *forward* — `/awos:hire` installs skills, specialists carry them, `implement` loads them; greenfield has no conventions doc and works fine. `conventions.md` runs *backward*, extracting conventions out of existing code and telling `implement`/`tasks` to conform. That's a second, brownfield-only convention mechanism — and in a codebase where a lot is exactly what you'd want to refactor away, it risks cementing tech debt rather than improving it.
- **`structure.md`** is the one genuinely new artifact — directory layout / navigation context the framework doesn't capture elsewhere. No overlap.

So the three shouldn't be adopted wholesale; each needs a deliberate keep-or-fold decision.

### How I'd do this properly

Split into two PRs and scope the artifact set:

1. **Core PR — define the `context/codebase/*` contract, producer-agnostic.** Names, shape (templates), and which command reads which doc, wired into the existing commands and documented in CLAUDE.md, with a nudge when files are absent on a brownfield repo. Crucially, **scope the artifacts** rather than taking all three:
   - `structure.md` — keep; new navigation info.
   - `stack.md` — keep only if `architecture` formally defers stack ownership to it (resolve the two-sources question first), otherwise drop and let architecture stay the single owner.
   - `conventions.md` — route into the *forward* layer instead: a step that **proposes** observed conventions for the user to fold into skills / specialists / CLAUDE.md, rather than a backward doc `implement` blindly conforms to. If it stays a doc, make it canonical for **both** greenfield and brownfield so there's one mechanism, not two.
2. **Producer PR — the detection/pre-fill behavior.** Per r4, fold this into `/awos:product` and `/awos:architecture` (where brownfield context belongs in the flow) rather than a parallel command. If a separate producer is still wanted, it ships as its own plugin — not folded into the audit plugin (reasoning in the [`context/codebase` thread](https://github.com/provectus/awos/pull/130#discussion_r3341620033)).

Net: core depends on a *format*, not a *plugin*; the artifact set carries only what AWOS doesn't already own; and conventions keep a single, forward-flowing mechanism.

Inline comments below cover two smaller items (Update Mode agent contract, lint coverage).

commands/scan.md
```diff
@@ -0,0 +1,134 @@
...
+---
+
+# INPUTS & OUTPUTS
+
+- **User Prompt (Optional):** <user_prompt>$ARGUMENTS</user_prompt>. Can scope the scan to a specific area (e.g., `src/api`) or request a full scan.
+- **Template 1:** `.awos/templates/structure-template.md`.
+- **Template 2:** `.awos/templates/decisions-template.md`.
+- **Existing Scan Docs (Optional):** `context/spec/knowledgebase/structure.md`, `context/spec/knowledgebase/decisions.md` — if present, triggers Update Mode.
+- **Output 1:** `context/spec/knowledgebase/structure.md` — project structure (directory layout, module boundaries, architectural patterns, data flow, file placement rules).
+- **Output 2:** `context/spec/knowledgebase/decisions.md` — non-standard project decisions (architecture patterns and why chosen, data flow boundaries, error handling strategy, testing strategy, integration points, known constraints).
```
dustyo-O, line 24: we should explicitly point out, that decisions.md content should contain something that goes beyond the typical hired agents behavior, expands or overwrites it

commands/scan.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+**Agent 2 — Decisions Analysis:**
+
+```text
+Agent(subagent_type="Explore", description="Analyze project decisions", prompt="
+Explore this codebase and document the non-standard decisions that shaped it. Focus on WHY things were done a certain way, not just what exists. Be thorough and path-specific.
+
+Analyze:
+- Architecture patterns (what patterns were chosen and what constraints drove each decision — cite config files, directory structures, framework choices)
```
dustyo-O, line 103: same here, if hired agents already exists, avoid documenting their default behavior

CLAUDE.md
```diff
@@ -89,14 +89,18 @@ Each file in `claude/commands/{name}.md` is a tiny wrapper that points at `.awos
 
 ## Architecture: Document-Centric Workflow
 
-AWOS is **spec-driven** — all project state lives in markdown files under `context/` in the user's project, not in chat history. An AI agent can rehydrate full context by reading the files alone. The canonical flow (each command is a markdown prompt under `commands/`):
+AWOS is **spec-driven** — all project state lives in markdown files under `context/` in the user's project, not in chat history. An AI agent can rehydrate full context by reading the files alone.
+
+**Brownfield projects** can populate `context/spec/knowledgebase/` with two documents that give downstream commands awareness of the existing codebase: `structure.md` (directory layout, module boundaries, data flow) and `decisions.md` (non-standard project decisions that override or extend default agent behavior). These files are producer-agnostic — they can be populated by `/awos:scan` (codebase scan), `/awos:archive` (post-implementation learnings), or manually.
```
AlexanderMakarov, line 94: **Placement: project-wide knowledge inside the per-feature spec namespace.**

`context/spec/` is the indexed per-feature list — numbered directories documenting specific feature implementations, allocated by `create-spec-directory.sh`. A `knowledgebase/` folder (and `feature-log.md`) gets lost inside that list, and it's the only non-numbered entry in a namespace whose whole convention is `NNN-feature/`.

Project-stable knowledge already has a home: `context/product/` — product-definition, roadmap, and architecture all live there. `structure.md`, `decisions.md`, and `feature-log.md` are exactly that kind of document: project-level, stable, updated across features. I'd move all three under `context/product/`.

Separately, this block frames the knowledgebase as a brownfield feature ("Brownfield projects can populate…"), but `/awos:archive` makes it valuable on greenfield too — every archived feature feeds it. Worth reframing as a project-lifecycle artifact that brownfield projects can additionally seed via `/awos:scan`.

commands/spec.md
```diff
@@ -30,6 +30,7 @@ Your primary task is to create a new functional specification file. You will det
 - **Template File:** `.awos/templates/functional-spec-template.md`.
 - **Context File 1:** `context/product/product-definition.md`.
 - **Context File 2:** `context/product/roadmap.md`.
+- **Knowledgebase (Optional):** `context/spec/knowledgebase/structure.md` — if present, provides awareness of existing features and capabilities.
```
AlexanderMakarov, line 33: **Dead input + wrong description.**

Two problems with this declaration:

1. **No PROCESS step reads it.** Step 2 reads only `product-definition.md` and `roadmap.md` — every other command that declares a knowledgebase input got a matching process line ("If … exists, read it too"); spec is the one place the declaration is dead.
2. **The description doesn't match the file.** "Awareness of existing features and capabilities" — but `structure.md` holds directory layout, module boundaries, and file placement rules. The capability inventory is `feature-log.md` (per `archive.md`), which nothing reads. @workshur's point from the design thread is still unanswered: nothing in the artifact set that commands actually consume captures *what the product already does*.

If the input stays, wire Step 2 to read it (and arguably `feature-log.md`, which is what this description is reaching for) and fix the description to match what's actually read.

This line is also the worst instance of a PR-wide consistency problem (summary point 3): `structure.md` is described three different ways across the seven commands that mention it — "awareness of the existing codebase" (product), "for brownfield planning" (roadmap), "existing features and capabilities" (here). One canonical INPUTS phrasing and one canonical PROCESS phrasing, stamped identically everywhere, would prevent this drift.

commands/archive.md
```diff
@@ -0,0 +1,147 @@
...
+---
+
+# INPUTS & OUTPUTS
+
+- **User Prompt (Optional):** <user_prompt>$ARGUMENTS</user_prompt>
+- **Primary Context:** A spec directory under `context/spec/` containing:
+  - `functional-spec.md`
+  - `technical-considerations.md`
+  - `tasks.md`
+- **Knowledgebase (Optional):** `context/spec/knowledgebase/structure.md` and `context/spec/knowledgebase/decisions.md` — updated with learnings from the archived feature. If absent, knowledgebase updates are skipped.
```
AlexanderMakarov, line 24: **Greenfield gap: learnings dropped exactly where the mechanism was built to preserve them.**

"If absent, knowledgebase updates are skipped" + Step 7 deletes the spec directory anyway = on a greenfield project (no `/awos:scan` run, so no knowledgebase docs exist), archive extracts nothing and destroys the spec. But the stated motivation for this whole mechanism is that quality drops once specs are deleted — and that problem is identical on greenfield.

Fix: create the knowledgebase docs from the templates when missing, the same way `feature-log.md` is already created when missing (line 110). That also collapses the brownfield/greenfield split into one mechanism — archive becomes a producer that works everywhere, which is what "producer-agnostic" should mean.

commands/scan.md
```diff
@@ -0,0 +1,134 @@
...
+
+- If **both exist**, proceed to **Step 2A: Update Mode**.
+- If **either is missing**, proceed to **Step 2B: Creation Mode**.
+
+---
+
+## Step 2A: Update Mode
+
+1. Read the existing `structure.md` and `decisions.md` from `context/spec/knowledgebase/` in parallel.
+2. Tell the user you found existing scan documents. Update Mode performs a full rescan — the existing documents are overwritten entirely.
```
AlexanderMakarov, line 50: **Producer conflict: Update Mode's full overwrite destroys what archive carefully merges.**

`/awos:archive` merges into these files with deliberate care — dedupe rules, the 500-line budget, summarize-don't-trim. Its entries come from spec content and git archaeology: the "why" behind decisions, constraints introduced by specific features. A fresh code scan can't re-derive any of that — an `Explore` agent won't rediscover "constraint introduced by feature 007" from the code alone. So one rescan after a few archived features wipes exactly the knowledge this PR exists to preserve.

Step 2A.1 even reads the existing docs first — and then nothing uses them; the Step 4 agent prompts don't include them, so the read is dead weight.

Two producers writing the same files need shared merge semantics, not last-writer-wins. Cheapest fix that uses what's already read: overwrite `structure.md` only (it *is* rediscoverable from code), and for `decisions.md` pass the existing content into the agent prompt as prior context with an instruction to preserve entries still supported by the code. If full overwrite is intentional, at minimum warn that archive-contributed learnings will be lost and confirm before proceeding.

commands/archive.md
```diff
@@ -0,0 +1,147 @@
...
+- Trim overly verbose descriptions to their essential content.
+
+### Step 5: Update Feature Log
+
+Append an entry to `context/spec/feature-log.md`. If the file does not exist, create it with the following header:
+
+```markdown
+# Feature Log
+
+Archived features and the capabilities they introduced. Read by `/awos:spec` and `/awos:tech` to find patterns available for reuse.
```
AlexanderMakarov, line 107: **False consumer claim.** This header tells future readers the log is "Read by `/awos:spec` and `/awos:tech`", but neither command references `context/spec/feature-log.md` anywhere in this PR — the claim is aspirational. Either wire the reads (spec is arguably the right consumer: this log is the "what does the product already do" inventory, which is what spec's knowledgebase input description is reaching for), or trim the header to describe what the log *is* rather than who reads it. As-is, archive writes a document nothing consumes.

README.md
```diff
@@ -21,32 +21,34 @@ npx @provectusinc/awos
...
-| `/awos:product`      | Defines the Product — what, why, and for who.                                     | [Details](docs/commands/product.md)      |
-| `/awos:roadmap`      | Builds the Product Roadmap — features and their order.                            | [Details](docs/commands/roadmap.md)      |
-| `/awos:architecture` | Defines the System Architecture — stack, DBs, infra.                              | [Details](docs/commands/architecture.md) |
-| `/awos:hire`         | Hires specialist agents — finds, installs skills and MCPs, generates agent files. | [Details](docs/commands/hire.md)         |
+| Command              | What it does                                                                                  | Docs                                     |
+| -------------------- | --------------------------------------------------------------------------------------------- | ---------------------------------------- |
+| `/awos:scan`         | Scans existing codebase — produces structure and decisions documents for brownfield projects. | [Details](docs/commands/scan.md)         |
```
AlexanderMakarov, line 32: **Two broken links.** This Details link points at `docs/commands/scan.md`, and the `/awos:archive` row (line 51) at `docs/commands/archive.md` — neither file exists on this branch, so the README ships two 404s. Every other row in these tables resolves to a real page under `docs/commands/`. Needs the two docs pages (they're short; `docs/commands/verify.md` shows the shape) or link-less rows until they exist.

commands/scan.md
AlexanderMakarov: The rebuild resolves part of this thread: `stack.md` is gone with `architecture.md` pinned as sole stack authority, conventions narrowed to `decisions.md` (non-standard deltas only), and the knowledgebase contract is producer-agnostic in core rather than plugin-coupled. That's real progress — thanks.

What remains open is the [r4](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS-dont-reorder-to-support-hyperlinks-just-cle_tuWz1ceN/r4) divergence itself, and the rebuild moved in the opposite direction: r4 asks to fold brownfield awareness into the existing commands; round 1 shipped one new command, this round ships two. `/awos:archive` in particular is a separable feature with its own risk profile — it deletes spec directories and introduces a new `feature-log.md` contract — bundled into a brownfield PR without its own roadmap item or review surface. Whatever we settle on scan's shape, archive deserves its own PR.

Details in my round-3 review summary (points 1–3), including two further architectural notes: `context/spec/knowledgebase/` placement (project-stable docs belong in `context/product/`, not the indexed per-feature namespace) and the seven-command "if exists" conditional wiring that keeps this a bolt-on side-channel rather than an integration through the documents commands already read.

commands/scan.md
AlexanderMakarov: Agree with both of these self-notes (here and on line 103) — and I'd treat them as merge-blocking rather than follow-up, because this rule defines what `decisions.md` *is*: a record of deltas against what the hired specialists and stack defaults already cover, not a survey of the codebase's practices. Without it, Agent 2 documents error-handling/testing/integration wholesale, defaults included, and we're back to the backward conventions extractor from round 2.

Concretely, the rule needs encoding in three places: Agent 2's prompt in Step 4 (read `.claude/agents/*.md` or `context/product/hired-agents.md` first, record only what differs), the `decisions-template.md` section headers, and `archive.md` Step 4 — which has the same exposure when extracting learnings from a completed feature.

CLAUDE.md
AlexanderMakarov: Both halves of this are resolved by the rebuild — `/awos:archive` ships now, and the duplicated flow lines are gone. Thanks.

One remaining point on this block moved to a new comment on the current CLAUDE.md:94: the placement of the knowledgebase under `context/spec/` and the brownfield-only framing.

claude/commands/scan.md
AlexanderMakarov: Fixed in the rebuild — the description names the outputs and the `argument-hint` is there. Matches the other wrappers now. Resolving from my side.

commands/architecture.md
```diff
@@ -49,12 +50,13 @@ Follow this logic precisely.
 ## Scenario 1: Creation Mode
 
 1.  Read and synthesize the product definition and roadmap, paying close attention to features planned for Phase 1.
-2.  Work through the template section by section — not all at once.
-    - For each architectural area, propose a concrete title from the template placeholder.
-    - For each component, propose a specific technology with one or more alternatives, justified by the project context.
+2.  **Brownfield check:** Check if `context/codebase/structure.md` exists.
+    - **If it exists (brownfield):** Read it in parallel with the prerequisite files. Note the existing modules and directory layout so you can acknowledge what already exists. The architecture document remains the sole owner of the technology inventory — work through the template section by section normally, but frame questions around what the existing system already has and what needs to change or be added.
+    - **If it does not exist (greenfield):** The architecture document is the first and only technology record. Work through the template section by section, proposing concrete technology choices with alternatives for each area.
```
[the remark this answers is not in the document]
AlexanderMakarov: The gap is real but the suggested fix targets a gate that no longer exists — the rebuild removed brownfield/greenfield gating entirely; Step 1 now just optionally enriches context. The actual problem is that `decisions.md` is declared in INPUTS (line 23) and then no process step reads it — the same dead-declaration issue as `spec.md`. The fix is extending the Step 1 read to `decisions.md` (its Known Constraints & Tech Debt section is relevant to architecture), or dropping it from INPUTS if architecture genuinely only needs `structure.md`.

commands/scan.md
AlexanderMakarov: Closing note: the `[*-template content here]` placeholder pattern carried over into the rewritten `commands/scan.md` (lines 92, 114, 118). Per our earlier exchange — not re-blocking on it; flagging so it isn't lost when roadmap r32 (XML-tag migration) lands and this file gets swept up in that pass.

AlexanderMakarov (changes requested):
This round fixes a lot from rounds 1–2. Thanks for engaging seriously! But my core concern from [the earlier review](https://github.com/provectus/awos/pull/130#pullrequestreview-4417275834) and [the design comment](https://github.com/provectus/awos/pull/130#issuecomment-4602221514) — fold brownfield awareness *into* the existing flow rather than bolt a parallel flow onto it — is **still unaddressed**, and the rewrite made it sharper:

1. **Still a separate flow, now two of them.** [Roadmap r4](https://coda.io/d/_dCoNoiEfeSz/AWOS-Roadmap_suIpbGNP#Improvements-of-AWOS-dont-reorder-to-support-hyperlinks-just-cle_tuWz1ceN/r4) asks to grow the existing commands; round 1 shipped one new command, this round ships two. `/awos:scan` is still delivered as a standalone command in the same PR, and `/awos:archive` is a separable feature with its own risk profile — it deletes spec directories and introduces a new `feature-log.md` contract. Whatever we settle on scan's shape, archive deserves its own PR and review surface.

2. **`context/spec/knowledgebase/` is the wrong home.** `context/spec/` is the indexed per-feature namespace — numbered directories documenting specific feature implementations, allocated by `create-spec-directory.sh`. A `knowledgebase/` folder gets lost inside that list, and project-stable knowledge already has a home: `context/product/` (product-definition, roadmap, architecture all live there). `structure.md`, `decisions.md`, and `feature-log.md` are project-level, stable documents — they belong under `context/product/`.

3. **The wiring confirms it's a bolt-on, not an integration.** Seven core commands now carry a guarded side-channel — `**Knowledgebase (Optional):** … if present` in INPUTS plus `If context/spec/knowledgebase/… exists, read it` in PROCESS. `architecture.md` is already the document downstream commands read for exactly this kind of context; if the knowledgebase carried weight through the existing chain, most of these conditionals wouldn't exist. And the seven mentions don't agree with each other: `structure.md` is described as "awareness of the existing codebase" (product), "for brownfield planning" (roadmap), and "existing features and capabilities" (spec — factually wrong, it holds directory layout); some steps reference files in short form (`structure.md`), others full path. If the side-channel stays, it needs one canonical INPUTS phrasing and one canonical PROCESS phrasing, stamped identically everywhere.

Mechanical gaps inline: archive drops learnings on greenfield, scan's Update Mode overwrites what archive carefully merges, `spec.md` declares an input its process never reads, `feature-log.md` claims readers it doesn't have, README links two docs that don't exist.

PR #132, opened 2026-06-03
/awos:* prompts gate the deliverable Write on AskUserQuestion approval, breaking `claude -p`

### LINKED ISSUE #131: /awos:* prompts gate the deliverable Write on AskUserQuestion approval, breaking `claude -p`
## Summary

Several `/awos:*` commands (`/awos:tasks`, `/awos:tech`, possibly more) end with an `AskUserQuestion` that asks the operator to confirm the drafted plan *before* the orchestrator persists the deliverable file. In `claude -p` (unattended) mode this question gets silently dismissed, the orchestrator's recovery text says it'll "proceed with defaults," but no `Write` call is ever issued — it just prints the draft as the final assistant text reply and ends the turn.

The downstream effect: any automated runner that drives these prompts unattended cannot rely on the deliverable being on disk after the session exits, even when the contract is otherwise satisfied (specialist invoked, inputs loaded, no hallucinations).

## Affected commands

Confirmed from awos-qa e2e session logs:

- **`/awos:tasks`** — gates on "Approve plan or request changes?" / "Plan good? Or adjust?" / "Approve to write \`tasks.md\`?" before writing `tasks.md`. Hit by:
  - [`tasks-enumerates-agents`](https://github.com/provectus/awos-qa/blob/main/e2e/scenarios/tasks-enumerates-agents) (intermittent)
  - [`tasks-discovers-plugin-agents`](https://github.com/provectus/awos-qa/blob/main/e2e/scenarios/tasks-discovers-plugin-agents) (intermittent)
  - [`tasks-emits-feature-testing-regression-slice`](https://github.com/provectus/awos-qa/blob/main/e2e/scenarios/tasks-emits-feature-testing-regression-slice) (intermittent)
- **`/awos:tech`** — gates on "Confirming high-level approach and key normalization knobs before saving" before writing `technical-considerations.md`. Hit by:
  - [`tech-invokes-specialist-subagents`](https://github.com/provectus/awos-qa/blob/main/e2e/scenarios/tech-invokes-specialist-subagents)

## Evidence — a representative `/awos:tech` trace

```
→ Agent subagent_type=python-expert      ← specialist drafted the tech spec
→ TaskUpdate
  Specialist drafted. Confirming high-level approach and key
  normalization knobs before saving.
→ AskUserQuestion ⚠ unanswered in -p
  User skipped question. Proceeding with recommended defaults
  (TEXT[], lowercase+trim+dedupe, 32/20, ...). Presenting full
  draft for final review.
[done: success]
```

All process contracts passed:

- ✓ Read functional-spec.md
- ✓ Read architecture.md
- ✓ Examined codebase via `Agent(Explore)`
- ✓ Invoked matched specialist (`python-expert`)
- ✓ No hallucinated subagent names
- ✗ **`technical-considerations.md` not written** ← the actual deliverable never lands on disk

Equivalent traces from `/awos:tasks` runs end with text like:

> "Plan drafted. Awaiting your feedback before writing `tasks.md`. Adjust/approve as you'd like."

— and the file is never written.

## Root cause hypothesis

The orchestrator prompts (likely `commands/tasks.md` and `commands/tech.md`) include a step that says something equivalent to: "before writing the file, present the plan to the user and ask for approval." The model interprets this as a gating step. In interactive mode the human types approval and the next turn calls `Write`. In `-p` mode the question is dismissed, and the orchestrator's fallback ("proceed with defaults") only updates the *narrative* — it doesn't actually call `Write` because the prompt's flow places the `Write` step strictly after the approval branch.

## Suggested fixes

Two viable shapes:

1. **Drop the approval gate.** Write the deliverable first, then notify the operator with a one-line "wrote N slices to `tasks.md`, run `/awos:tasks adjust` to revise" pointer. Aligns with the principle that destructive/expensive work asks for confirmation, but writing a draft tasks/tech file is neither destructive nor expensive — it's reversible (the file can be re-generated) and the operator typically just hits enter anyway.

2. **Keep the gate but make the no-answer path call Write.** Reword the orchestrator prompt so that when `AskUserQuestion` is dismissed or skipped, the explicit fallback is: "call `Write` with the drafted content; surface a deferred confirmation pointer." This preserves the interactive UX while making `-p` mode succeed.

Option 1 is simpler and matches what most ops teams expect from an automation-friendly CLI tool. Option 2 preserves the existing interactive convention.

## Acceptance criteria

- A `claude -p "/awos:tasks <spec>"` invocation against a valid fixture writes `tasks.md` without operator interaction.
- A `claude -p "/awos:tech <spec>"` invocation against a valid fixture writes `technical-considerations.md` without operator interaction.
- Interactive (`claude "/awos:tasks <spec>"`) sessions still surface the plan to the operator for review and adjustment — either via the Write happening first and a pointer to revise, or via an explicit revise path.
- The awos-qa scenarios listed above pass under `bun run e2e:run --all` without needing `manual.json` reclassification.

AlexanderMakarov:
Fixes #131.

## Problem

`/awos:tasks` and `/awos:tech` placed an `AskUserQuestion` approval gate *before* the deliverable `Write`. In unattended mode (`claude -p`) that question is silently dismissed: the orchestrator narrates "proceeding with defaults," but no `Write` is ever issued — the draft is printed as the final assistant text and the turn ends. Result: `tasks.md` / `technical-considerations.md` never land on disk, even when every other contract (specialist invoked, inputs read, no hallucinations) passes.

## Fix — issue Option 2 (preserve interactive UX; make the no-answer path write)

The right default is *ask first when a human is present; fall back to a documented assumption only when no one can answer.* So rather than dropping the questions, this makes a **skipped/unanswered question a signal to fall back and continue, never to stop** — and makes the deliverable `Write` unconditional.

- **Both commands** gain an `# INTERACTION` rule: a skipped or unanswered question (as in `claude -p`) is never a stop signal — record a documented assumption / default and continue through the remaining steps, including writing the deliverable.
- **`commands/tech.md`** — keeps the ask-first "assume but verify" drafting (consequential choices are still surfaced to a present user); reorders to `Draft → Write (unconditional) → Surface for review`. An unanswered per-section question is recorded as an explicit `**Assumption:**` and drafting continues.
- **`commands/tasks.md`** — the Step 3a QA-agent fallback no longer "stops and asks"; if unanswered it defaults to `general-purpose` (flagged in the Recommendations table) and continues. `File Generation` is split into `Write the Task List` (unconditional) + `Surface for Review`.

Out of scope: `/awos:spec` and `/awos:architecture` also gate writes, but their interactivity is legitimate (spec resolves `[NEEDS CLARIFICATION]` answers; architecture asks follow-ups) and they are not part of issue #131's acceptance criteria.

## Acceptance criteria

- [x] `claude -p "/awos:tasks <spec>"` writes `tasks.md` without operator interaction
- [x] `claude -p "/awos:tech <spec>"` writes `technical-considerations.md` without operator interaction
- [x] Interactive sessions still surface consequential choices for review before committing

## Testing

- `bun test tests/` — prompt-lint + installer + fixtures all pass; `prettier --check` clean
- awos-qa e2e (`bun run e2e:run`, real `claude -p` sessions) against the affected scenarios — see PR thread for the run output

## Summary by CodeRabbit

* **Documentation**
  * Generate task plans immediately without requiring approval; allow an optional top-of-file marker to skip tests
  * Continue unattended runs when questions are skipped, recording clear “Assumption:” notes instead of stopping
  * Default QA agent set to general-purpose when tester prompts are unanswered; show recommendations table when needed
  * Save and report drafts with next-step command suggestions and surfaced assumptions for review

### DIFF
--- a/commands/tasks.md
+++ b/commands/tasks.md
@@ -29,6 +29,7 @@ A **slice** is the top-level grouping checkbox — a vertical, end-to-end runnab
 # INTERACTION
 
 - Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+- **A skipped or unanswered question — as happens in an unattended `claude -p` run — is never a stop signal. Fall back to the documented default for that question and continue through the remaining steps, including writing `tasks.md`.**
 
 ---
 
@@ -85,8 +86,9 @@ Skip this step if `SKIP_TESTS = true`.
     - The built-in `general-purpose` agent as the last resort.
 2.  **If no project-specific tester or AWOS testing agent is found,** stop and ask the user via `AskUserQuestion`. Present exactly three options:
     1.  **Install a testing agent now** — run `/awos:hire` to add `testing-expert` (or a more specific tester) from the registry, then re-run `/awos:tasks`.
-    2.  **Generate the slice with `general-purpose`** — proceed and produce the Feature Testing & Regression slice, marking its tasks `**[Agent: general-purpose]**`. Flag this in the Recommendations table.
+    2.  **Generate the slice with `general-purpose`** — proceed and produce the Feature Testing & Regression slice, marking its tasks `**[Agent: general-purpose]**`. Flag this in the Recommendations table. (Default when the question is skipped.)
     3.  **Skip the Feature Testing & Regression slice** — set `SKIP_TESTS = true` for this run only; the user can re-run `/awos:tasks` later once a tester is hired.
+
 3.  **Emit the slice** using the template below. Substitute `{qa-agent}` with the agent name selected above. Substitute `N` with the next slice number. Keep the wording — downstream automations depend on this exact structure.
 
     ```md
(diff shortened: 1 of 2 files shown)


commands/tasks.md
```diff
@@ -29,6 +29,8 @@ A **slice** is the top-level grouping checkbox — a vertical, end-to-end runnab
 # INTERACTION
 
 - Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+- Ask about genuinely ambiguous or consequential choices before committing to them — surfacing a decision while a human can answer beats a silent default the user later has to undo.
```
dustyo-O: irrelevant change

commands/tasks.md
```diff
@@ -83,10 +85,13 @@ Skip this step if `SKIP_TESTS = true`.
     - A project-specific tester for the actual stack (e.g. `react-testing`, `pytest-tester`, a custom `acceptance-tester` in `.claude/agents/`).
     - A general AWOS testing agent if installed (e.g. `testing-expert` from the `awos-recruitment` registry).
     - The built-in `general-purpose` agent as the last resort.
-2.  **If no project-specific tester or AWOS testing agent is found,** stop and ask the user via `AskUserQuestion`. Present exactly three options:
+2.  **If no project-specific tester or AWOS testing agent is found,** ask the user via `AskUserQuestion`. Present exactly three options:
```
dustyo-O: I don't think this change is required

commands/tasks.md
```diff
@@ -83,10 +85,13 @@ Skip this step if `SKIP_TESTS = true`.
...
     1.  **Install a testing agent now** — run `/awos:hire` to add `testing-expert` (or a more specific tester) from the registry, then re-run `/awos:tasks`.
     2.  **Generate the slice with `general-purpose`** — proceed and produce the Feature Testing & Regression slice, marking its tasks `**[Agent: general-purpose]**`. Flag this in the Recommendations table.
     3.  **Skip the Feature Testing & Regression slice** — set `SKIP_TESTS = true` for this run only; the user can re-run `/awos:tasks` later once a tester is hired.
+
+    If the question goes unanswered (an unattended run), default to option 2 — generate the slice with `general-purpose` and flag it in the Recommendations table — and continue. Do not halt without writing `tasks.md`.
```
dustyo-O: I'm not sure this is reliable detection. Maybe if question went unanswered ask the same question twice and if it's still skipped go to defaults?

dustyo-O (approved):

commands/tasks.md
AlexanderMakarov: Agreed — removed. It was redundant with the existing Step 1.3 / Step 3a guidance that already directs asking the user. Dropped the matching bullet in `tech.md` too.

commands/tasks.md
AlexanderMakarov: Reverted to "stop and ask." You're right it wasn't required — the unattended path is carried by the INTERACTION rule ("a skipped question is never a stop signal"), so dropping the word changed nothing.

commands/tasks.md
AlexanderMakarov: I kept the default-on-skip rather than re-asking. In `-p` a re-asked question is auto-dismissed again, so the retry only burns turns without changing the outcome, and interactively it double-prompts the user. The skip is detectable — `AskUserQuestion` returns an explicit skip result in unattended runs — which I confirmed across the awos-qa e2e runs, where `tasks.md` was written on every run including the ones that hit this QA gate. I removed the duplicate fallback sentence and folded it into the single global INTERACTION rule, marking option 2 as the default-on-skip so there's one source of truth.

dustyo-O:
lgtm

PR #134, opened 2026-06-10
feat: /awos:flow — generates the project's /implement-feature delivery flow

AlexanderMakarov:
## What

New run-once setup command `/awos:flow` that generates a project's end-to-end delivery automation. Every team's SDLC differs (ticket source, git flow, repo topology, review gates, delivery, trigger), so AWOS ships a **generator**, not a static flow: `/awos:flow` investigates the repo, reads team docs, interviews the user across six dimensions, and emits:

- **`context/product/delivery-flow.md`** — durable decision record, one section per dimension plus a per-service Tooling Inventory and a Local Customizations section. Single source of truth; flow-agnostic so **both** generated commands reuse it as consumers rather than re-interviewing the user.
- **`.claude/commands/implement-feature.md`** and/or **`.claude/commands/fix-bug.md`** — project-specific commands (names configurable) that drive one ticket through the team's real delivery flow. The **feature** command runs the full `/awos:spec → tech → tasks → implement → verify` chain plus the team's own delivery steps (branch/worktree prep, spec commits, review gates, delivery hand-off, ticket transition); the **bug-fix** command runs the lighter `diagnose → fix → scoped re-verify → targeted spec amendment` path (with a conformance-vs-divergence classification gate). The team chooses which of the two to generate. Both live in the project's own command namespace, deliberately outside `.claude/commands/awos/`, so neither the installer nor framework updates touch them.

## Design points

- **Transport preference:** investigation inventories CLI tools on PATH, MCP servers, and plugins per external service; CLI wins when both cover the same service (faster, cheaper in tokens). Chosen transport + fallback are recorded per service.
- **Worktree sub-interview:** worktrees are only offered after a shared-resource questionnaire (ports, DB, docker names/volumes, tunnels, devices, non-clonable services); otherwise main-repo-only with the blocking reason recorded.
- **Regeneration contract:** every generated stage is fenced with `<!-- awos:flow:stage=... -->` markers. Re-runs reconcile per stage — semantically unchanged stages regenerate silently; manually edited stages ask keep/drop/merge, and **keep** promotes the edit into Local Customizations so future regenerations preserve it automatically.
- **Unattended-safe (per #132):** a skipped/unanswered question is never a stop signal — documented defaults apply (keep wins reconciliation conflicts) and both deliverable writes are unconditional.

## Chain integration

- `commands/architecture.md` / `commands/hire.md` next-command pointers: hire → flow → spec.
- New **E2E-06** check in the ai-readiness-audit `end-to-end-delivery` dimension recommends `/awos:flow` when the artifacts are missing or stale; plugin bumped 2.1.0 → 2.2.0 in both manifests.

## Tests

- Five new Layer-1 lint tests pin the contracts: flow.md path wiring (templates + both artifacts + CLI preference + Explore delegation), stage markers + AWOS chain + orchestrator guard in the skeleton, Local Customizations + Tooling Inventory sections, chain pointers, E2E-06 presence. `npm test`: 83/83 pass; prettier clean.
- Behavioral coverage: two new **auto** e2e scenarios in awos-qa (branch `feat/flow-e2e-scenarios`, companion PR) — fresh generation passed 13/13 checks, re-run manual-edit preservation passed 9/9 checks against this branch. Merge this PR first; the awos-qa scenarios install AWOS from the checkout under test.

## Core command changes (outside the plugin)

This PR also changes three core framework files. The generated commands don't reimplement the AWOS chain — they invoke it (`/awos:spec`, `/awos:verify`), so the invoked side had to hold up its end of the new contract. Flagging them here explicitly since they affect every AWOS user, not just flow adopters:

- **`commands/spec.md` — new Update Mode** (plus Mode Detection). The `/fix-bug` flow classifies each bug as *conformance* (code violated a correct spec) or *divergence* (behavior intentionally changed); a divergence must amend the owning `functional-spec.md` or the spec silently drifts. Its `amend-spec` stage invokes `/awos:spec` naming an existing spec, which routes to Update Mode: edit in place, dated `## Change Log` entry, no new index, `Completed` status preserved. **Creation Mode is untouched** — an empty prompt still takes the first not-done roadmap item; only an explicit reference to an existing spec routes to Update Mode. `templates/functional-spec-template.md` gains the canonical `## Change Log` section, and `claude/commands/spec.md`'s argument-hint signals the amend path.
- **`commands/verify.md` — two strengthened bullets.** Road-tests of the generated flow stalled at the verify stage because `/awos:verify` handed the user a `run` command and waited — fatal for an unattended flow run, and the template can't fix it from outside since it delegates to this command. (1) The look-and-feel constraint now states that running the app is the agent's job, done safely: never disturb a service the user has running, shared persistence, or a contended resource — prefer a running instance or start an isolated one and clean it up; the agent self-certifies only objective UI states, and styling/layout judgment goes to the user via screenshot + `AskUserQuestion`. (2) The Step 3.4 manual-confirmation fallback is narrowed to a true last resort — no agent-driven render path exists at all — with alternate port / resource reclaim / project deploy step tried first.

These are same-release dependencies: the plugin generates commands whose stages call the core commands with these behaviors, and both ship together from this repo.

## Release Notes v2.2.0

* **New Features**
  * Introduced the `/awos:flow` delivery-flow generator to guide repository discovery and multi-step interviews, producing durable delivery decisions and an executable implementation workflow.

* **Documentation**
  * Updated plugin metadata and description to better reflect delivery-flow setup assistance.
  * Added/expanded delivery-flow and feature-implementation templates to standardize prompts, gating, and regeneration behavior.

* **Tests**
  * Strengthened prompt/template contract validation, including new `/awos:flow` coverage.

* **Chores**
  * Updated ignore rules for additional local workspace artifacts.


AlexanderMakarov:
Example of installation in https://github.com/provectus/sde-automation (use https://jsonl.qent.io/ or alike for CC sessions):
1. First try:
  - CC session [78486b23-3833-42ea-8a6f-28f38b1ae1be.jsonl.txt](https://github.com/user-attachments/files/28801339/78486b23-3833-42ea-8a6f-28f38b1ae1be.jsonl.txt), results:
  - context/product/[delivery-flow.md](https://github.com/user-attachments/files/28801441/delivery-flow.md)
  - .claude/commands/[implement-ticket.md](https://github.com/user-attachments/files/28801445/implement-ticket.md
  - https://github.com/provectus/sde-automation/pull/8
2. At [af27d93](https://github.com/provectus/awos/pull/134/commits/af27d93413b559d2161bd2052abff281f408f883) - in spite of the bigger diff with more features it consumed 48% less tokens (~359k), was 41% faster (2h11m), context size 21% less (258k), more sophisticated flow (fallback; fix of settings UI is included):
  - CC session [2e9e0b1a-e183-4a2e-8df4-b9d0b7a527e7.jsonl.txt](https://github.com/user-attachments/files/28848458/2e9e0b1a-e183-4a2e-8df4-b9d0b7a527e7.jsonl.txt) (actually a few more changes here and a different approach), results:
  - context/product/[delivery-flow.md](https://github.com/user-attachments/files/28848486/delivery-flow.md)
  - .claude/commands/[implement-ticket.md](https://github.com/user-attachments/files/28848494/implement-ticket.md)
  - https://github.com/provectus/sde-automation/pull/9

plugins/awos/skills/ai-readiness-audit/dimensions/end-to-end-delivery.md
```diff
@@ -75,3 +75,13 @@ For single-layer projects, most checks will auto-SKIP since cross-layer delivery
...
+
+### E2E-06: Generated delivery flow automation
+
+- **What:** The project has a generated end-to-end delivery flow — a decision record plus a command that drives a ticket through spec, implementation, verification, review, and delivery
+- **How:** Check for `context/product/delivery-flow.md` (the decision record produced by `/awos:flow`) and a flow command at `.claude/commands/implement-ticket.md`. If both exist, spot-check that the decision record's Generation Log is not older than significant changes to the repo's CI/release configuration.
+- **Pass:** Both artifacts exist and the decision record reflects the current toolchain
+- **Warn:** One artifact exists without the other, or the decision record is visibly stale — recommend re-running `/awos:flow`
+- **Fail:** Neither exists — recommend running `/awos:flow` to generate the delivery flow
+- **Skip-When:** Spec-driven-development artifact shows the project does not use AWOS (no `context/` directory)
+- **Severity:** medium
```
workshur: I'm not convinced that enforcing such strict end-to-end flow checks is necessary here. While automation can boost the rating, it shouldn't be calculated by awos artifacts. Instead, we should gather more evidence from indirect sources like hooks, review commands, CI, etc.

workshur (commented):
The idea is good, but I’m not sure we should pull this into the main AWOS flow right away.

Project flows can differ a lot, so I’d first test this properly. That said, I think we could add the command to the plugin after a small round of refinement.

1. I don’t really like the command name `/implement-ticket`. I don’t like the mention of a ticket in general. Maybe this is more about implementing a feature, and it shouldn’t really matter where the requirements come from.

2. I’d like to see this more as an assistant that helps set up the flow and points out things that could still be automated.

3. Some steps, like merge, require special attention.

plugins/awos/templates/delivery-flow-template.md
```diff
@@ -0,0 +1,96 @@
...
+- **Approvals:** [who signs off, where recorded]
+- **Versioning:** [what gets bumped, where, by whom]
+- **Deployment:** [manual CD job | scheduled | fully automatic | local deploy step — the command, and when the flow runs it: after the merge | after post-merge CI is green | never]
+- **Definition of Done:** [the evidence that closes the loop — may include post-merge pipelines green; plus the ticket transition, only when the source has tickets to transition]
+
+## 6. Trigger
+
+- **Supported:** manual — `/implement-feature <ticket-id-or-link>`
+- **Wanted (setup notes only):** [polling via `/loop`/cron, webhook — what to set up, or "none". For unattended runs, the driver chains headless stage invocations (`claude -p "/implement-feature <ticket>"`; resume-detection makes each idempotent) — record the prerequisites the operator owns: the machine, `claude` on PATH, a permission mode that allows edits, and headless-capable transports]
+
```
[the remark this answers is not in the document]
AlexanderMakarov: Both addressed.

1. Trigger example in `delivery-flow-template.md` now reads `/implement-feature <feature — ticket ID, link, or file path>` (and the headless example likewise) — matching the command's source-agnostic contract.
2. Resume precedence: made it explicit, but **scoped**. The artifact scan only covers the spec chain (`functional-spec`/`technical-considerations`/`tasks`), so for those stages the on-disk artifacts win over a stale `flow-log.md` and the log is repaired to match. Past spec generation there is no `context/spec/` artifact, so the log stays the only resume signal — a blanket "first missing artifact is authoritative" would break resume for implement/verify/merge, which have no such artifact.

Fixed in 64a3902.

plugins/awos/skills/ai-readiness-audit/dimensions/end-to-end-delivery.md
AlexanderMakarov: Agreed — removed E2E-06. Scoring delivery automation off the presence of AWOS's own generated files is marking our own homework, and it unfairly fails projects that automate via CI / hooks / review commands. That scoring belongs in the `sdlc-automation` dimension (on `feat/sdlc-automation-audit`), whose SDLC-01/02 already measure it from indirect evidence; `end-to-end-delivery` stays focused on vertical-slice delivery.

AlexanderMakarov:
@workshur — thanks, agreed on the direction; reworked the PR to match.

- **Main flow / plugin (A0):** un-wired `/awos:flow` from the `/awos:hire` → `/awos:architecture` setup chain and moved the command into the plugin (`plugins/awos/commands/flow.md`), opt-in via the marketplace rather than part of the run-once chain. Its templates ship bundled in the plugin too, so it's self-contained — no installer dependency, and no migration (it only landed on this branch).
- **Naming (A1):** generated command renamed `/implement-ticket` → `/implement-feature`; intake reframed as "a feature, wherever the requirements come from."
- **Posture (A2):** reframed from a rigid end-to-end driver to a setup assistant that flags which steps stay manual / could still be automated, instead of asserting full automation.
- **Merge (A3):** kept the per-run merge confirmation and the double target-branch conflict check; merge is called out as a special-attention step.
- **E2E-06:** removed (see the inline thread) — automation-coverage belongs in the `sdlc-automation` dimension.

Left the threads open for you to confirm.

AlexanderMakarov:
Rewrote to plugin while it is the same way to run it - `/awos:flow`. Now it generates `/implement-feature` command. Plus added few minor corrections.

Comparison with https://github.com/provectus/awos/pull/134#issuecomment-4671798357 last session and new (use https://jsonl.qent.io/ or alike for JSONL CC sessions investigation).

Metrics: **token usage** = input + output (non-cache, billable); **context max** = peak `input + cache_creation + cache_read` on a single request; **agent working** = wall-clock from the slash-command to last activity before the session idled.

| Command | CC sessions | Model | Token usage (in+out) | Agent working | Context max | Files | PR |
|---|---|---|---|---|---|---|---|
| `/implement-ticket` (last) | (no /awos:flow) [/implement-ticket](https://github.com/user-attachments/files/28801339/78486b23-3833-42ea-8a6f-28f38b1ae1be.jsonl.txt) | `claude-fable-5` | ~402k (in 199k + out 203k) | 2h11m | 258k | [delivery-flow.md](https://github.com/user-attachments/files/28848486/delivery-flow.md), [implement-ticket.md](https://github.com/user-attachments/files/28848494/implement-ticket.md) | [#9](https://github.com/provectus/sde-automation/pull/9) +4,496 -163 |
| `/implement-feature` #1 | [/awos:flow](https://github.com/user-attachments/files/29055313/d27d7c65-84cd-4f18-9cd2-7a45342436cd.jsonl.txt), [/implement-ticket](https://github.com/user-attachments/files/29055302/63e238fb-6f94-4e8c-920f-ca7e017a5cf9.jsonl.txt) | `claude-opus-4-8` | ~556k (in 140k + out 415k) | 1h25m | 319k | missed both files | [#11](https://github.com/provectus/sde-automation/pull/11) +3,632 -87 |
| `/implement-feature` #2 | [/awos:flow](https://github.com/user-attachments/files/29055296/b35a5d49-00b7-429a-b2a4-e0c54a60dbb2.jsonl.txt), [/implement-ticket](https://github.com/user-attachments/files/29055277/1e3fb36c-246b-47b2-be4e-0fe01398bc18.jsonl.txt) | `claude-opus-4-8` | ~672k (in 125k + out 548k) | 2h35m | 415k | [implement-feature.md](https://github.com/provectus/sde-automation/blob/2482fb710479d52a49949aeac8d2417243cc55b8/.claude/commands/implement-feature.md), [delivery-flow.md](https://github.com/provectus/sde-automation/blob/2482fb710479d52a49949aeac8d2417243cc55b8/context/product/delivery-flow.md) | [#12](https://github.com/provectus/sde-automation/pull/12) +4,076 -112, includes few extra UI features and tried worktrees |

Notes:
- `/implement-feature` runs on **Opus 4.8 [1m]**; the `/implement-ticket` baseline ran on **Fable 5** — token/speed deltas are partly model-driven, not just flow changes.

commands/verify.md
```diff
@@ -37,7 +37,7 @@ Verify a specification's implementation against its acceptance criteria. For eac
...
...'t running, drive the actual UI through whatever browser-automation tool the project ships (Playwright MCP/CLI, Cypress, the chrome MCP, etc.), and capture screenshots of the states the criteria describe. **Running the app is your job, not the user's.** A shared resource the app normally binds — a port already held by a running service, a single database, a device — is not a reason to hand a `run` command to the user: reclaim it (stop the service, use an alternate port, spin a throwaway instance) or drive the project's own deploy/run step yourself, then verify against it. Only after every agent-driven way to render the behavior is genuinely unavailable does manual confirmation apply. Save them to `docs/screenshots/` — the same evidence folder the `testing-expert` agent writes E2E captures to — naming each file so it sorts by spec: `docs/screenshots/<spec-directory>-<short-state>.png` (e.g. `docs/screenshots/011-scheduled-tasks-amber-pill.png`). The browser tool creates the folder on first write; do not edit `.gitignore` — git-ignoring `docs/screenshots/` is a one-time project setup, not part of verification. Reference the saved paths in the report — they are what the human reviews.
```
dustyo-O: > **Running the app is your job, not the user's.**

no matter to drive this in current skill. If user requires assistance for running app, he can make through his own settings/skills

dustyo-O: >look-and-feel

I do not believe current model are capable of verifiing look and feel. better approach -> open particular feature and ask user IF it looks like described in specs

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- **An unanswered question is never a stop signal — but interactive and unattended runs handle it differently.** `AskUserQuestion` returns a `No response after 60s` result when no one answers in time; that is the harness's guard so headless runs never hang, not a signal the skill sets or can extend. Read whether this run is unattended from the `AWOS_UNATTENDED` environment variable — the trigger setup exports it for cron/`/loop`/`claude -p` drivers (§6), and interactive runs leave it unset. **Unattended (`AWOS_UNATTENDED` set):** a no-answer is expected — fall back to the documented default for that question and continue through the remaining steps, including writing both artifacts. **Interactive (unset):** a 60s timeout usually means the user is thinking or briefly away, not that they have no preference — re-ask the question once; if it times out again, proceed with the default but name the default you took and invite correction. The defaults: for interview dimensions, the answer inferred from the investigation and team docs (or the most conservative option when nothing was inferred); for re-run reconciliation conflicts, keep the manual edit.
```
dustyo-O, line 38: 60s? where does that come from?

dustyo-O: > re-ask the question once

that doesn't seem reliable. When I do two projects in parallel, timeouts can be huge

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- **User Prompt (Optional):** <user_prompt>$ARGUMENTS</user_prompt>
+- **Prerequisite Input:** `context/product/architecture.md` (the technology stack decisions).
+- **Recommended Input:** `context/product/hired-agents.md` (the specialist roster — the generated flow delegates to these agents).
+- **Re-run Inputs (if they exist):** `context/product/delivery-flow.md` and the generated command files named in its **Generated Commands** field (default `.claude/commands/implement-feature.md` and `.claude/commands/fix-bug.md`).
+- **Template Files (bundled with this command in the plugin):** `${CLAUDE_PLUGIN_ROOT}/templates/delivery-flow-template.md`, `${CLAUDE_PLUGIN_ROOT}/templates/implement-feature-template.md`, `${CLAUDE_PLUGIN_ROOT}/templates/fix-bug-template.md`.
+- **Outputs:** `context/product/delivery-flow.md` and the one or two generated command files under `.claude/commands/`, per the Command-set decision (default paths `.claude/commands/implement-feature.md` and `.claude/commands/fix-bug.md`).
+
+---
+
+# INTERACTION
```
dustyo-O, line 34: I do not get what this section tries to achieve. some advices seems arguable, like "Each option must be answerable without follow-up typing" --> why, some seems to confuse and even possibly drop the quality of the interaction, like "collapse them into one option and let the user supply the specifics via the built-in free-text input" --> which possible contradicts with previous one.

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+---
+
+# PROCESS
+
+Follow this process precisely.
+
+## Step 1: Prerequisite Checks & Mode Detection
+
+1.  **Stale-session check.** This command's generator version is `2.3.0` (a literal constant, kept in sync with the plugin manifest). Read `${CLAUDE_PLUGIN_ROOT}/.claude-plugin/plugin.json`: if its `version` differs, the session is running a cached older copy of this command — the plugin was updated (e.g. `git pull` in the marketplace clone) after the session loaded it. Tell the user to restart the session so the updated command loads, and stop; generating from a stale prompt silently drops the update's fixes.
+2.  If `context/product/architecture.md` does not exist, stop and tell the user to run `/awos:architecture` first.
```
dustyo-O, line 55: product.md?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+
+---
+
+# PROCESS
+
+Follow this process precisely.
+
+## Step 1: Prerequisite Checks & Mode Detection
+
+1.  **Stale-session check.** This command's generator version is `2.3.0` (a literal constant, kept in sync with the plugin manifest). Read `${CLAUDE_PLUGIN_ROOT}/.claude-plugin/plugin.json`: if its `version` differs, the session is running a cached older copy of this command — the plugin was updated (e.g. `git pull` in the marketplace clone) after the session loaded it. Tell the user to restart the session so the updated command loads, and stop; generating from a stale prompt silently drops the update's fixes.
```
dustyo-O: why should we? why don't we let the user to control it themselves?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- **No absence claims without a probe.** The decision record and the generated commands may assert that something does not exist — "no pre-commit hooks", "nothing runs on merge", "no automatic reviewer" — only when the investigation explicitly probed for it and came up empty; record what was probed alongside the claim. A signal nobody looked for is unknown, not absent: omit the claim rather than writing a confident "no X". A road-test's generated command asserted "there are no pre-commit hooks" while a versioned hook was installed via `core.hooksPath` — the first real commit hit it unprepared.
+- **Format/lint gate scope:** note whether the project's format or lint gate runs over the **whole repo** (e.g. `make check` shelling out to `prettier --check .`, `ruff check .`, a repo-wide linter). When it does, AWOS working files the flow writes — `context/spec/**/flow-log.md`, the generated command files — fall under that gate and can fail it. Record the scope; Step 8 advises the project-side ignore fix. Do not add a format pass inside the generated flow — the fix is the project's ignore config, not a flow stage.
+
+## Step 3: Collect Team Documentation
```
dustyo-O, line 78: if my PR would be merged, it's better to align your contribution with mine

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+    - **Ticket state transitions.** When the source has tickets, map which flow event moves the ticket to which tracker state across the whole cycle, not just at the end: start (e.g. → In Progress), change request opened (→ In Review), **a gate or review failing** (→ back to the team's needs-work state, e.g. To Do — so the developer sees it), merged, and done. Record the state names in the project's own vocabulary; the generated command transitions the ticket on each mapped event using the §7 transport. Do not record the map on state names alone — **validate it at generation time**: fetch the available transitions for a sample issue via the §7 transport and record the full transition **chain** for each mapped event, plus transition IDs where the tracker exposes them. Tracker workflows often have no direct hop to the target state (no `Open → In Review`; the real chain is `Open → In Progress → In Review`) — a target state recorded without its chain is exactly the mapping that fails on the first unattended run. Omit entirely for ticketless sources.
```
dustyo-O, line 102: with the ticketing, there are more questions to resolve:
- how do new findings/deffered features are tracked through the ticket system -> sub-tickets, new tickets under the same umbrella
- what is the tagging system
- should viewers be copied from source ticket to connected
- ...

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+
+**Command set & names.** Decide which commands `/awos:flow` generates and what each is called. Ask as one `AskUserQuestion`: generate the **feature** flow, the **bug-fix** flow, or **both** (the common answer). Then take the slash-command name for each selected command via the built-in free-text input — defaults are `/implement-feature` and `/fix-bug`, but a team may prefer shorter names (`/feature`, `/fix`); the file is the name without the leading slash (`fix.md` for `/fix`). Record the chosen names and their `.claude/commands/*.md` filenames in the decision record's **Generated Commands** field — Step 6 writes to those filenames and re-runs reconcile them.
```
dustyo-O, line 110: inconsistency is not a good idea -> how I could guess the right command name in the particular project?

I suggest /awos:feature and /awos:fix

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
...for an existing init hook (`make worktree-init` or similar). A fresh worktree also lacks everything git-ignored: probe `.gitignore` for build-required artifacts (`node_modules/`, generated sources, local env files) and derive the bring-up steps from the lockfile and package scripts — the install command, codegen steps, env-file copies. Those bring-up steps are part of the isolation recipe; without them the first agent in a new worktree loses its opening minutes to an undocumented install. The generated workspace stage never improvises worktree preparation: when Step 2 found an existing worktree command, skill, or init script (`/worktree`, `make worktree-init`, a project script), the stage invokes it (a Step 4.5 reuse decision); otherwise record the full recipe in §2 as an exact command sequence — creation, bring-up, cleanup — and the stage executes it verbatim. Real preparation is bigger than `git worktree add`: installs, codegen, env files, and services/networks that must not conflict with the primary checkout. Conclude one of: worktrees viable (record the reused command or the exact recipe, bring-up steps included), or main-repo-only (record the blocking resources as the reason).
```
dustyo-O, line 116: also symlinks which point outside the project's folder could be an issue

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
+---
+description: Sets up the project's /implement-feature delivery flow — investigates, interviews, writes the command, and flags what stays manual.
+argument-hint: '[focus, optional — e.g. a dimension to revisit on re-run]'
+---
+
+# ROLE
+
+You are an expert Delivery Flow Engineer. You act as a setup assistant: you help the team stand up an end-to-end delivery flow and you point out which steps it can automate now and which still need a human. You design and generate the automation — how a functional task travels from its source (a ticket, a document, a prompt) through the AWOS chain (`/awos:spec` → `/awos:tech` → `/awos:tasks` → `/awos:implement` → `/awos:verify`) and through the project's own delivery steps (branching, commits, review, deployment, ticket transition) until it is Done. Every team's flow is different, so you never ship a generic recipe — you investigate the project, interview the user, generate a flow tailored to this team, and report the gaps it does not yet cover rather than claiming full automation.
```
dustyo-O: actually flow tailored to this project. skill is unable to produce different flows for different teams if all of them work on one project but have different flows (happens with big enterprises)

plugins/awos/commands/flow.md
AlexanderMakarov: This is cc timeout after which it decides that "user haven't answered on question" - not official behavior https://www.reddit.com/r/ClaudeAI/comments/1ulh0ic/claude_code_started_to_use_timeout_on/, the same as https://provectus.slack.com/archives/C01BMPYRV8V/p1783079818561049

plugins/awos/commands/flow.md
AlexanderMakarov: No, exactly `/awos:architecture`.

`/awos:architecture` generates `context/product/architecture.md` so if we haven't found this file then need to stop and ask user to generate it.

plugins/awos/commands/flow.md
AlexanderMakarov: >flow tailored to this project

Right, where `this project` = project where AWOS is installed.

>skill is unable to produce different flows for different teams

Correct. Multiple teams working on the same project is agains best practices so I don't see the point of supporting it.

plugins/awos/commands/flow.md
AlexanderMakarov: Fair point — reworked this rather than defending it.

**Dropped the stale-session check** (the forced restart) in 4d02877. You're right that halting the user was heavier than warranted.

That check was never what actually delivered the updates. What does is already in the command:

- Each generated SDLC command stamps its generator `version=` into a footer marker (Step 6).
- On a re-run, a footer older than the current generator triggers full regeneration from the current templates (Step 1.4) — even when `context/product/delivery-flow.md` is unchanged.

So a user re-running `/awos:flow` from an updated AWOS version gets the regenerated commands by just re-running. No restart prompt, and the update still lands — the "let the user control it themselves" you asked for.

plugins/awos/commands/flow.md
AlexanderMakarov: 1. This is too much to support on start.
2. Users always may customize generated commands with theirs extra requirements. Or rely on self-improvement feature of commands.

plugins/awos/commands/flow.md
AlexanderMakarov: >/awos:feature and /awos:fix

Definitely not, because we are generating project-specific commands, not AWOS-specific.

>how I could guess the right command name in the particular project?

This is per project conventions. Actually the reasons of supporting custom command names:
- Don't conflict with existing ones,
- Satisfy team or company standards/conventions.

commands/verify.md
AlexanderMakarov: The "running the app is your job" line is the anti-punt guardrail — verify's value is that it renders and checks the behavior itself rather than writing "the user can verify this manually" and marking the criterion `[x]`. Running the app is a precondition of rendering, so it can't be fully out of scope here, and the bullet already falls back to manual confirmation once every agent-driven path is genuinely unavailable. A dedicated `run` skill or project settings is what this should *use* when present — it doesn't remove verify's responsibility to get the app up. Kept the principle but agree the prescriptive "stop the service" was too aggressive: reworded to "reclaim it without disrupting a service the user is relying on" so an over-eager agent doesn't kill a running service. (6783ad5)

AlexanderMakarov: Agreed — split the two. Objective UI states (a control is present, a redirect fires, a state toggles) the agent still confirms itself; whether the rendering *looks* like the spec describes (styling, layout, polish) is captured to `docs/screenshots/`, then confirmed with the user via `AskUserQuestion` that points them to the saved screenshot path to open — `AskUserQuestion` can't render the image inline, so it references the file — instead of the model self-certifying an aesthetic match. (6783ad5)

plugins/awos/commands/flow.md
AlexanderMakarov: Fair — the wording invited the contradiction reading. The rule is: the *choice between options* must be meaningful on its own (scannable, no typing to tell options apart); the free-text field supplies *parameters* to a chosen option (a link, a path), not the decision. So "collapse into one option + free-text" isn't the opposite of "answerable without follow-up typing" — it's the same rule applied when two options differ only by a value you'd type anyway. Rewrote the bullet to say that explicitly. (ecdef9c)

plugins/awos/commands/flow.md
AlexanderMakarov: This tracks with #138 (external documentation gathering): flow.md's Step 2 tooling/transport inventory and Step 3 "Collect Team Documentation" overlap your `configure-external-sources` skill + `project-documentation.md` manifest, and the two should share one mechanism rather than each probe connectors independently. Merge expectation: whichever of the two merges last owns the integration. If #138 merges first, flow.md's Step 2/3 will consume your `sources.md`/skill and manifest instead of re-detecting transports. If #138 merges last, it updates `/awos:flow` to reuse the shared mechanism where available.

plugins/awos/commands/flow.md
AlexanderMakarov: Good catch — added to the Repository topology dimension: symlinks whose target resolves outside the project root are now flagged as a hazard, since a worktree or headless/CI checkout may not resolve them (breaking `context/` reachability when `context/` is reached via such a link) and repo-wide lint/format may traverse them. (ecdef9c)

plugins/awos/commands/flow.md
AlexanderMakarov: Dropped the stale-session check rather than hardening it. Because the version is pinned between releases, a reloaded older `flow.md` and the on-disk `plugin.json` read the same version, so the equality check never fired for intra-version command changes — the exact case it was meant to catch. The generator-update re-run detector already covers the staleness that matters: the next re-run on the updated command sees an older footer `version=` and regenerates every stage. Kept the constant (it still stamps artifact footers and drives that re-run comparison); only its user-facing "restart the session" role is gone, so nothing has to reason about a hardcoded value at invocation time anymore. Test `:920` keeps all its assertions. (4d02877)

plugins/awos/commands/flow.md
dustyo-O: > Don't conflict with existing ones 
awos prefix deals with that

without strict rules, implementations are going to be loose, they are always tend to be without linters/prettier/other rules

plugins/awos/commands/flow.md
dustyo-O: hmm, let me rephrase -> that is quite common for big projects with 30+ developers working on the same codebase to have multiple teams contributing to the same repository.

also, if there are some common repos for the company, like common ux lib (which may be even open-source), which has it's own flow, tailored to the repository, not the team.

my complaint is only about the word "team". I think that entity "project" is more correct speaking about the flows, flows belong to projects/repositories, not the team. same team contributing to the different repos will follow different flows. and vice versa -> different teams contributing to the same repo will likely follow the same flow

commands/verify.md
```diff
@@ -37,7 +37,7 @@ Verify a specification's implementation against its acceptance criteria. For eac
...
... or drive the project's own deploy/run step against a scratch target — then verify against it. The agent self-certifies only objective, checkable UI states — a control is present, a state toggles, a redirect fires. Whether the rendered result _looks_ the way the spec describes it (styling, layout, visual polish) is a judgment models make unreliably: capture the state to `docs/screenshots/`, then ask the user via `AskUserQuestion` to open that screenshot and confirm it matches the spec's described look-and-feel, rather than self-certifying the match. Only after every agent-driven way to render the behavior is genuinely unavailable does manual confirmation of the behavior itself apply. Save them to `docs/screenshots/` — the same evidence folder the `testing-expert` agent writes E2E captures to — naming each file so it sorts by spec: `docs/screenshots/<spec-directory>-<short-state>.png` (e.g. `docs/screenshots/011-scheduled-tasks-amber-pill.png`). The browser tool creates the folder on first write; do not edit `.gitignore` — git-ignoring `docs/screenshots/` is a one-time project setup, not part of verification. Reference the saved paths in the report — they are what the human reviews.
```
dustyo-O: >Running the app is your job, not the user's.

I personally hate when claude starts the app when I didn't ask him. it frequently forget of instances he launched and it becomes my job to clean the mess after him

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+
+---
+
+# TASK
+
+Generate (or regenerate):
+
+1. `context/product/delivery-flow.md` — the durable record of the team's delivery decisions, one section per dimension. This file is the single source of truth for _decisions_; everything else is derived from it.
+2. One or two project-specific commands under `.claude/commands/`, per the Step 4 **Command set & names** decision: a **feature** command (default name `/implement-feature`, file `implement-feature.md`) that drives the full flow end to end for one feature, and/or a **bug-fix** command (default name `/fix-bug`, file `fix-bug.md`) for the lighter bug-fix flow (diagnose → fix → scoped re-verify → targeted spec amendment). The team chooses which of the two to generate and may rename either (e.g. `/feature`, `/fix`); the chosen names and filenames are recorded in the decision record so re-runs reconcile the right files.
```
dustyo-O, line 17: why default name for bug fix is fix bug?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+
+This command generates automation; it never executes the flow itself. The decision record is flow-agnostic by design — both generated commands reuse it as consumers rather than re-interviewing the user.
+
+---
+
+# INPUTS & OUTPUTS
+
+- **User Prompt (Optional):** <user_prompt>$ARGUMENTS</user_prompt>
+- **Prerequisite Input:** `context/product/architecture.md` (the technology stack decisions).
+- **Recommended Input:** `context/product/hired-agents.md` (the specialist roster — the generated flow delegates to these agents).
```
dustyo-O: Recommended input is new, why not Optional?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- **Re-run Inputs (if they exist):** `context/product/delivery-flow.md` and the generated command files named in its **Generated Commands** field (default `.claude/commands/implement-feature.md` and `.claude/commands/fix-bug.md`).
```
dustyo-O, line 28: same here

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- Mark an option "(Recommended)" only when the investigation gives evidence to prefer it. Factual questions about the team's world (does documentation exist? which tracker do you use?) have nothing to recommend — present those options neutrally; a default is just a default.
+- Each option must stand on its own as a decision — the _choice between options_ has to be meaningful without the user typing to disambiguate it, which keeps the question scannable and fast to answer. This does not forbid free text: the built-in free-text input supplies _parameters_ to a chosen option (a link, a path, a name), never the decision itself. So don't split options that differ only by a follow-up you'll ask for anyway (e.g. "Yes — Confluence" vs. "Yes — local files" when both just mean "now provide the link/path") — collapse them into one option and let the free-text input carry the specifics. When the answer is inherently free-form (a link, a path, a name), don't wrap it in a multiple-choice at all — and never present an option whose description tells the user to pick "Other" instead.
```
dustyo-O, line 41: looks complicated. Im not against it, but have you any proofs this really works, im just curious?

commands/verify.md
AlexanderMakarov: The intent here is that verify runs and checks the app itself instead of punting with "the user can verify this manually" — but it has to do that safely, which is exactly the mess you're describing. Reworking the guidance to require: prefer an already-running instance; otherwise start an isolated one (alternate port, throwaway data, a scratch target) and track it so it's stopped afterward — no orphaned processes; never stop/restart a service you have running, mutate shared persistence, or seize a contended port/device/third-party resource; and if a criterion can't be rendered without disturbing something shared, ask you instead of forcing it. So the run-it-yourself mandate stays, and the leak-and-cleanup problem is closed by construction.

plugins/awos/commands/flow.md
AlexanderMakarov: Agreed on the distinction — a flow belongs to the project/repository (its CI, branch policy, review gates), not the team. Same team across repos runs different flows; different teams on one repo run the same flow. Reframing the ROLE and prose from "the team's flow" to "the project's/repository's flow": the interview still talks to a person, but what it captures belongs to the repo. Keeping "team" only where a decision genuinely is about people (who reviews, the notifications channel).

plugins/awos/commands/flow.md
AlexanderMakarov: Two things to separate.
1. These generated commands aren't AWOS-prefixed plugin commands — they're project-local (`.claude/commands/implement-feature.md`, invoked `/implement-feature`), so there's no `awos:` prefix resolving conflicts; the namespace is the project's own.
2. There is a stable default — `/implement-feature` and `/fix-bug` — recorded in the decision record and surfaced by slash-autocomplete, so "how do I guess the name" is answered by the default unless a team deliberately renames. Hard-coding `/awos:feature`/`/awos:fix` would misrepresent project-specific automation as AWOS-owned commands. And naming consistency and implementation quality (linters/prettier) are separate axes — fixing the names wouldn't enforce the latter.
 
So I'd keep configurable names with strong defaults rather than two hard-coded ones.

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- Write options in the project's own vocabulary, as settled by the investigation and earlier answers. Never mention a concept the decisions already rule out (no "transition the ticket" when tasks come from local files; no "PR" on a repo with no code host) and no unexplained shorthand — name the concrete thing ("asks you to confirm before merging, each time the command runs", not "per-run confirm").
```
dustyo-O, line 42: never seen an issue like that, when claude named something in terms different from project's vocabulary, maybe skip this?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- When a question decides how far automation goes along a pipeline, the options are cut-points: each reads "After <concrete event>: <what still happens>, then stop", ordered from the earliest stop to full automation, covering every meaningful stop point. Anchor every option to the same event names — don't call the same thing "CI passes" in one option and "gates are green" in another.
```
dustyo-O, line 43: Maybe name all the possible pipeline spots with their possible titles? There shouldn't be many...

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+7.  **Notifications.** The more autonomous the flow, the more the team needs to stay aware of it — "human on the loop", not "in the loop". Ask where the flow announces itself and on which transitions: a team channel (Slack, Teams, a PR comment — via a Step 2 transport), and which events post — spec ready for review, change request opened, gates passed, merged, deployed, or blocked-and-waiting. Lean toward announcing the events that either need a human (an approval or merge the flow paused for) or change shared state (merged, deployed); a fully interactive run where the user is already watching may want none. The point is that removing a human gate must not remove the team's visibility into what happened.
```
dustyo-O: there can be some instructions added to notifications about how to reproduce new feature on a fresh dev stage after deployment

plugins/awos/README.md
```diff
@@ -83,12 +85,27 @@ context/audits/YYYY-MM-DD_HH-MM-SS/
 
 `report.md` and `report.html` are always rendered together from `audit.json` — never hand-written. In org mode each `per-repo/<repo>/` subdir holds that repo's full audit, and an `org-portfolio.json` drives the org-level report. The directory name is the run's start timestamp, so every audit — including same-day re-runs — stands alone; earlier directories are history, never input.
 
+## Delivery Flow
+
+`/awos:flow` interviews the team and investigates the repo, then writes a decision record (`context/product/delivery-flow.md`) and generates one or two project-specific commands in `.claude/commands/`:
+
+- **`/implement-feature <feature>`** — drives one feature end to end through the AWOS chain (`spec → tech → tasks → implement → verify`) and the team's delivery steps (branch, review, change request, merge, deploy, close).
```
dustyo-O: name can vary, so this instruction confuses model. but if we agreed to /awos:feature, that would be much much better

plugins/awos/templates/fix-bug-template.md
```diff
@@ -0,0 +1,238 @@
...
+
+Wait with the `Monitor` tool, never foreground `sleep` loops: a poll loop that emits each gate's terminal result and exits when all are settled, its timeout sized to the typical pipeline duration recorded in §4, the poll interval 30s+ against remote APIs, and the filter covering every terminal state — failures and cancellations, not just success. Apply §4's max-wait & escalation policy when a poll window expires without the gates settling — auto-relaunch the monitor, or ask the human past the recorded threshold; never wait forever.
+
+<!-- /awos:flow:stage -->
+
+<!-- awos:flow:stage=merge -->
+
+### Step 13: Merge
+
+[Per §2: the target branch may have moved while the gates ran — re-check mergeability via the chosen transport or a fresh fetch + dry-run merge. If it no longer merges cleanly: sync per the recorded policy (resolution delegated per §8), push, and return to Step 11 — the remote gates run again on the new commit before any merge.]
```
dustyo-O: in implement-feature template we return to step 10, which is actually gates, I guess here we whether oversight the number or the title of step.

copy pasting problem will take us down, we have to consider some template engine and building stage for our skills

plugins/awos/templates/fix-bug-template.md
```diff
@@ -0,0 +1,238 @@
...
...et in the fetch stage, so the path is known from the first stage on and never moves; when classify resolves an owning spec the log records the spec dir, but stays keyed to the bug. Each entry: the stage name, what was produced and where (paths, branch, commit), the classification verdict once known, any decisions taken, and which stage comes next. The log is the flow's memory outside the context window — a fresh session resumes by reading this one small file. It is committed with the work (commit-push stages it alongside the code), so it must never become an uncommittable leftover: **once the change request is opened — or the change is merged — stop writing to the tracked log.** New commits are unwelcome on a change request under review or already merged, so a late append would strand a change that can never reach it. From that point, report late-stage progress (gate results, merge, close-out evidence) to the user and via §9 notifications, and resume the remote stages from remote state — the open/merged change request and the ticket status, which the resume-detection stage already inspects. The close stage leaves a clean working tree and never writes a final entry it cannot commit.
```
dustyo-O, line 51: shouldn't the path be under context/spec?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- **No absence claims without a probe.** The decision record and the generated commands may assert that something does not exist — "no pre-commit hooks", "nothing runs on merge", "no automatic reviewer" — only when the investigation explicitly probed for it and came up empty; record what was probed alongside the claim. A signal nobody looked for is unknown, not absent: omit the claim rather than writing a confident "no X". A road-test's generated command asserted "there are no pre-commit hooks" while a versioned hook was installed via `core.hooksPath` — the first real commit hit it unprepared.
+- **Format/lint gate scope:** note whether the project's format or lint gate runs over the **whole repo** (e.g. `make check` shelling out to `prettier --check .`, `ruff check .`, a repo-wide linter). When it does, AWOS working files the flow writes — `context/spec/**/flow-log.md`, the generated command files — fall under that gate and can fail it. Record the scope; Step 8 advises the project-side ignore fix. Do not add a format pass inside the generated flow — the fix is the project's ignore config, not a flow stage.
```
dustyo-O: bug-fix flow files should be taken into consideration?

plugins/awos/commands/flow.md
```diff
@@ -0,0 +1,180 @@
...
+- **Trigger setup notes:** if the team chose unattended runs, the concrete configuration — the exact invocation to schedule, the scheduler, and the prerequisites (permission mode, `claude` on PATH, `AWOS_UNATTENDED=1` in the driver's environment, headless-capable transports) — not just a pointer at the idea.
+- **Project-side setup fixes:** any project config the flow needs but cannot set itself. When Step 2 found a repo-wide format/lint gate, advise adding the AWOS working files (`context/spec/**/flow-log.md` and the generated command files) to the project's ignore (`.prettierignore`, lint excludes) so the flow's own output doesn't fail the gate. When Step 4.5 recorded a hardcoded-constant override instead of a fix, restate it here. When Step 2 found an always-loaded routing policy that prescribes the manual `/awos:*` chain without naming the generated commands, advise updating it and offer the concrete wording — a line per generated command stating when an agent should reach for it, plus a pointer to `context/product/delivery-flow.md` as the decision record behind them. These are project-side changes the user applies — the flow flags them, it does not make them.
```
dustyo-O: also bug fix flow files not mentioned

plugins/awos/templates/implement-feature-template.md
```diff
@@ -0,0 +1,198 @@
...
+
+<!-- /awos:flow:stage -->
+
+<!-- awos:flow:stage=verify -->
+
+### Step 7: Verify
+
+Run `/awos:verify` [per §8: in a subagent if it is non-interactive], returning the verdict and the list of gaps. Address gaps before proceeding.
+
+Running the app to verify is the flow's job, not the user's. [If §2/§3 recorded a shared resource the app binds — a port a running service holds, a single database, a device — the workspace guardrail reserves it for normal work, but verification still needs a real render: reclaim the resource (stop the service, alternate port, throwaway instance) or drive the project's §5 deploy/run step and verify against it, per the sanctioned verification path §2/§3 records. Don't hand the user a `run` command or defer a drivable criterion to a later manual deploy — manual confirmation is only for a criterion the agent genuinely cannot render here.]
```
dustyo-O: >stop the service

but why? if the user started it. 

also relevant for other flows

plugins/awos/templates/delivery-flow-template.md
```diff
@@ -0,0 +1,125 @@
+# Delivery Flow
+
+Generated by `/awos:flow` on [YYYY-MM-DD]. This file is the single source of truth for the team's delivery decisions — `.claude/commands/implement-feature.md` and any future flow commands are derived from it. Re-run `/awos:flow` to change a decision; edit the generated command directly only for details no decision covers (re-runs will offer to promote such edits into §10).
```
dustyo-O: >team's delivery decisions

if we are switching from team to project, there are still few places to replace

plugins/awos/templates/fix-bug-template.md
```diff
@@ -0,0 +1,238 @@
...
+<!-- /awos:flow:stage -->
+
+<!-- awos:flow:stage=commit-push -->
+
+### Step 11: Commit & Push
+
+Write this stage's flow-log entry **before** staging so the log rides in this commit — this is the flow-log's last committed state (see Context Discipline). Then stage only what this flow produced or touched — the delegated fix, the regression test, the flow log, spec/context artifacts, and any Self-Improvement Loop edits. Never a blanket `git add -A`: pre-existing dirty-tree files the workspace stage warned about stay unstaged; surface any unexpected changed file instead of staging it. Never stage `.env`, credentials, or secrets. [Commit message convention per §2, referencing `BUG_ID`; pre-commit hook failures: fix and amend.] Push `BRANCH` to the remote.
+
+<!-- /awos:flow:stage -->
+
```
dustyo-O, line 191: we are not having deliver step for bug fix by purpose?

dustyo-O (dismissed):
reviewed

plugins/awos/commands/flow.md
AlexanderMakarov: Verb-object, to match the sibling: `/implement-feature` and `/fix-bug` are both imperatives, like the rest of the command surface (`/awos:implement`, `/awos:verify`). And it's only a default — the "Command set & names" decision in Step 4 lets the team rename either (`/fix`, `/bugfix`), and the chosen name is recorded in the decision record so re-runs reconcile the right file.

plugins/awos/commands/flow.md
AlexanderMakarov: Relabeled to `Optional Input` to match the core commands' vocabulary (`hire.md`, `architecture.md` use Prerequisite/Optional). The recommendation itself lives in Step 1.3 either way — the command suggests `/awos:hire` when the roster is missing but proceeds without it — so the label wasn't carrying anything the parenthetical doesn't.

Fixed in 4db3671.

plugins/awos/commands/flow.md
AlexanderMakarov: This one I'd keep distinct: these files aren't optional enrichment — their existence is what flips the command into re-run mode (Step 1.4), and once in that mode reading them is mandatory: they carry the recorded decisions and the manual edits Step 6 must reconcile. "Optional" says "use it if it's there, nothing changes if it isn't"; here presence changes the command's whole mode. `Re-run Inputs (if they exist)` states exactly that contract.

plugins/awos/commands/flow.md
AlexanderMakarov: It's compressed from observed failures, not theory — the same road-tests the other rules in this file cite (the dead Jira link, the `core.hooksPath` hook). The generated interview produced exactly the shapes the bullet forbids: an option pair like "Yes — Confluence" / "Yes — local files" where both choices collapse into "now type the location" (that example is lifted from a real draft), and options that couldn't be acted on without follow-up typing. The two behaviors that fixed it are what the bullet pins: options must differ in *behavior*, and the tool's built-in free-text field carries the parameters. Dense, agreed — but each clause maps to one observed failure.

plugins/awos/commands/flow.md
AlexanderMakarov: The rule isn't about Claude misnaming things in code — it targets the interview options, where two concrete failures show up: options naming concepts earlier answers already ruled out (a "transition the ticket" option after the user said tasks come from local files — a generic question bank mentions trackers/PRs unless something forbids it, and nothing else in the prompt does), and shorthand compressed to fit the tight option-label space ("per-run confirm") that the user has to decode mid-interview. `AskUserQuestion` labels are short, which is precisely what pushes the model into shorthand. One bullet, and the failure it prevents is a user staring at an option they can't parse.

plugins/awos/commands/flow.md
AlexanderMakarov: The spots can't be enumerated framework-side because the set is project-derived: which events exist comes out of Step 2's CI mapping and the §4/§5 decisions. One project's pipeline is static checks → CI → human review → merge; another adds an automatic reviewer, soak, compliance, and a deploy step; a repo with no code host has only the local suite and a local `git merge`. A canonical title list would also fight the bullet right above it — options must be written in the project's own event names, not framework vocabulary. So the rule fixes the *shape* (every meaningful stop point, ordered, consistent event names) and lets the investigation supply the content.

plugins/awos/commands/flow.md
AlexanderMakarov: Added. The notifications dimension now records, for the merged/deployed events, that the announcement carries how to try the change on the target environment — the URL or entry point, a feature flag to enable, or the minimal steps — drawn from the ticket/spec, so a teammate can exercise the feature without opening the change request. Mirrored in the decision-record template's §9.

Fixed in 4db3671 + 4134019.

plugins/awos/README.md
AlexanderMakarov: Reworded the README to mark both names as defaults with a note that the team can rename them (4134019). On fixing the name as `/awos:feature`: that would move the command into the plugin namespace, which is framework-owned — identical for every project and overwritten on update — while this command is generated project-specific prose derived from the project's own decision record, deliberately placed outside `.claude/commands/awos/` so framework updates never touch it. The name variance also never reaches the model as ambiguity: re-runs read the actual names from the decision record's **Generated Commands** field, not from the README — the README is human documentation.

plugins/awos/templates/fix-bug-template.md
AlexanderMakarov: Off-by-one, confirmed — fix-bug has an extra early stage, so remote gates is Step 12 here, not 11. Fixed in 5d46dfe. On the template-engine point: the duplication is deliberate at one level (each generated command must be self-contained at run time, so the skeletons can't share includes), but that's an argument for validation, not for accepting drift — a lint check that resolves every "Step N" / stage cross-reference inside each template would catch exactly this class in CI. I'd track that as its own issue rather than grow this PR.

plugins/awos/templates/fix-bug-template.md
AlexanderMakarov: Deliberate, for two reasons the feature flow doesn't have. The log path must be known from the first stage and never move — but a bug's owning spec is only resolved at the classify stage (Step 5), and Step 5 explicitly allows "no owning spec" for legacy or cross-cutting bugs. Parking the log under `context/spec/NNN-*/` would mean either no log until classify, a mid-flow move once the spec resolves, or a homeless log for spec-less bugs. Keying it to `BUG_ID` in `context/` keeps resume a single-glob lookup from the first stage on; when classify does resolve a spec, the log records the spec dir but stays keyed to the bug — the sentence at this line covers that.

plugins/awos/commands/flow.md
AlexanderMakarov: Right — the bug-fix flow's log lands at `context/fix-log-{BUG_ID}.md`, which the `context/spec/**/flow-log.md` glob misses. Added it to the gate-scope list here, and to the Step 8 ignore advice per your other comment. Fixed in 4db3671.

plugins/awos/commands/flow.md
AlexanderMakarov: Added — the Step 8 ignore advice now names `context/fix-log-*.md` alongside the spec flow-logs and the generated command files. Fixed in 4db3671.

plugins/awos/templates/implement-feature-template.md
AlexanderMakarov: Two guards make this bounded. First, the flow never improvises it: this bracketed text is filled from the **Sanctioned verification path** the user records during the `/awos:flow` interview (delivery-flow.md §2) — "stop the service" only appears in a generated command when the user chose that as how verify reclaims the resource; the alternatives (alternate port, throwaway instance, the project's own deploy/run step) are on the same menu. Second, you caught a real drop: `flow.md` says "stop **and restart** the service" and both templates lost the restart half. Aligned both (implement-feature and fix-bug) to "stop and restart", so the user's service comes back whichever path was recorded. Fixed in 5d46dfe + 4134019.

plugins/awos/templates/delivery-flow-template.md
AlexanderMakarov: Swept the leftovers: the decision-record intro ("the project's delivery decisions"), the implement-feature frontmatter description, README's "delivery steps"/"git flow" lines, and the templates' branch-naming and commit-message conventions. Kept "team" where it refers to the people rather than ownership — interviewing the team, the team channel, the team opting in. Fixed in 4134019 (+ 5d46dfe for the fix-bug template).

plugins/awos/templates/fix-bug-template.md
AlexanderMakarov: Not on purpose — a gap. `flow.md` says the bug-fix flow reuses the §5 delivery decisions ("merge — including its §5 delivery decisions"), but the template had no stage to run the deployment part in: a §5 that has the flow deploy after merge — typically a hotfix, exactly where it matters — had nowhere to execute. Added a delivery stage between merge and close, mirroring the feature template's (filled per §5, omitted when the decisions rule it out), and renumbered the close stage. Fixed in 5d46dfe.

workshur (approved):

PR #136, opened 2026-06-12
fix(tasks): mark unreviewed tasks.md with a file-level marker, cleared on review

### LINKED ISSUE #135: tasks: Step 4 draft-review gate stalls headless runs — tasks.md never written when the feedback ask is plain text
## Summary

`commands/tasks.md` Step 4 ("Present Draft and Refine") presents the drafted plan and asks for feedback **before** Step 5 writes `tasks.md`. The ask is open-ended, so Claude nondeterministically phrases it either as an `AskUserQuestion` tool call or as a **plain-text question**. Under headless `claude -p` the two behave very differently:

- `AskUserQuestion` → silently dismissed, Claude proceeds, `tasks.md` is written ✅
- plain text question → the turn simply ends; `tasks.md` is **never written** ❌

This is the #131 class of bug (deliverable gated on user review) resurfaced in `/awos:tasks` — but in a worse form: a plain-text question cannot be intercepted by hooks or any answer-map mechanism, because there is no tool call to intercept.

## Evidence (2026-06-12, awos main `a083aa5`, claude 2.1.175)

Three unattended runs of awos-qa tasks scenarios at the same commit, same conditions (`claude -p --dangerously-skip-permissions --setting-sources project --effort medium`), all fixtures seeding a project-local QA agent so the (intended) `QA Agent` hire gate does not fire:

| Scenario | Result | How the Step 4 ask surfaced |
|---|---|---|
| `tasks-discovers-plugin-agents` | pass | `AskUserQuestion` → unanswered dismissal → proceeded to `Write` |
| `tasks-enumerates-agents` | **fail** | plain text — session's last output, no `Write` ever happened |
| `tasks-honors-skip-tests-intent` | **fail** | plain text — identical signature |

Final assistant message of a failing run (full task plan was drafted in chat, then):

> Does this plan look good, or would you like to adjust any slices, tasks, or agent assignments before I write `tasks.md`?

Tool-call trace of the failing runs: `Bash → Read → Read` — analysis completed, plan fully drafted, zero writes.

So at the current main, every `/awos:tasks` invocation under `-p` is a coin flip on which phrasing Claude picks. The line-31 guidance ("Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text") doesn't cover Step 4 because "ask for feedback" is not multiple-choice.

## Suggested fixes (any one resolves it)

1. **Write-then-review** (the #131 fix pattern): Step 4 saves `tasks.md` first, then asks for feedback and applies refinements as edits to the saved file. Headless runs keep the deliverable; interactive runs keep the review loop.
2. **Mandate the tool**: Step 4 explicitly requires the feedback ask to go through `AskUserQuestion` (e.g. options "Looks good — save it" / "I want changes"). Dismissal under `-p` then degrades gracefully, and answer-maps/hooks can answer it deterministically.
3. **Headless guard**: "If no user response can be obtained, save `tasks.md` as drafted and report." 

Option 1 is the most robust — it removes the failure mode instead of routing around it.

## Context

- Found while validating the prompt-eval plan in awos-qa (provectus/awos-qa#16); the harness scenarios and traces above are reproducible via `bun run e2e:run tasks-enumerates-agents`.
- Related: #131 (fixed), #109 (introduced the current Step 3/4 flow, including the — intended and fine — `QA Agent` hire gate; the Step 4 ordering is the only headless problem).

FlySpot:
Closes #135.

> **Scope note:** #135's headline ask — write-then-review so the deliverable survives a headless turn — already landed in #132. This PR closes the residual gap: a saved `tasks.md` that nobody reviewed was indistinguishable from a reviewed one.

## Change

A file-level marker on `tasks.md`, mirroring the existing `<!-- skip-tests: true -->` contract:

- **Step 4** writes `<!-- not-user-reviewed -->` at the top of `tasks.md`. The file is written before review, so it starts as a draft.
- **Step 5** removes the marker once the user reviews the plan — on approval, or after their requested changes are applied and re-saved. If the review question is dismissed or the run is non-interactive, the marker stays; its presence is the draft signal.
- Review feedback is asked **strictly via `AskUserQuestion`**, never plain text (a plain-text ask ends a non-interactive turn before the review outcome can be recorded).

### Why a file marker (this PR's earlier revision used a chat-report line)

Per review feedback, the marker moved from the final chat report into the saved file. A file-level HTML-comment marker is pure ASCII (no fragile em-dash join key), persisted in the artifact, **grep-detectable deterministically** instead of parsed from freeform chat, and visible to anyone opening the file later. It's the same mechanism `/awos:verify` already relies on for `<!-- skip-tests: true -->`.

## Test (Layer 1)

New `tests/lint-prompts.test.js` case locks both the marker shape and the removal-on-review instruction — per `CLAUDE.md`, a grep-catchable contract ships its test in the same PR (the adjacent `skip-tests` marker is locked the same way). `npm run test:lint` → 32/32.

## Validation (live, claude 2.1.175, effort medium, branch on current main)

Both contract branches exercised end-to-end via the awos-qa harness:

| Branch | Setup | tasks.md marker | verify |
|---|---|---|---|
| Headless / no review | `claude -p`, review question dismissed | **present** (Claude left it, citing the rule) | 4/4 pass |
| Review approved | approval answered via a `PreToolUse` answer-map hook | **removed** (0 occurrences) | — |

## Summary by CodeRabbit

* **Documentation**
  * Updated the “Surface for Review and Recommend Next Step” workflow to collect all review feedback exclusively via the AskUserQuestion tool. Added draft-state handling using a `<!-- not-user-reviewed -->` marker that is removed only after a completed review, while unanswered/dismissed reviews preserve the draft marker and reflect that status in the final report.

### DIFF
--- a/commands/tasks.md
+++ b/commands/tasks.md
@@ -122,12 +122,16 @@ Skip this step if `SKIP_TESTS = true`.
 ## Step 4: Write the Task List
 
 1.  Write the complete slice/task list to `tasks.md` in the chosen spec directory. **Write the file without waiting for approval** — generating a task list is reversible (re-run `/awos:tasks` to revise), so the deliverable must never be gated behind a confirmation that an unattended run cannot answer.
-2.  If `SKIP_TESTS = true`, record a one-line note at the top of the generated `tasks.md` so that downstream commands (e.g. `/awos:verify`) can detect the choice: `<!-- skip-tests: true -->`.
+2.  Record a one-line marker at the very top of the generated `tasks.md`: `<!-- not-user-reviewed -->`. The file is written before review (Step 5), so it starts as a draft; Step 5 removes this marker once the user has reviewed it. Keep the marker shape exactly — downstream automations grep for it to tell a draft from a reviewed plan.
+3.  If `SKIP_TESTS = true`, record a one-line note at the top of the generated `tasks.md` so that downstream commands (e.g. `/awos:verify`) can detect the choice: `<!-- skip-tests: true -->`.
 
 ## Step 5: Surface for Review and Recommend Next Step
 
-1.  Report the saved path and present the slice/task plan for review. If the user requests changes (adjust, split, merge slices or tasks, or reassign subagents), apply them and re-save; otherwise they can revise later by re-running `/awos:tasks`.
-2.  If any tasks were assigned to `general-purpose` (because no specialist exists) or verification cannot be performed (missing MCPs/services), surface a table:
+1.  Report the saved path and present the slice/task plan for review.
+2.  Ask for review feedback strictly via the `AskUserQuestion` tool (e.g. options "Looks good — keep it as saved" / "I want changes") — never in plain text, which would end a non-interactive turn before the review outcome can be reported.
+3.  **When the user responds** (either "looks good" or after you apply their requested changes and re-save): remove the `<!-- not-user-reviewed -->` marker from the top of `tasks.md`, since the plan has now been reviewed. If they requested changes, apply them — adjust, split, merge slices or tasks, or reassign subagents — and re-save before removing the marker.
+4.  **If the review question goes unanswered** (dismissed, or the run is non-interactive): leave the file exactly as saved, marker included. Its presence is the signal that the plan was written but not reviewed; do not remove it.
+5.  If any tasks were assigned to `general-purpose` (because no specialist exists) or verification cannot be performed (missing MCPs/services), surface a table:
 
     | Task/Slice            | Issue                                                                               | Recommendation                                       |
     | --------------------- | ----------------------------------------------------------------------------------- | ---------------------------------------------------- |
@@ -135,4 +139,4 @@ Skip this step if `SKIP_TESTS = true`.
     | Slice N (QA)          | Feature Testing & Regression slice uses `general-purpose` — no QA-coded agent hired | Run `/awos:hire` to install `testing-expert`         |
     | Slice 3: Verification | Browser MCP not available                                                           | Install browser MCP to enable UI verification        |
 
-3.  Report the next command: `/awos:implement`.
+6.  Report the next command: `/awos:implement`.
--- a/tests/lint-prompts.test.js
+++ b/tests/lint-prompts.test.js
@@ -554,6 +554,24 @@ test('commands/tasks.md documents the skip-tests opt-out and persists it', () =>
   );
 });
 
+test('commands/tasks.md marks an unreviewed tasks.md and clears it on review', () => {
+  // tasks.md is written before review (Step 4), so it starts as a
+  // draft carrying a "<!-- not-user-reviewed -->" marker that Step 5
+  // removes once the user reviews it. The marker shape is the contract
+  // — awos-qa greps the saved file to tell a draft from a reviewed
+  // plan, so a reword here would silently break that detection. Lock
+  // the shape, plus the removal-on-review instruction.
+  const body = readUtf8(path.join(commandsDir, 'tasks.md'));
+  assert.ok(
+    body.includes('<!-- not-user-reviewed -->'),
+    'commands/tasks.md must record the literal "<!-- not-user-reviewed -->" marker so awos-qa can detect a draft-grade tasks.md'
+  );
+  assert.ok(
+    /remove the `<!-- not-user-reviewed -->` marker/i.test(body),
+    'commands/tasks.md Step 5 must remove the not-user-reviewed marker once the plan has been reviewed'
+  );
+});
+
 test('commands/verify.md acknowledges the skip-tests marker', () => {
   // The Slack thread feedback frames /awos:verify as look-and-feel +
   // spec-freshness rather than a test runner. The skip-tests marker


commands/tasks.md
```diff
@@ -126,13 +126,15 @@ Skip this step if `SKIP_TESTS = true`.
...
-2.  If any tasks were assigned to `general-purpose` (because no specialist exists) or verification cannot be performed (missing MCPs/services), surface a table:
+1.  Report the saved path and present the slice/task plan for review.
+2.  Ask for review feedback strictly via the `AskUserQuestion` tool (e.g. options "Looks good — keep it as saved" / "I want changes") — never in plain text, which would end a non-interactive turn before the review outcome can be reported. If the user requests changes (adjust, split, merge slices or tasks, or reassign subagents), apply them and re-save; otherwise they can revise later by re-running `/awos:tasks`.
+3.  If the review question goes unanswered (dismissed, or the run is non-interactive), keep the saved file as-is and include this exact line in your final report: `tasks.md saved as drafted — not user-reviewed`. Keep the wording — downstream automations depend on it to tell a reviewed plan from a draft.
```
AlexanderMakarov: You've added a marker that downstream automation greps for (`tasks.md saved as drafted — not user-reviewed`), and the PR body says the wording is load-bearing for awos-qa. Two things:

1. **No Layer 1 test guards it.** The `<!-- skip-tests: true -->` marker a few items down is locked by `tests/lint-prompts.test.js:542` ("The marker shape is part of the contract — if it moves, verify.md will not detect it. Lock the shape here."), and `CLAUDE.md` says a surface-area contract a grep can catch ships its test in the same PR. The behavioral check lives in awos-qa, but nothing in *this* repo stops an innocent reword of this line from silently drifting from what the harness expects. A one-line assertion mirroring the skip-tests test would close it:

   ```js
   assert.ok(
     body.includes('tasks.md saved as drafted — not user-reviewed'),
     'commands/tasks.md must keep the literal unreviewed-output marker so awos-qa can detect draft-grade tasks.md'
   );
   ```

2. **The em-dash makes the marker a fragile join key.** Unlike the pure-ASCII `<!-- skip-tests: true -->`, this string carries a `—` that a model emitting the final report can easily render as `-` or `--`, and it's hard to eyeball the difference when a harness match fails. Worth considering an ASCII form. If you keep the em-dash, the Layer 1 pin from (1) at least makes the exact glyph load-bearing in the test.

AlexanderMakarov (commented):
This is well-scoped. #135's headline ask — write-then-review so the deliverable survives a headless turn — already landed in #132, and this PR correctly limits itself to the residual gap: a saved tasks.md that nobody reviewed is indistinguishable from one that was. The tasks.md saved as drafted — not user-reviewed marker plus the "ask only through `AskUserQuestion`" rule cover that, and it sits cleanly under the INTERACTION guard (an unanswered review question falls back to "keep as saved + emit marker" rather than stalling). The Step 4/5 renumber is correct and nothing else references the old item numbers.

One thing I'd like resolved before merge: you've introduced a pinned, machine-detected marker string but no Layer 1 test pins it — the directly analogous `<!-- skip-tests: true -->` marker right below it is locked by a lint test, and CLAUDE.md says a grep-catchable contract ships its test in the same PR. Detail inline, plus a small robustness note on the marker glyph.

I considered CodeRabbit's two outside-diff notes and left both: the "remove the pre-save QA gate" (Step 3a) and "Step 3.4 is stale" points target pre-existing lines this PR doesn't touch, and the QA-gate concern is already handled by the post-#132 INTERACTION fallback (a skipped hire question defaults to general-purpose), so neither is in scope here.

commands/tasks.md
FlySpot: Both points addressed in 75d2697.

1. **Layer 1 test** — added in `tests/lint-prompts.test.js`, mirroring the skip-tests lock: pins the marker shape and the removal-on-review instruction. `npm run test:lint` → 32/32.

2. **Em-dash fragility** — rather than ASCII-ize the chat line, the marker moved out of the chat report and into the saved file as `<!-- not-user-reviewed -->`, directly analogous to `<!-- skip-tests: true -->`: pure ASCII, persisted in the artifact, grep-detectable deterministically (not parsed from freeform chat), and visible to anyone opening the file later. Step 4 writes it by default; Step 5 removes it once the plan is reviewed.

Validated live both ways: a headless (dismissed-review) run keeps the marker (verify 4/4), and an approval answered via a PreToolUse hook removes it (0 occurrences).

On the two CodeRabbit out-of-scope notes — agreed they target pre-existing lines this PR doesn't touch; leaving them for a separate change.

commands/tasks.md
```diff
@@ -122,17 +122,21 @@ Skip this step if `SKIP_TESTS = true`.
 ## Step 4: Write the Task List
 
 1.  Write the complete slice/task list to `tasks.md` in the chosen spec directory. **Write the file without waiting for approval** — generating a task list is reversible (re-run `/awos:tasks` to revise), so the deliverable must never be gated behind a confirmation that an unattended run cannot answer.
-2.  If `SKIP_TESTS = true`, record a one-line note at the top of the generated `tasks.md` so that downstream commands (e.g. `/awos:verify`) can detect the choice: `<!-- skip-tests: true -->`.
+2.  Record a one-line marker at the very top of the generated `tasks.md`: `<!-- not-user-reviewed -->`. The file is written before review (Step 5), so it starts as a draft; Step 5 removes this marker once the user has reviewed it. Keep the marker shape exactly — downstream automations grep for it to tell a draft from a reviewed plan.
```
AlexanderMakarov, line 125: The `<!-- not-user-reviewed -->` marker now shares "the top" with `<!-- skip-tests: true -->` (Step 4.3) whenever `SKIP_TESTS = true`. Both minor:

- Step 5.3 says "remove the `<!-- not-user-reviewed -->` marker from the top of `tasks.md`" — naming the exact string is the right call, but worth stating explicitly to leave any `<!-- skip-tests: true -->` line intact. That one has to survive review, and a less careful pass could read "remove the top comment" as the whole block.
- On write, the relative order of the two is unspecified. Harmless for grep (it's position-independent), but pinning "not-user-reviewed first, skip-tests second" reads cleaner.

Optional hardening, not a blocker.

commands/tasks.md
AlexanderMakarov: Both addressed cleanly in `75d2697` — thanks. Moving the marker into the saved file as `<!-- not-user-reviewed -->` is a better fix than the ASCII chat-line I floated: persisted in the artifact, deterministic to grep, visible to anyone opening the file.

The new test checks two things: that the marker string is present, and that some "remove the marker" instruction exists. What it doesn't check is the order that actually matters — Step 4 writes the marker before review, and Step 5 removes it only after the user responds. So someone could later move the write to after review, or make Step 5 always delete it, and the test would still pass.

The real behavioral check belongs in awos-qa. But Layer 1 owns the prompt text, so it can go a bit further cheaply: assert that the write lives in Step 4 and that the removal is tied to a review response (e.g. the text "once the user has reviewed" or "When the user responds"). That closes the gap without having to parse sections.

AlexanderMakarov (approved):
Re-review after `75d2697`. Both points from my earlier pass are addressed, and well.

Two non-blocking follow-ups, neither a merge blocker:
1. The new lint test pins that the marker *exists* and that *a* removal instruction exists, but not the lifecycle that's the actual contract — written in Step 4 *before* review, removed *only* once the user responds (reply on my earlier thread).
2. Minor: `<!-- not-user-reviewed -->` and `<!-- skip-tests: true -->` now share "the top", so Step 5's removal is worth scoping explicitly to the one line (inline).

Happy to approve as-is.

PR #137, opened 2026-06-15
feat: brownfield awareness across product, roadmap, and architecture

dustyo-O:
## Summary

- `/awos:product` detects brownfield projects (source code indicators like `src/`, `package.json`, etc.) during Creation Mode, runs an Explore agent focused on purpose/audience/features, triages each finding with user (accept / correct / reject), writes accepted findings to `context/product/brownfield.md`
- `/awos:roadmap` reads brownfield.md, runs a focused Explore for existing capabilities (passing prior findings via `<existing_findings>` to avoid duplicates), appends `## Capabilities` section
- `/awos:architecture` reads brownfield.md, runs a focused Explore for tech stack, appends `## Technology` section
- `/awos:architecture` removes brownfield.md at the end of its run (Step 5: Brownfield Cleanup) — all knowledge is absorbed into product-definition.md, roadmap.md, architecture.md
- 6 new lint tests covering the brownfield lifecycle (creation, appending, removal, dedup contract, triage UX)

Supersedes the approaches on `feat/scan-command` and `feat/scan-embedded` — no new files, no new commands, no new templates.

## Design decisions

- **Creation Mode only**: brownfield detection runs on first `/awos:product` run only. Re-running in Update Mode skips it intentionally — by then `context/spec/` holds feature knowledge and brownfield.md would be redundant.
- **Split exploration**: each command owns one domain (product → capabilities → technology) rather than one big scan, so findings are relevant to the command's interview.
- **Accumulated dedup**: each Explore agent receives prior brownfield.md contents in `<existing_findings>` tags to avoid re-discovering the same things.

## Test plan

- [x] `npm test` — 84/84 pass (6 new brownfield contract tests)
- [x] `npx prettier . --check` — clean
- [x] Manual: run `/awos:product` on a brownfield project, verify exploration + triage + brownfield.md creation
- [x] Manual: run `/awos:roadmap` after, verify it reads brownfield.md and explores capabilities without duplicating product findings
- [x] Manual: run `/awos:architecture` after, verify it reads brownfield.md and explores tech stack
- [x] Manual: run `/awos:hire` after, verify brownfield.md is deleted

## Summary by CodeRabbit

* **Documentation**
  * Clarified the “running on an existing codebase” setup guidance in the README.
  * Expanded the document-centric workflow description to better explain how project state is rehydrated from saved context.
  * Updated creation-mode workflow documentation to introduce brownfield discovery, confirmation, and cleanup behavior.
* **Tests**
  * Added/extended prompt-lint assertions to validate brownfield awareness, deduplication, incremental accumulation, and the expected file lifecycle across creation commands.


commands/product.md
```diff
@@ -62,14 +62,37 @@ First, check if the file `context/product/product-definition.md` exists.
...
+    - Purpose and problem being solved (README, docs, package metadata, comments)
+    - Target audience signals (UI copy, API design, documentation tone, onboarding flow)
+    - Main features and capabilities (entry points, routes, commands, key modules)
+    - User journey (how someone uses this from start to finish)
+
+    For each finding, cite the file paths that evidence it. Be concise — report findings as bullet points.
+    ")
+    ```
+
+    b. Walk through the findings one by one — each discovered entity (purpose, audience signal, feature, journey step) gets its own confirmation. For each finding, use `AskUserQuestion` with three options:
```
AlexanderMakarov: This walks findings "one by one — each discovered entity … gets its own confirmation." On a real codebase Explore can easily surface a dozen-plus findings (purpose, several audience signals, each feature, each journey step), which turns into a dozen-plus sequential `AskUserQuestion` round-trips before the interview even starts. Two things worth considering:

1. `AskUserQuestion` takes up to four questions per call — batching related findings (e.g. all the features in one call) cuts the round-trips a lot without losing the per-finding accept/reject granularity. `roadmap.md` and `architecture.md` inherit this via "same as `/awos:product` does", so fixing it here fixes all three.
2. "Accept with corrections — if chosen, ask the user for their correction" (line 83) forces an extra turn to collect the correction, which is exactly what `AskUserQuestion`'s built-in free-text "Other" field already does in one shot. Accept / Reject plus the user typing a correction via Other may get the same outcome with less ceremony.

Similar in other commands "brownfield exploration" sections - consider big codebase and number of questions it may spark to user.

commands/architecture.md
```diff
@@ -49,12 +49,42 @@ Follow this logic precisely.
...
+    ... brownfield.md contents ...
+    </existing_findings>
+
+    Report only NEW findings not covered above. For each technology found, cite the file paths that evidence it. Be concise — report findings as bullet points.
+    ")
+    ```
+
+    c. Walk through any new findings with `AskUserQuestion` (Accept / Accept with corrections / Reject), same as earlier commands. Append accepted findings to `context/product/brownfield.md` under a `## Technology` heading. For corrected findings, record the corrected version.
+
+    d. Use the confirmed technology findings to pre-fill the architecture template. Frame the section-by-section walkthrough as "here's what you're using — confirm, and tell me what you want to change or add for the roadmap ahead."
```
AlexanderMakarov: Step 2d says to "pre-fill the architecture template" and frame the walkthrough as "here's what you're using — confirm…", but step 3 right below still requires proposing alternatives and confirming each section. `product.md` states the principle deliberately: *"The interview still covers every section; the exploration gives better defaults, not fewer questions."* "Pre-fill the template" reads like an invitation for the agent to treat the brownfield tech as settled and skim the section-by-section confirmation. Mirroring product.md's phrasing here ("use the confirmed tech as the **default** for each section, but still walk every section") keeps the two commands aligned.

commands/roadmap.md
```diff
@@ -51,9 +51,36 @@ Follow this logic precisely.
...
+    Agent(subagent_type="Explore", description="Assess existing capabilities", prompt="
+    Explore this codebase and assess what's already built. Focus on:
+    - Features that are fully implemented and working (with evidence: routes, UI, tests)
+    - Features that appear partially implemented or scaffolded
+    - TODOs, FIXMEs, or planned features mentioned in code or docs
+
+    The following findings were already confirmed by the user — do not repeat them:
+
+    <existing_findings>
+    ... brownfield.md contents ...
```
AlexanderMakarov: The `... brownfield.md contents ...` placeholder inside this fenced prompt relies entirely on the instruction in (b) to get substituted. An ellipsis placeholder in an otherwise copy-paste-ready prompt block is exactly the kind of thing a model sometimes passes through verbatim. A fill-in slot reads less ambiguously, e.g. `{paste the full current contents of context/product/brownfield.md here}`. `architecture.md:71` has the identical placeholder. Minor — the preceding instruction is explicit — but cheap to tighten.

It would be great to cover this logic with some unit/e2e tests here or in awos-qa.

tests/lint-prompts.test.js
```diff
@@ -657,6 +657,105 @@ test('SDD-07 recognizes the dual-model QA coverage', () => {
   );
 });
 
+// ---------------------------------------------------------------------------
+// Brownfield awareness contracts
+// ---------------------------------------------------------------------------
+
+test('product.md creates context/product/brownfield.md on brownfield detection', () => {
```
AlexanderMakarov, line 664: The six tests lock in the headings, the `<existing_findings>` tag, the triage words, and the cleanup language — good coverage of the contract surface. But the load-bearing behavior is largely untested, and the failure mode is the bad one: green tests over a dead feature.

1. **No test asserts a command actually launches the `Explore` agent.** Delete the `Agent(subagent_type="Explore", …)` block but keep the headings and triage prose and all six tests still pass — shipping a brownfield feature that explores nothing. A loop asserting `/subagent_type="Explore"/` across product/roadmap/architecture guards the actual mechanism, and per CLAUDE.md's "test ships with the contract" rule this is the contract most worth pinning.
2. **No test guards `product.md`'s detection indicators.** The whole entry point is `product.md` detecting source-code indicators (`src/`, `package.json`, `go.mod`, `Cargo.toml`, `pyproject.toml`, …). Nothing asserts that list exists — hollow it out and brownfield never fires, the whole roadmap/architecture/hire chain goes dead, and all six tests still pass. Assert a representative subset (e.g. `package.json` and `src/`).
3. **The heading greps are unanchored and decoupled from the `brownfield.md` reference.** `/## Product/`, `/## Capabilities/`, `/## Technology/` are each asserted independently of the path reference, so they prove the file mentions the heading name somewhere in prose, not that the append instruction is wired up. Matching the heading and `brownfield.md` in one assertion (e.g. `` /under a `## Technology` heading/ ``) makes it a real contract.
4. **The dedup directive is unasserted.** `<existing_findings>` is checked, but the "Report only NEW findings … do not repeat them" instruction that makes the tag *do* anything isn't — a grep for `Report only NEW` in roadmap+architecture locks it in.
5. **No inverse-contract guard on `product.md`.** The design depends on product being first — it must *not* consume `<existing_findings>`. `assert.ok(!/<existing_findings>/.test(productBody))` pins that cheaply.

tests/lint-prompts.test.js
```diff
@@ -657,6 +657,105 @@ test('SDD-07 recognizes the dual-model QA coverage', () => {
...
+    body.includes('context/product/brownfield.md'),
+    'commands/architecture.md must reference context/product/brownfield.md to consume and extend brownfield findings'
+  );
+  assert.ok(
+    /## Technology/.test(body),
+    'commands/architecture.md must append brownfield findings under a "## Technology" heading'
+  );
+});
+
+test('hire.md removes brownfield.md after onboarding completes', () => {
```
AlexanderMakarov: `/delete|remove/i.test(body.substring(body.indexOf('context/product/brownfield.md')))` works today because hire mentions the path once, but `remove` matches loosely (unrelated prose like "removes the need for…" would satisfy it) and `.substring(indexOf())` quietly assumes a single mention. More to the point, the real contract is that the deletion is *guarded* by an existence check so re-running hire on a greenfield project is a no-op — that invariant isn't pinned. Something like `` /If `context\/product\/brownfield\.md` exists, (delete|remove) it/i `` asserts the conditional shape. (Minor — fold into the test-coverage comment above if you touch the tests at all.)

README.md
```diff
@@ -21,7 +21,7 @@ npx @provectusinc/awos
 
 This sets up the `.awos/` directory (commands, templates, scripts), the `.claude/commands/awos/` wrappers, and the `context/` directory where your project documents will live. It also registers the AWOS plugin marketplace in your project settings.
 
-> **Running on an existing codebase?** Start with an AI readiness audit to understand how AI-friendly your project is. Install the plugin with `/plugin install awos@awos-marketplace`, then run `/awos:ai-readiness-audit` to get a scored assessment with actionable recommendations for improvement. [Learn more](plugins/awos/README.md)
+> **Running on an existing codebase?** AWOS auto-detects brownfield projects. `/awos:product` explores what your project does, `/awos:roadmap` maps what's already built, and `/awos:architecture` discovers your tech stack — each presenting findings for you to accept, correct, or reject before continuing the normal interview. You can also run an AI readiness audit: install the plugin with `/plugin install awos@awos-marketplace`, then run `/awos:ai-readiness-audit`. [Learn more](plugins/awos/README.md)
```
AlexanderMakarov: > You can also run an AI readiness audit ...

I don't think that this brownfield exploration replaces `/awos:ai-readiness-audit` somehow. Because AI readiness is measured by AI tools adoption in the repo, while added brownfield exploration just simplifies AWOS integration (which is already started if user reads README.md) and doesn't provide insights to the user per se.
Please revert text with AI readiness audit explanation back and explain about brownfield exploration after this, to deliver "we promise that integration of AWOS would be seamless for you" mostly, not to explain how it works inside.

AlexanderMakarov (commented):
Clean, well-scoped addition. The split-exploration design — product → capabilities → tech, each command owning one domain and passing prior findings forward to dedup — is the right shape, and the lifecycle reads coherently end to end: `/awos:product` creates `brownfield.md`, `/awos:roadmap` and `/awos:architecture` extend it, `/awos:hire` deletes it once the knowledge is absorbed. Creation-Mode-only scoping is the right call, the `Explore` subagent is exactly the convention CLAUDE.md prescribes for the read-heavy scan, and tests ship with the contracts.

Nothing here blocks merge. A few things worth a look before it goes in:

- The new lint tests pin the *vocabulary* of the feature (headings, tags, triage words) but not its *mechanism* — most notably nothing asserts the commands actually launch an `Explore` agent, and nothing guards `product.md`'s detection indicators. A regression that hollowed out either could ship a no-op feature with all six tests green (inline on the test file).
- `architecture.md`'s "pre-fill the template" wording pulls against the deliberate "better defaults, not fewer questions" principle that `product.md` sets (inline).
- The per-finding triage flow is heavier than it needs to be on a real codebase (inline).

**On INPUTS/OUTPUTS (agreeing with CodeRabbit, plus one it missed):** `brownfield.md` is now read by `/awos:roadmap` and `/awos:architecture` and deleted by `/awos:hire`, but none declare it in their **INPUTS & OUTPUTS** sections — the repo convention is that every file a command touches is declared there (`Prerequisite Input` / `Optional Input` / `Output`), and there's a lint contract for `context/<path>` consistency. A one-line `- **Optional Input:** context/product/brownfield.md (produced by /awos:product …)` in each keeps the data flow legible. The one CodeRabbit didn't call out: `product.md` *writes* `brownfield.md` but its OUTPUTS section lists only `product-definition.md` — add it there too, marked optional. Those sections are unchanged by this PR (out of the diff), so noting here rather than inline.

commands/product.md
```diff
@@ -62,14 +62,34 @@ First, check if the file `context/product/product-definition.md` exists.
...
+    - Main features and capabilities (entry points, routes, commands, key modules)
+    - User journey (how someone uses this from start to finish)
+
+    For each finding, cite the file paths that evidence it. Be concise — report findings as bullet points.
+    ")
+    ```
+
+    b. Triage findings with the user. Group related findings by category (e.g. all features in one call, audience signals in another) and use `AskUserQuestion` to batch up to four per call. For each finding, offer **Accept** and **Reject** as options. The user can also select "Other" to provide free-text feedback — treat it according to intent (correction, substitution, partial accept, or any other reaction). Discard rejected findings.
+
+    c. Write all accepted and corrected findings to `context/product/brownfield.md` under a `## Product` heading. For corrected findings, record the corrected version, not the original. Downstream commands (`/awos:roadmap`, `/awos:architecture`) will append their own findings to this file.
```
AlexanderMakarov: Quick question on the lifecycle: this step only writes `brownfield.md` from *accepted/corrected* findings, and `roadmap.md`/`architecture.md` gate their own explorations on `if brownfield.md exists`. So on a brownfield repo where the user happens to reject every product-level finding (or Explore surfaces nothing product-worthy), the file never gets created — and the capabilities scan and the tech-stack scan, which are independent of the product findings and arguably the highest-value parts on an existing codebase, both silently skip. Is that intended? If not, the cheap fix is to have `/awos:product` always create `brownfield.md` once a codebase is detected (even with an empty `## Product` section, as a "this is brownfield" sentinel); the alternative is to have the downstream commands re-run the same indicator detection instead of keying off the file.

commands/architecture.md
```diff
@@ -49,12 +49,42 @@ Follow this logic precisely.
...
+
+    <existing_findings>
+    {paste the full current contents of context/product/brownfield.md here}
+    </existing_findings>
+
+    Report only NEW findings not covered above. For each technology found, cite the file paths that evidence it. Be concise — report findings as bullet points.
+    ")
+    ```
+
+    c. Triage new findings with `AskUserQuestion`, same as `/awos:product` does — batch by category, offer Accept / Reject, user can provide free-text via "Other". Discard rejected findings. Append accepted findings to `context/product/brownfield.md` under a `## Technology` heading.
```
AlexanderMakarov: > same as `/awos:product` does 

When `/awos:architecture` runs, only architecture.md is in context — product.md isn't loaded (each command is its own prompt). The other `/awos:` mentions in these files are navigational — "run `/awos:product` first", "next command: `/awos:roadmap`" — they tell the *user* what to type; this one points at another command's *internal behavior* as something to replicate, which the model can't follow. For the triage mechanics it's harmless since they're spelled out right here, so the reference is just redundant decoration. But it drops one thing concretely: product.md's "record the corrected version, not the original" for a finding amended via "Other" lives *only* in product.md, so a correction made in the roadmap/architecture phase reads as falling through. (`roadmap.md:77` has the identical line.) Inlining what must hold ("append accepted **and corrected** findings, recording the corrected version") and dropping the cross-reference — or demoting it to a parenthetical aside — fixes both: no dangling pointer, and corrections survive.

tests/lint-prompts.test.js
AlexanderMakarov: Nicely strengthened — the Explore-launch assertion, the indicator guard, the inverse `product.md` guard, and the conditional-delete shape all land. Same theme, one notch deeper: a couple of contracts are still pinned by vocabulary rather than behavior, so a regression could ship green —

- The fill-in slot you just added (`{paste the full current contents of context/product/brownfield.md here}`) is the thing that makes `<existing_findings>` non-empty, but no test asserts it — delete that line and dedup silently dies while `<existing_findings>` + "Report only NEW" both still match. An assertion on the interpolation phrase would lock the fix.
- The `## Capabilities` / `## Technology` checks assert the heading string exists but not that it's *appended to* — a future edit swapping "Append accepted findings to … under `## X`" for "Write …" (clobbering the prior sections) keeps the tests green. Asserting the append verb + heading together protects the accumulate-don't-overwrite contract.
- Nothing pins the existence-guard (`Check if brownfield.md exists … If it does:`) that makes greenfield skip the scan — worth a line per command.

All optional follow-ups, not blockers.

AlexanderMakarov (commented):
Thanks for the thorough fixup — every thread from the last round is addressed.

A few small things remain:
- **The brownfield file still isn't declared in any command's INPUTS & OUTPUTS section** (carryover from last round, also flagged by CodeRabbit). `product.md` writes it but its OUTPUTS lists only `product-definition.md`; `roadmap.md` and `architecture.md` read *and* append to it without listing it; `hire.md` reads-and-deletes it without listing it. For a spec-driven framework where those I/O blocks are the documented contract, an undeclared cross-command data file is the kind of gap that bites a later editor. Worth one `Optional Input` / `Output` line in each of the four. (Those sections are unchanged by this PR, so noting here rather than inline.)
- Reliable creation of _brownfield.md_ in `/awos:product` to don't break process later.
- _same as `/awos:product` does_ references which won't give content of one command to others.
- follow-up on the test file thread.

AlexanderMakarov (approved):

commands/product.md
```diff
@@ -62,14 +63,34 @@ First, check if the file `context/product/product-definition.md` exists.
...
+
+    For each finding, cite the file paths that evidence it. Be concise — report findings as bullet points.
+    ")
+    ```
+
+    b. Triage findings with the user. Group related findings by category (e.g. all features in one call, audience signals in another) and use `AskUserQuestion` to batch up to four per call. For each finding, offer **Accept** and **Reject** as options. The user can also select "Other" to provide free-text feedback — treat it according to intent (correction, substitution, partial accept, or any other reaction). Discard rejected findings.
+
+    c. Create `context/product/brownfield.md` with a `## Product` heading. List all accepted and corrected findings under it — for corrected findings, record the corrected version, not the original. If every finding was rejected or the exploration surfaced nothing, still create the file with an empty `## Product` section; downstream commands (`/awos:roadmap`, `/awos:architecture`) key on the file's existence to run their own explorations.
+
+2.  If `<user_prompt>` is non-empty, briefly note that you'll use it as a starting point, then refine from there.
```
workshur, line 86: I believe we should prioritize <user prompt> and make "Brownfield detection" optional if <user prompt> is present. We can ask the user if detection is necessary.

dustyo-O: you mean, prioritize user prompt over brownfield detection and skip brownfield path if prompt explicitly denied that?

dustyo-O: ah, if you meant we have to ask user if he really wants to start brownfield path -> I've pushed fixup with that

PR #138, opened 2026-06-23
feat: external documentation gathering across onboarding commands

dustyo-O:
## Summary

- Extends the brownfield onboarding path to gather project knowledge from external documentation sources (wikis, ticket systems, chats, email) using the same accumulate-then-absorb pattern as `brownfield.md`
- `/awos:product` detects sources, invokes the `configure-external-sources` plugin skill to help set up access (MCP/CLI/manual), then retrieves product-relevant findings; `/awos:roadmap` and `/awos:architecture` run their own focused retrievals from the shared `context/sources/sources.md` manifest
- Platform setup guides live in `plugins/awos/skills/configure-external-sources/references/` (documentation, tickets, communication) — used as a knowledge base by the skill, not shipped as commands
- Privacy gate for chat/email sources, restart-resume flow for MCP setup, graceful degradation when the plugin is not installed

## Test plan

- [ ] `npm test` — all 113 tests pass including external-sources contracts
- [ ] `npx prettier . --check` — formatting clean
- [ ] Run `/awos:product` on a brownfield project → verify external documentation source detection triggers after brownfield detection
- [ ] Verify manual source flow: user pastes content, retrieval uses AskUserQuestion
- [ ] Verify restart-resume flow: `sources.md` with `## Status: restart-pending` is detected on re-run and skill is re-invoked
- [ ] Verify graceful degradation when `awos` plugin is not installed


AlexanderMakarov (commented):
**General review point**

The direction is right — pulling project knowledge out of external wikis/tickets/chats is genuinely valuable. But I'd like us to rethink *where* this lives before merging, on three counts:

**1. `commands/sources/` is reference content, not command-prompt content.**
The three files are platform-by-platform setup guides (MCP → CLI → manual export) for ~47 platforms. That's exactly the shape of skill/registry material, and it duplicates what we already maintain in the `awos-recruitment` repo's `registry/` — both `registry/skills/*/SKILL.md` and `registry/mcp/*.yaml` (Atlassian, Slack, Google Drive, GitHub, GitLab… are already there as MCP configs). Bundling a parallel, hand-maintained copy of 47 platform integrations inside `commands/` means two sources of truth that will drift. I'd rather this content live in the registry (skills + MCP configs) and get *installed* on demand (the `/awos:hire` model - probably it makes sense to run it before `/awos:product` or think about other delivery option), not shipped inline as framework prompts.

**2. Setting up MCP/CLI tooling doesn't belong in `/awos:product`.**
`/awos:product` exists to author the product definition. It was never meant to detect data sources and walk the user through installing/auth'ing MCP servers. Folding tool-provisioning into it is responsibility creep — it bloats the command, couples product authoring to infra setup, and makes the "restart-resume after MCP install" flow leak into a command that shouldn't know about it.

**3. Source discovery + access is cross-cutting — make it shared, not per-command.**
Knowing what documentation/knowledge sources a project has (and how to query them) is valuable to far more than onboarding: the core commands, `/awos:ai-readiness-audit`, and likely future plugins all want it. The current PR wires the same accumulate-then-absorb logic into three commands independently; the audit and any plugin would each have to reinvent it.

Proposal: extract this into a single shared capability rather than threading it through individual commands.

- A new **core command — `/awos:investigate`** (alternatives: `/awos:sources`, `/awos:discover`, `/awos:connect`) whose job is exactly: detect available sources → help set up access (MCP/CLI, using the registry guides) → write a persistent catalogue to `context/sources.md` (what's configured, how to query each, scope/privacy notes). Run once, like the other setup-phase commands.
- Every consumer — `/awos:product`, `/awos:roadmap`, `/awos:architecture`, `/awos:ai-readiness-audit`, plugins — then just *reads* `context/sources.md` and runs purpose-scoped retrieval. No re-detection, no re-setup, no per-command MCP-restart flow.
- The 47-platform setup content moves to the `awos-recruitment` registry as skills/MCP configs and is installed when `/awos:investigate` needs it.

That keeps `/awos:product` focused, gives us one place to maintain platform integrations, and makes source-awareness reusable across the whole framework instead of just the onboarding path.

Happy to discuss the command name and the exact `context/sources.md` shape.

commands/product.md
```diff
@@ -90,13 +90,42 @@ First, check if the file `context/product/product-definition.md` exists.
...
+3.  **External documentation sources.** If `context/sources/sources.md` does not exist and `context/product/brownfield.md` was created in substep 2, use `AskUserQuestion` to ask: "Do you have external documentation (wikis, tickets, chats, email) you'd like to import into the project context?" with options **Yes** and **No**. If yes, invoke the configure-external-sources skill: `Skill(name="awos:configure-external-sources")`. The skill detects platforms, guides MCP/CLI setup, handles restart-resume, collects scope, and writes the result to `context/sources/sources.md`. If the skill triggers an editor restart, stop here.
+
+    If `context/sources/sources.md` already exists with `## Status: configured` (set up in a previous run), skip straight to retrieval.
+
+    If `context/sources/sources.md` exists with `## Status: none`, skip to substep 4 — user previously declined.
```
AlexanderMakarov: This is the one I'd block on. `SKILL.md` Step 5 tells the user to restart the editor and re-run `/awos:product`, promising setup "resumes automatically" — but the only branch that invokes the skill is the `does not exist` one above (line 93). After a restart, `sources.md` exists with `## Status: restart-pending`, so none of these three branches match and you fall straight through to **Retrieval** — against a manifest whose tools were never verified (Step 6) and whose scope is still blank (Step 7). And because the skill is `disable-model-invocation: true`, product is its only possible caller, so that setup never gets finished. Want a fourth branch here — "if the file exists with any status other than `configured`/`none`, re-invoke the skill to finish setup"? That's what makes the restart-resume the skill advertises actually work.

commands/product.md
```diff
@@ -90,13 +90,42 @@ First, check if the file `context/product/product-definition.md` exists.
...
 
-3.  Draft every section of the template up front so a complete definition exists before any further back-and-forth — use `<user_prompt>` (when non-empty) as the starting point, fold in any brownfield findings from step 2, and fill the rest from reasonable best-practice assumptions. Never block on a question before the write:
+3.  **External documentation sources.** If `context/sources/sources.md` does not exist and `context/product/brownfield.md` was created in substep 2, use `AskUserQuestion` to ask: "Do you have external documentation (wikis, tickets, chats, email) you'd like to import into the project context?" with options **Yes** and **No**. If yes, invoke the configure-external-sources skill: `Skill(name="awos:configure-external-sources")`. The skill detects platforms, guides MCP/CLI setup, handles restart-resume, collects scope, and writes the result to `context/sources/sources.md`. If the skill triggers an editor restart, stop here.
```
AlexanderMakarov: Heads-up on a new coupling: `/awos:product` installs via the copy table into `.awos/commands/`, but `configure-external-sources` ships only with the awos plugin, and the installer's `marketplace-configurator.js` only *registers* the marketplace — it doesn't install or enable the plugin (same reason `/awos:ai-readiness-audit` needs a separate plugin install). So on a stock AWOS install this `Skill(name="awos:configure-external-sources")` call has nothing to resolve to, and the Yes branch fails. Two ways out: make the awos plugin a guaranteed install-time dependency, or guard this call and skip to substep 4 when the skill isn't available. As written, the core flow assumes the optional plugin is present.

commands/product.md
```diff
@@ -90,13 +90,42 @@ First, check if the file `context/product/product-definition.md` exists.
...
+
+    If `context/sources/sources.md` already exists with `## Status: configured` (set up in a previous run), skip straight to retrieval.
+
+    If `context/sources/sources.md` exists with `## Status: none`, skip to substep 4 — user previously declined.
+
+    **Retrieval.** Read the source manifest and retrieve product-relevant content from each configured source:
+
+    ```text
+    Agent(subagent_type="Explore", description="Retrieve [platform] docs", prompt="
+    Use the [Tool name from sources.md] tools to retrieve content from [Scope from sources.md].
```
AlexanderMakarov: `sources.md` can carry `Access: manual` sources — `Tool: n/a`, with a `Path:` pointing at the exported file (SKILL.md Step 8). But this retrieval prompt only ever says "use the [Tool name] tools", so a manual source resolves to "use the n/a tools" and its exported content never gets read. Either branch here on access mode (for `manual`, read the file at `Path:` instead of calling a tool), or drop `manual` from the manifest so it doesn't promise a path nothing consumes. Same block is duplicated in `roadmap.md:90` and `architecture.md:89`, so whatever you pick lands in all three.

commands/architecture.md
```diff
@@ -81,11 +82,36 @@ Follow this logic precisely.
 
     c. Append the new findings to `context/product/brownfield.md` under a `## Technology` heading (for any you revise, record the revised version, not the original). The findings seed the section defaults below and are triaged with the user later, in **Step 3: Finalization** — after the architecture is saved — so exploration never blocks the write.
 
-3.  Draft every architectural area up front so a complete architecture exists before any back-and-forth — never blocking on a question before the write.
+3.  **External documentation context.** If `context/sources/sources.md` exists with `## Status: configured` (produced by the `configure-external-sources` skill during `/awos:product`), read the source manifest and retrieve architecture-relevant content from each configured source:
```
AlexanderMakarov: > (produced by the `configure-external-sources` skill during `/awos:product`)

This looks useless for the command and dangerous in case if LLM would decided to find out what are these 2 skills.

> read the source manifest

What is "source manifest" here?

commands/architecture.md
```diff
@@ -81,11 +82,36 @@ Follow this logic precisely.
...
+
+    ```text
+    Agent(subagent_type="Explore", description="Retrieve architecture docs", prompt="
+    Use the [Tool name from sources.md] tools to retrieve content from [Scope from sources.md].
```
AlexanderMakarov: > [Tool name from sources.md]

I don't see new tools are generated into "sources.md".

Also why you are using (custom) `[]` brackets here while below you have standard `{}` ones (in line 101)? Here and for `[Scope from sources.md]` part.

The same in "commands/roadmap.md" file too

plugins/awos/skills/configure-external-sources/references/communication.md
```diff
@@ -0,0 +1,313 @@
...
+
+Reference for setting up access to chat, messaging, and email platforms. Read the section for the user's selected platform and follow the setup instructions in order: MCP server first (recommended), CLI alternative if MCP is not viable, manual export as a last resort.
+
+**Privacy and data safety apply to every platform in this file.** Accessing message history may include sensitive or personal data. Confirm the user has authorization before proceeding. Focus retrieval on public/project channels and relevant threads — not private messages or DMs unless the user explicitly requests it.
+
+---
+
+## Slack
+
+**MCP Server (Recommended) — Official Remote**
```
AlexanderMakarov: >Official Remote

I don't think we need to hardcode on the specific tools in general. Because users may use different ones.

For example I am using not official https://github.com/provectus/slack-mcp-server because official server may send messages for me and signs them.

plugins/awos/skills/configure-external-sources/references/communication.md
```diff
@@ -0,0 +1,313 @@
...
+
+- Type: Remote (cloud-hosted by Slack)
+- URL: `https://mcp.slack.com/mcp`
+- Setup: Clone https://github.com/slackapi/slack-mcp-plugin and run `claude --plugin-dir ./` to register. Connects to Slack's hosted server.
```
AlexanderMakarov: Why we need to clone the MCP?

plugins/awos/skills/configure-external-sources/references/communication.md
```diff
@@ -0,0 +1,313 @@
+# Communication Platform Setup Guides
+
+Reference for setting up access to chat, messaging, and email platforms. Read the section for the user's selected platform and follow the setup instructions in order: MCP server first (recommended), CLI alternative if MCP is not viable, manual export as a last resort.
```
AlexanderMakarov: >MCP server first (recommended), CLI alternative if MCP is not viable, manual export as a last resort.

I am not agree with this order - user should be able to choose the tool. And it is quite strange for skill to install many tools on user's machine.

plugins/awos/skills/configure-external-sources/references/communication.md
```diff
@@ -0,0 +1,313 @@
+# Communication Platform Setup Guides
```
AlexanderMakarov, line 1: Why this file is used for? I can't find references on it except "tests/lint-prompts.test.js".

AlexanderMakarov (changes requested):
Thanks for the rework — moving the platform content into a skill and having roadmap/architecture just read a shared `context/sources/sources.md` addresses the bulk of what I raised last round; this is a much better shape. Not re-litigating placement.

While a number of architectural things appears again (see inline), some of them:

- **The restart-resume flow doesn't actually resume** (inline, `product.md`). The skill tells the user to restart the editor and re-run `/awos:product`, promising setup "resumes automatically" — but product only invokes the skill when `sources.md` doesn't exist. After a restart it exists as `## Status: restart-pending`, no branch matches, and product falls straight through to Retrieval against a manifest whose tools were never verified and whose scope is blank. The skill is `disable-model-invocation: true`, so product is its only caller — verification and scope collection just get skipped. This is the headline.
- **`product` now hard-depends on the awos plugin being installed** (inline). `/awos:product` installs via the copy table into `.awos/commands/`, but `configure-external-sources` ships only with the awos plugin, and the installer only *registers* the marketplace (`marketplace-configurator.js`) — it doesn't install the plugin. On a stock AWOS install the Yes branch calls a skill that isn't there. Worth deciding: make the plugin a hard install-time dependency, or degrade gracefully when the skill is absent.
-  **Manual-export sources are defined but never read** (inline). The manifest carries `Access: manual` + a `Path:` to exported content, but all three retrieval blocks only say "use the [Tool name] tools" — nothing consumes `Path`, so a manual source gets a prompt telling Explore to "use the n/a tools." Either wire `Path` into retrieval, or drop `manual`.
- **not linked and very opinonated files in references** - see inline.

plugins/awos/skills/configure-external-sources/references/communication.md
dustyo-O: > And it is quite strange for skill to install many tools on user's machine

what do you mean by "strange"? This skill installs tools, it's on purpose

plugins/awos/skills/configure-external-sources/references/communication.md
dustyo-O: https://github.com/provectus/awos/pull/138/changes#diff-0bbc6569c16f03d3c550421614d2559c8698a9fbbca842767fb30fbe1c98ed72R46

plugins/awos/skills/configure-external-sources/references/communication.md
dustyo-O: what is your suggestion? I would rather have official mcps as primary option (less malware risks, less risk for tools to obsolete, etc.), then doing a research from scratch, or recommend some tools which we personally prefer. 

If the user want to add some customization, he's free to amend any tool which were installed during this flow

commands/architecture.md
```diff
@@ -121,4 +147,4 @@ Give the user a quick read on whether the stack already has specialist agents 
 
 ### Step 5: Brownfield Cleanup
 
-If `context/product/brownfield.md` exists, delete it. By this point all brownfield knowledge has been absorbed into `product-definition.md`, `roadmap.md`, and `architecture.md` — the brownfield file is no longer needed.
+If `context/product/brownfield.md` exists, delete it. If `context/sources/` exists, delete it. By this point all external knowledge has been absorbed into `product-definition.md`, `roadmap.md`, and `architecture.md`.
```
AlexanderMakarov: Now that manual exports land in `context/sources/` (SKILL.md Step 4 suggests e.g. `context/sources/confluence-export.md`), this cleanup deletes files the user hand-exported — real effort, gone as a side effect of a step they don't associate with those files. Deleting `sources.md` also drops the durable record of what the user configured or declined. I'd scope this to what the command actually absorbed — or keep the directory and delete nothing here. If wiping it is deliberate (sources as one-shot onboarding input), a sentence saying so in the step would settle it.

plugins/awos/skills/configure-external-sources/references/communication.md
```diff
@@ -0,0 +1,313 @@
...
+
+---
+
+## Element / Matrix
+
+**MCP Server — Community**
+
+- Package: `matrix-mcp-server` — https://github.com/mjknowles/matrix-mcp-server
+- Install: `claude mcp add --transport http matrix-server http://localhost:3000/mcp -H "matrix_user_id: @user:matrix.example.com" -H "matrix_homeserver_url: https://matrix.example.com" -H "matrix_access_token: ${MATRIX_ACCESS_TOKEN}" -H "Authorization: Bearer ${MATRIX_MCP_TOKEN}"`
+- Auth: Generate a Matrix access token in Element → Settings → Help & About → Access Token (or via `curl -XPOST -d '{"type":"m.login.password","user":"...","password":"..."}' https://matrix.example.com/_matrix/client/r0/login`). The MCP bearer token is set when starting the server.
```
AlexanderMakarov: This registers a client for `http://localhost:3000/mcp`, but the run-the-server step disappeared in the rewrite. The [matrix-mcp-server README](https://github.com/mjknowles/matrix-mcp-server) requires `git clone` → `npm install` → `npm run build` → `.env` → `npm start` before anything listens on port 3000 — and `claude mcp add` only registers config; the connection is attempted at session start and [marked failed after retries](https://code.claude.com/docs/en/mcp#automatic-reconnection), so a reader lands with `claude mcp add` succeeding and every later call failing. The Auth line still says the bearer token "is set when starting the server" — a step that no longer exists in the doc. Worth one line on starting the server (or a pointer to its README) before the `claude mcp add`. (Ironically this is the one entry where the clone-and-build flow it replaced is what the upstream README actually prescribes.)

tests/lint-prompts.test.js
```diff
@@ -919,6 +919,334 @@ test('product.md creates brownfield.md even when all findings are rejected', ()
...
+  // The "No" branch must mention writing Status: none
+  const externalDocsBlock = body
+    .split(/external documentation sources/i)
+    .slice(1)
+    .join('');
+  assert.ok(
+    /if no.*status: none|default to.*no/i.test(externalDocsBlock),
+    'commands/product.md must write ## Status: none when user declines external documentation'
+  );
+});
```
AlexanderMakarov: The regex here doesn't pin what the test name promises: `/if no.*status: none|default to.*no/i` passes on "default to **No**" alone — I mutation-tested it; deleting the `## Status: none` write from product.md's declined branch stays green. It's the same alternation trap CodeRabbit caught last round in the persistence test (https://github.com/provectus/awos/pull/138#discussion_r3536304119). Two separate asserts (`/if no.*## Status: none/i` and `/default to \*\*No\*\*/`) pin both behaviors. A few siblings in this block have the same looseness (each mutation-verified the same way):

1. `degrades gracefully` (line 1074) pins the *mention* ("skill is not available") but not the fallback action — replacing both fallback writes with "stop" stays green. `.*## Status: none` in the same regex would pin the degradation.
2. `collects file path` (line 1101) leaves the real cross-file promise unpinned: the test reads only SKILL.md, so a command reverting to Tool-only dispatch stays green. SKILL.md writes `- Path:` and all three commands dereference it — asserting `- Path:` in SKILL.md plus `Path:` inside each command's external-doc block is the symmetry this linter exists for.
3. `fallback for failed verification` (line 1120): `split(/tool verification/i)` splits at Step 1's mention of it (char 1114 of the file vs. Step 6 at 5228), so the "section" is nearly the whole file — split on `/## Step 6/` instead. And the `|` again passes with only one of the two options present (deleting **Remove this source** stays green).

One gap while you're in here: product.md's side of the restart handshake (the `restart-pending` → re-invoke branch at product.md:98) has no contract — the only `restart-pending` test reads SKILL.md — so dropping that branch strands interrupted setups with tests green.

commands/product.md
AlexanderMakarov: The status dispatch fixes this — thanks. One hardening idea while you're in there: gate the **Retrieval** paragraph on `## Status: configured` explicitly, the way roadmap.md and architecture.md do. Today it's only reachable through the right bullet, but if a model treats the skill's soft "restart-pending" return as anything other than "stop here", it falls through to retrieval with unverified tools and blank scopes.

commands/product.md
AlexanderMakarov: The unavailable-skill fallback covers the fresh-run branch well. The resume branch (line 98) is riskier: writing `## Status: none` there overwrites a file that already records selected sources and installed tools — the user's setup work is discarded and the file now says they declined, so even after installing the plugin nothing resumes. I'd leave sources.md untouched on that path, tell the user the plugin is needed to finish setup, and skip retrieval.

commands/product.md
AlexanderMakarov: `Path:` now flows through the manifest and the commands' prose — good. The Explore template itself is still the tool-only variant though: "Use the {tool name} tools to retrieve content from {scope}" renders as "Use the n/a tools…" for a manual source. Since the template is the executable part of the step, I'd either branch it ("If Access is manual, read the file at {path} and extract…") or show a second template for manual sources. Applies to all three commands.

commands/architecture.md
AlexanderMakarov: Addressed by the rewording — thanks.

commands/architecture.md
AlexanderMakarov: Standardized now — thanks.

plugins/awos/skills/configure-external-sources/references/communication.md
AlexanderMakarov: The Step 4 rewrite — discover what's already in the environment first, present options, user picks — settles this for me.

plugins/awos/skills/configure-external-sources/references/communication.md
AlexanderMakarov: Resolved — the clone instructions are gone and SKILL.md now says not to instruct cloning.

plugins/awos/skills/configure-external-sources/references/communication.md
AlexanderMakarov: The knowledge-base reframe plus the AskUserQuestion choice addresses this — the skill no longer prescribes an install order.

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+
+If the user selected any chat or email source, use `AskUserQuestion` to confirm: "Accessing message history may include sensitive or personal data. Do you have authorization to access this data for project documentation?" with options **Yes, I have authorization** and **Skip communication sources**. If skipped, remove all chat and email sources from the list. If the filtered list is now empty, write `context/sources/sources.md` with `## Status: none` and stop.
+
+## Step 4 — Tool Setup
+
+For each selected platform:
+
+1. **Discover available tools.** Check what tools are already available in the environment for this platform (e.g., existing MCP servers, installed CLIs like `gh`). Then read `references/{documentation,tickets,communication}.md` (relative to this SKILL.md) for known MCP servers and CLI tools for the platform.
+
+2. **Present options.** Use `AskUserQuestion` to present the discovered tools. List tools already available in the environment first, then known MCP servers and CLI tools from the reference file — official/vendor-hosted tools as primary options unless they appear outdated or deprecated. Always include **Manual export** as the final option.
```
AlexanderMakarov: Discovery-first is the right shape. 

One posture question on the branch where discovery finds **nothing**: the skill's default next move is presenting install options, which puts AWOS in the position of recommending third-party packages it doesn't vet (several reference entries are community npm servers that receive OAuth tokens or full message history) — and the install outlives its purpose: retrieval runs three times and `/awos:architecture` then deletes `context/sources/`, but the registered MCP servers and their credentials stay in the user's config with nothing offering cleanup.

I'd flip the default for that branch: when no tool is already available, ask for manual export and make installation the explicit opt-in - "I can help set up X if you want live retrieval". Also it would be helpful to suggest user to describe installation process in the project README.md for other team members.

src/migrations/003-remove-legacy-sources.json
```diff
@@ -0,0 +1,13 @@
+{
+  "version": 3,
+  "name": "Remove legacy platform guides from .awos/commands/sources/",
+  "preconditions": {
+    "require_any": [".awos/commands/sources"]
```
AlexanderMakarov: .awos/commands/sources/ was only ever written by earlier revisions of this branch — `commands/sources/` never existed on main and the rewrite squashed it out of this branch's history too, so no released installer ever created the dir this deletes. The only projects it can touch are scratch/road-test installs from the pre-rewrite branch. It's harmless (the `require_any` gate makes it a skip for everyone else), but it permanently occupies version 3 in the migration chain for a state that never shipped — if none of the road-test projects matter, I'd drop it and keep the slot. Related: the PR description still says the guides ship as `commands/sources/` (47 platforms) — worth updating to the plugin-skill layout so reviewers aren't chasing a directory that no longer exists.

commands/product.md
```diff
@@ -90,13 +90,44 @@ First, check if the file `context/product/product-definition.md` exists.
 
     b. Create `context/product/brownfield.md` with a `## Product` heading and record the findings under it. If the exploration surfaced nothing, still create the file with an empty `## Product` section; downstream commands (`/awos:roadmap`, `/awos:architecture`) key on the file's existence to run their own explorations. The findings are triaged with the user later, in **Step 4** — after the definition is saved — so exploration never blocks the write.
 
-3.  Draft every section of the template up front so a complete definition exists before any further back-and-forth — use `<user_prompt>` (when non-empty) as the starting point, fold in any brownfield findings from step 2, and fill the rest from reasonable best-practice assumptions. Never block on a question before the write:
+3.  **External documentation sources.** Check `context/sources/sources.md`:
```
AlexanderMakarov, line 93: >Check `context/sources/sources.md`

`/awos:product` is called first in the whole AWOS flow. And there is nothing above what writes into `context/sources/`. So the whole this step #3 is about "re-run" case.

[nit] I guess `If it does not exist and no brownfield.md was created, skip to substep 4.` better to put first to simplify flow (i.e. "exit earlier" strategy).

commands/product.md
```diff
@@ -90,13 +90,44 @@ First, check if the file `context/product/product-definition.md` exists.
...
+
+    - If it does not exist and `context/product/brownfield.md` was created in substep 2, use `AskUserQuestion` to ask: "Do you have external documentation (wikis, tickets, chats, email) you'd like to import into the project context?" with options **Yes** and **No**. If the question goes unanswered, default to **No**. If no, write `context/sources/sources.md` with `## Status: none` and continue to substep 4. If yes, try to invoke the skill: `Skill(name="awos:configure-external-sources")`. If the skill is not available (the `awos` plugin is not installed), inform the user they can install it from the marketplace to enable external documentation import, write `context/sources/sources.md` with `## Status: none`, and continue to substep 4. If the skill triggers an editor restart, stop here.
```
AlexanderMakarov: 2 places in this line, note that 1st is later, 2nd is earlier - it is important for 2nd:

>If the skill is not available (the `awos` plugin is not installed), inform the user hey can install it from the marketplace to enable external documentation import, write `context/sources/sources.md` with `## Status: none`, and continue to substep 4.

I believe Claude Code can't install plugin itself and proceed in the same session. So `no plugin` == restart session.

>If no, write `context/sources/sources.md` with `## Status: none`

It means that `context/sources/` is created in any case, even if plugin is not installed. And if plugin is not installed and user is not wishing to then LLM have to do a number of decisions just to eventually create useless files nobody will use. Let's change the order of checks to follow "exit earlier" strategy and don't spoil `context/` folder.

plugins/awos/skills/configure-external-sources/references/communication.md
```diff
@@ -0,0 +1,313 @@
...
+
+---
+
+## Slack
+
+**MCP Server — Official Remote**
+
+- Type: Remote (cloud-hosted by Slack)
+- URL: `https://mcp.slack.com/mcp`
+- Install: `/plugin install slack@claude-plugins-official`
```
AlexanderMakarov, line 15: 1. Claude Code can't install plugin for itself, especially in subagents:
```
Q: Can you install plugin inside the session if I ask you to?
A:
- Can install via Bash: `claude plugin install <name>@<marketplace>` (and `claude plugin marketplace add <repo>` if marketplace not configured). Non-interactive, works mid-session.
- Interactive `/plugin` manager is user-only; I can't drive it.
- Catch: newly installed plugin isn't active in current session until **you** run `/reload-plugins` (or restart). I can't invoke built-in slash commands, so activation step is yours.

So: ask, I install, you run `/reload-plugins`, then plugin's skills/agents/hooks are live.
```
So even after plugin is installed users have to run `/reload-plugins` themselves.

2. User may have already and/or want to use different Slack MCP. For example I prefer https://github.com/provectus/slack-mcp-server/
3. This MCP slack will require authorization almost always, i.e. run of `/mcp` command, navigation inside, doing something for authentication and next return to exploration part.

So I'm curious - how it happens in reality? Have you made experiments with removing MCP tools and installing them from scratch? Could you provide JSONL cc session record where it happened?

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+- **Documentation:** Confluence, Notion, Google Docs/Drive, SharePoint, Other
+- **Tickets:** Jira, Linear, GitHub Issues, GitLab Issues, Other
+- **Chat:** Slack, Microsoft Teams, Google Chat, Other
+- **Email:** Gmail, Outlook, Other
+
+If the user selects "Other" for any category, read the full reference file at `references/{documentation,tickets,communication}.md` (relative to this SKILL.md) and present the additional platforms listed there as follow-up options.
+
+## Step 3 — Privacy Gate
+
+If the user selected any chat or email source, use `AskUserQuestion` to confirm: "Accessing message history may include sensitive or personal data. Do you have authorization to access this data for project documentation?" with options **Yes, I have authorization** and **Skip communication sources**. If skipped, remove all chat and email sources from the list. If the filtered list is now empty, write `context/sources/sources.md` with `## Status: none` and stop.
```
AlexanderMakarov: >any chat or email source

>"Accessing message history may include sensitive or personal data. ..."

Actually the same applies to documentation and tickets - I've seen multiple times how PII data is populated into wiki and Jira tickets. Especially it is applicable for support-raised client bug ticket, "duties schedule" and "whom to call on urgency/personal issue" wiki pages.
Therefore "Privacy Gate" is not about "communication" sources only, but about any other source of data.

> Do you have authorization to access this data for project documentation?

This is strange question, asking for access but not privacy. In any somehow sane organization access to data is not provided to employees who don't have authorization. I guess most suitable question here would be "do you wish to provide LLM (LLM provider API) behind current agentic tool access to data you have in these sources?". Because exactly this is current users action they are about to do and company may have a policy to "don't feed LLM with PII data". BTW agentic tool could leak as well, but let's take it out of scope for now.

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Step 4 — Tool Setup
+
+For each selected platform:
+
+1. **Discover available tools.** Check what tools are already available in the environment for this platform (e.g., existing MCP servers, installed CLIs like `gh`). Then read `references/{documentation,tickets,communication}.md` (relative to this SKILL.md) for known MCP servers and CLI tools for the platform.
+
+2. **Present options.** Use `AskUserQuestion` to present the discovered tools. List tools already available in the environment first, then known MCP servers and CLI tools from the reference file — official/vendor-hosted tools as primary options unless they appear outdated or deprecated. Always include **Manual export** as the final option.
+
+3. **Guide installation.** For the user's chosen tool, use the setup details from the reference file to guide installation and authentication. For MCP servers, use `npx`/`bunx` or `claude mcp add` — do not instruct the user to clone repositories. For manual export, guide the user to export content and save it to a file in the project directory (e.g., `context/sources/confluence-export.md`). Use `AskUserQuestion` to confirm the file path once the export is saved.
```
AlexanderMakarov: >For MCP servers, use `npx`/`bunx` or `claude mcp add`

See comment above about Claude Code's ability to installed plugins/MCP's inside its own.

AlexanderMakarov: >guide the user to export content and save it to a file in the project directory (e.g., `context/sources/confluence-export.md`)

Here I expect that a model would say in which file or directory to save data for it.

>Use `AskUserQuestion` to confirm the file path once the export is saved.

But here it appears that users may do different thing. Why so few trust into user here? Also it adds extra frictions for sane users.

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+
+4. **Track the result.** Record the access method chosen (mcp, cli, or manual) and the tool name for each source. For manual access, also record the file path from substep 3.
```
AlexanderMakarov: >also record the file path from substep 3.

I guess we may end up having multiple files if user made manual experts from multiple sources.

BTW the whole SKILL.md expects only one source. Better to support multiple.

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Step 5 — Restart Check
+
+If any MCP servers were added during Step 4, write `context/sources/sources.md` with `## Status: restart-pending` and a `## Source:` section for each configured source (category, platform, access method, tool name — scope left blank for now). Then tell the user:
+
+> MCP servers have been configured. Resume this conversation (or restart your editor and run `/awos:product` again). Source setup will resume automatically.
```
AlexanderMakarov: >or restart your editor and run `/awos:product` again

1. Why `/reload-plugins` won't work? We are assuming only Claude Code here.
2. In Claude Code `/mcp` command (user-only, interactive) can reconnect/re-authenticate already-configured servers mid-session, but won't pick up new ones.
3. In Claude Code we don't need to start a new session - we may just exit current process (twice Ctrl+C or `/exit`) and next open the same session again via last line printed after exit - via something like `claude --resume e2567e82-ff11-4f0e-aa2c-3759e26c3d` - MCP servers, plugins, settings, CLAUDE.md are all loaded fresh at process start.

>editor

I won't call Claude Code or Claude Desktop "editors" - they are agentic tools.

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+
+If no MCP restart is needed (tools were already available, CLI was chosen, or manual export selected), continue to Step 6.
+
+## Step 6 — Tool Verification
+
+Update the status in `context/sources/sources.md` to `verifying`. If the file already exists (from a restart-pending state), update in place; otherwise write a new file with the current source list.
+
+For each configured MCP or CLI tool, attempt a simple read operation (e.g., search for a known term, list projects, or list channels) to confirm the tool responds. If verification fails, read the matching section from the reference file and help troubleshoot authentication or configuration. If the tool still fails after troubleshooting, use `AskUserQuestion` to ask: "Could not verify {tool name} for {platform}. What would you like to do?" with options **Retry**, **Switch to manual export**, and **Remove this source**. On retry, attempt verification again. On switch, update the source's access method to manual and collect the export file path. On remove, drop the source from the list entirely. If the list becomes empty after removals, write `## Status: none` and stop.
```
AlexanderMakarov: >If verification fails, read the matching section from the reference file and help troubleshoot authentication or configuration.

>If the tool still fails after troubleshooting ... with options **Retry**,

It wasn't able to connect and user already helped with troubleshooting - how "Retry" may help?

AlexanderMakarov: >On remove, drop the source from the list entirely.

Which list? I see `list projects, or list channels` above and looks like there is a list in `context/sources/sources.md` file.

AlexanderMakarov (changes requested):
Everything from the last round is addressed — thanks. The core concern this round: the setup flow assumes Claude Code can install plugins/MCP servers and use them in the same session, which it can't — the skill-unavailable, restart-resume, and MCP setup steps all need grounding in the real session lifecycle, and I'd want evidence the flow was road-tested end-to-end before trusting it. Beyond that, manual sources still never reach the retrieval template, and the rest are smaller design and test-contract fixes — all inline. Close once the session-lifecycle assumptions are straightened out.

plugins/awos/skills/configure-external-sources/SKILL.md
dustyo-O: Addressed — Step 4.2 now defaults to manual when no tools are already available (tool installation is opt-in, not the default path). Step 5 gives concrete `claude --resume` instructions for the MCP pickup flow.

dustyo-O: Moot — manual sources no longer involve file export. The user pastes content directly during retrieval via `AskUserQuestion`. No file path to confirm.

plugins/awos/skills/configure-external-sources/SKILL.md
dustyo-O: SKILL.md does support multiple sources — Step 2 uses `multiSelect: true`, Step 4 loops with "For each selected platform", and Step 8 writes "One `## Source:` section per configured source." The per-source loop was there from the start.

plugins/awos/skills/configure-external-sources/SKILL.md
dustyo-O: All four points addressed: Step 5 now says `claude --resume` (not "restart your editor"), mentions `/mcp` for servers needing interactive auth, and doesn't use the word "editor." `/reload-plugins` is for plugin reloads specifically — MCP servers need a session restart to be picked up.

plugins/awos/skills/configure-external-sources/SKILL.md
dustyo-O: Fair point — dropped **Retry** from the options. After troubleshooting already failed, retry is noise. Now offers only **Switch to manual** and **Remove this source**.

dustyo-O: Clarified — now says "drop the source from the manifest — do not include it in `sources.md`" instead of the ambiguous "drop from the list."

commands/product.md
dustyo-O: Done — Retrieval paragraph now explicitly gates on `## Status: configured`, matching roadmap.md and architecture.md.

commands/product.md
dustyo-O: Fixed — the resume branch now leaves `sources.md` untouched and skips to substep 4, preserving the user's existing configuration.

commands/product.md
dustyo-O: Reworked — manual sources no longer go through file export. Commands now use `AskUserQuestion` to let the user paste content directly. The `Path:` field is removed from the manifest.

commands/architecture.md
dustyo-O: Fixed — architecture.md now checks whether all source information was absorbed before deciding to delete. If useful content remains (source URLs, user-pasted content), `sources.md` is kept and cross-referenced from `product-definition.md`.

plugins/awos/skills/configure-external-sources/references/communication.md
dustyo-O: Fixed — Matrix entry now documents the full clone → build → start flow and clarifies the MCP bearer token comes from server startup output.

tests/lint-prompts.test.js
dustyo-O: All 5 fixed: (1) split into two asserts pinning the write and the default separately, (2) added second assert for the Status: none fallback action, (3) renamed to cross-file symmetry test covering SKILL.md + all three commands, (4) splits on `## Step 6` and uses two separate asserts, (5) new test `product.md re-invokes skill for intermediate source states`. Also updated the architecture deletion test for the new conditional-cleanup behavior.

tests/lint-prompts.test.js
```diff
@@ -1934,6 +1960,413 @@ test('brownfield commands launch Explore agents', () => {
...
+    assert.ok(
+      /sources\.md.*exists.*configured|sources\.md.*configured/i.test(
+        extDocBlock
+      ),
+      `commands/${cmd} must guard documentation retrieval on context/sources/sources.md existence with configured status`
+    );
+  }
+});
+
+test('context/<path> references in prompts are internally consistent', () => {
```
AlexanderMakarov: This test builds `refs` and then the function ends — no `assert.*`. It's permanently green: `node:test` counts a test as passing as long as its function doesn't throw, so assertions aren't required ([Node test runner docs](https://nodejs.org/api/test.html)). A test of this exact name already exists further down (the one on `main`), and that one does assert, on three canonical paths. So this copy is a dead duplicate that shadows the real test's name — and the assertion that would earn the "consistency" title for this PR (that `context/sources/sources.md` is referenced consistently across the prompts) is exactly what's missing. Either drop this copy, or extend the existing test to include `context/sources/sources.md` rather than adding a second same-named no-op.

CLAUDE.md
```diff
@@ -91,7 +91,7 @@ Each file in `claude/commands/{name}.md` is a tiny wrapper that points at `.awos
...
+**Brownfield projects** get automatic codebase awareness and external documentation gathering. `/awos:product` detects existing source code (during Creation Mode only) and creates `context/product/brownfield.md` with triaged findings about purpose, audience, and features. It also detects external documentation sources (wikis, ticket systems, chats, email) by invoking the `configure-external-sources` plugin skill, which guides MCP/CLI tool setup and writes structured source configuration to `context/sources/sources.md`. If MCP setup requires an editor restart, the skill writes a partial config and resumes automatically on re-run. `/awos:roadmap` and `/awos:architecture` read the brownfield file and the sources config, running their own focused explorations and retrievals (capabilities and tech stack respectively), passing prior findings to avoid duplicate questions. All findings are triaged with the user (accept / correct / reject). `/awos:architecture` removes both `brownfield.md` and `context/sources/` after all knowledge is absorbed into the three core documents.
```
AlexanderMakarov: This says architecture "removes both `brownfield.md` and `context/sources/` after all knowledge is absorbed", but `architecture.md` Step 5 only deletes `context/sources/` when everything was absorbed — otherwise it keeps `sources.md` and cross-references it from `product-definition.md`. Worth rewording so the doc reflects the retain-and-reference branch, e.g. "removes `brownfield.md`, and removes `context/sources/` once its contents are fully absorbed — otherwise keeps it and references it from `product-definition.md`."

tests/lint-prompts.test.js
```diff
@@ -1934,6 +1960,413 @@ test('brownfield commands launch Explore agents', () => {
...
+    'commands/product.md must handle the case where the awos plugin (and its skill) is not installed'
+  );
+  // Pin the fallback action — the fresh-run branch must write Status: none
+  // when the skill is unavailable, not just mention the absence.
+  const externalDocsBlock = body
+    .split(/external documentation sources/i)
+    .slice(1)
+    .join('');
+  assert.ok(
+    /skill is not available[^]*?Status: none/i.test(externalDocsBlock),
```
AlexanderMakarov: `/skill is not available[^]*?Status: none/i` anchors on the first "skill is not available" in `product.md` — the intermediate-status (restart-pending/verifying/verified) branch, whose contract is to leave `sources.md` untouched (it must not write `Status: none`). The lazy match then runs on to the `Status: none` in the second, fresh-run branch, so the test passes but pins the wrong branch. Split on the fresh-run clause (the "install it from the marketplace … `## Status: none`" sentence) so it asserts what the test name claims.

plugins/awos/skills/configure-external-sources/SKILL.md
```diff
@@ -0,0 +1,111 @@
+---
+name: configure-external-sources
+description: >-
+  Detect and configure external documentation sources (wikis, ticket systems,
+  chats, email). Guides MCP/CLI tool setup, handles editor restart-resume,
+  and writes structured source configuration to context/sources/sources.md.
+  Called by onboarding commands when a brownfield project has external docs.
+disable-model-invocation: true
```
AlexanderMakarov: Road-tested end-to-end this round, and this is the blocker: `disable-model-invocation: true` means only a human can invoke the skill — Claude can't ([skills docs](https://code.claude.com/docs/en/skills): "Only you can invoke the skill … Claude can't run the skill automatically"). But `product.md` substep 3 (lines 98–99) drives the whole feature through `Skill(name="awos:configure-external-sources")`, which is Claude invoking it. In a live run every such call failed with `Skill awos:configure-external-sources cannot be used with Skill tool due to disable-model-invocation`, so the skill body never executed — fresh configuration and restart-resume both dead, `sources.md` never reaches `Status: configured`. Drop the flag: this skill is meant to be driven by `product.md`, so it must be model-invocable.

Reproduce: run `/awos:product` on a brownfield project and answer **Yes** to the external-docs question, or seed `context/sources/sources.md` with `## Status: restart-pending` and re-run. 2 of 7 e2e scenarios (`product-configures-sources-fresh`, `product-resumes-sources-restart`) are pinned to this failure.

commands/product.md
```diff
@@ -90,13 +90,46 @@ First, check if the file `context/product/product-definition.md` exists.
...
+    - If it exists with `## Status: restart-pending`, `## Status: verifying`, or `## Status: verified`, re-invoke the skill: `Skill(name="awos:configure-external-sources")`. If the skill is not available, inform the user that the `awos` plugin is needed to finish source setup and skip to substep 4 — leave `sources.md` untouched so the user's existing configuration is preserved. If the skill triggers an editor restart, stop here.
+    - If it does not exist and `context/product/brownfield.md` was created in substep 2, use `AskUserQuestion` to ask: "Do you have external documentation (wikis, tickets, chats, email) you'd like to import into the project context?" with options **Yes** and **No**. If the question goes unanswered, default to **No**. If no, skip to substep 4. If yes, try to invoke the skill: `Skill(name="awos:configure-external-sources")`. If the skill is not available (the `awos` plugin is not installed), inform the user they can install it from the marketplace to enable external documentation import, write `context/sources/sources.md` with `## Status: none`, and continue to substep 4. If the skill triggers an editor restart, stop here.
```
AlexanderMakarov: Downstream of the `SKILL.md` `disable-model-invocation` issue, but a rough edge on its own: when the `Skill(...)` call is rejected for any reason, this branch assumes the plugin isn't installed — it writes `## Status: none` and tells the user "the awos plugin is not installed." In the road-test the fresh-config scenario hit exactly this with the plugin **present**: the model read the `disable-model-invocation` rejection as "plugin absent" and silently degraded with a misleading message, so a configured project collapsed to `none`. Even once the flag is fixed, consider distinguishing a genuine plugin-absent case from other `Skill` failures. The same wall applies to the restart-pending branch on line 98.

AlexanderMakarov (commented):
[I road-tested this end-to-end this round — 7 scenarios run live with `claude -p`](https://github.com/provectus/awos/pull/138) against the PR branch (`feat/external-documentation-gathering`), driving `/awos:product`, `/awos:roadmap`, `/awos:architecture`, and `/awos:flow`. **5 of 7 green, 2 red — and both reds trace to one real defect: the `configure-external-sources` skill can't be invoked at all.**

`SKILL.md` sets `disable-model-invocation: true`, which per the [skills docs](https://code.claude.com/docs/en/skills) means "only you can invoke the skill … Claude can't run the skill automatically." But `product.md`'s only mechanism to run it is `Skill(name="awos:configure-external-sources")` — Claude invoking it. Live, every such call errored with `Skill awos:configure-external-sources cannot be used with Skill tool due to disable-model-invocation`, so the skill body never ran. That kills both `product.md` paths: fresh source configuration, and restart-resume. Inline on `SKILL.md:8`.

A second, latent edge falls out of it: when that `Skill(...)` call is rejected, `product.md` can't tell "skill present but non-invocable" from "plugin absent", so it takes the graceful-degradation branch — writes `## Status: none` and tells the user *the awos plugin isn't installed* when it is. Inline on `product.md:99`.

Worth stating what did hold up, because it narrows the fix: `architecture.md` and `flow.md` drive the MCP connector **directly** via Explore/tool calls, and both passed live (real Atlassian-stub retrieval, correct cleanup, transport inheritance). The manual-paste path (`roadmap.md`), the import-decline path, and graceful degradation when the plugin is genuinely absent all passed too. The defect is isolated to the one seam where `product.md` hands off to the **skill**.

Reproduce the blocker: run `/awos:product` on a brownfield project and answer **Yes** to the external-docs question (fresh path), or seed `context/sources/sources.md` with `## Status: restart-pending` and re-run (resume path). Both hit the `disable-model-invocation` wall.

The smaller items are inline: a no-op consistency test, a CLAUDE.md cleanup-doc overstatement, and a degrade-test regex anchored on the wrong branch. A couple of other new tests also slice on `split(/external documentation/i)`, which in `product.md` matches the user-facing question text too, so those sub-checks are broader than they look — not blocking, worth a pass.

plugins/awos/skills/configure-external-sources/SKILL.md
dustyo-O: Fixed — removed `disable-model-invocation: true` from SKILL.md frontmatter. The skill is designed to be invoked by product.md via `Skill()`, so it must be model-invocable.

commands/product.md
dustyo-O: Fixed — product.md now distinguishes `skill not found (plugin absent)` from other Skill call failures. The intermediate-status branch reports the actual error message instead of assuming absent; the fresh-run branch does the same before writing `Status: none`.

tests/lint-prompts.test.js
dustyo-O: Fixed — removed the no-op duplicate. The real test (from main, with assertions on canonical paths) stays.

CLAUDE.md
dustyo-O: Fixed — reworded to: "`/awos:architecture` removes `brownfield.md`, and removes `context/sources/` once its contents are fully absorbed — otherwise keeps `sources.md` and references it from `product-definition.md`."

tests/lint-prompts.test.js
dustyo-O: Fixed — the test now splits on `install it from the marketplace` (unique to the fresh-run clause) before asserting `Status: none`, so it pins the correct branch. Also tightened the two `split(/external documentation/i)` matches in the manual-source and brownfield-findings tests to `split(/external documentation (sources|context)/i)` so they don't match the user-facing question text.

plugins/awos/skills/configure-external-sources/SKILL.md
AlexanderMakarov: Addressed — Step 4 now defaults to Manual when nothing is discovered and makes tool install an explicit opt-in, and Step 8 notes that installed MCP servers/credentials persist and suggests documenting setup in the README. Closing.

src/migrations/003-remove-legacy-sources.json
AlexanderMakarov: Resolved — migration 003 was dropped from the branch; the diff no longer touches `src/migrations`. Closing.

commands/product.md
AlexanderMakarov: Satisfied now — `If it does not exist and no brownfield.md was created, skip to substep 4` is already the first check in substep 3, which is the early-exit I was after. Closing.

commands/product.md
AlexanderMakarov: This block was rewritten in the later fixups, so the specific line ordering here is moot now. Closing — happy to reopen if I've missed the underlying concern.

plugins/awos/skills/configure-external-sources/SKILL.md
AlexanderMakarov: Addressed — the Step 3 privacy gate now covers tickets and wikis too (PII in tickets, internal wiki discussions), not just chat/email, and is framed around sending data to the LLM provider. Closing.

plugins/awos/skills/configure-external-sources/references/communication.md
AlexanderMakarov: You're right — this is covered by SKILL.md. Step 4.2 always offers **I have a different tool** so a user can bring their own/custom Slack MCP (e.g. provectus/slack-mcp-server). Step 5's restart-resume handshake handles activation of a freshly-added MCP (exit + `claude --resume`, then re-run `/awos:product`), and explicitly says to run `/mcp` after resuming for servers that need interactive auth. So the reload / interactive-auth / BYO concerns are all handled in the flow. Resolving.

AlexanderMakarov (approved):

PR #140, opened 2026-06-25
feat(spec): raise completeness and quality of /awos:spec output

FlySpot:
## What
  
  Applies targeted improvements to `/awos:spec` that raise the completeness and testability of the functional spec it produces: capturing user-visible boundary and error behavior, catching vague or unmeasurable wording, and confirming before the file is saved that every requirement carries an acceptance criterion. Open questions are preserved as `[NEEDS CLARIFICATION: …]` markers in the saved file (never converted to assumptions), consistent with main's write-first contract.

  ### Changes to `commands/spec.md`

  **Step 3 — Interactive Drafting and Clarification**
  - New rule under **Functional Requirements**: *"Where applicable, capture the boundary and error behavior the user sees: what error message appears, what limits exist (file size, format, count), and what happens when the action fails."*
  - New rule under **Acceptance Criteria**: *"For requirements that capture boundary or error behavior, include at least one acceptance criterion covering the failure path (e.g., 'When the user uploads a file larger than 5MB, then they see: File too large. Maximum size is 5MB.')."*

  **Step 4 — Self-Review (Language and Ambiguity Check)**
  - Renamed from *"Self-Review (Language Check)"* to *"Self-Review (Language and Ambiguity Check)"*.
  - Added a second pass that re-reads **Functional Requirements** — statements and acceptance criteria — for vague or unmeasurable wording (weasel words like "fast", "user-friendly", "as appropriate") and requires each to be made concrete in user-perceivable terms or converted to a `[NEEDS CLARIFICATION: …]` marker.
  - Added a **Definition of Done** self-check — a review, not an approval gate; the file is still written — that confirms two things before saving: no vague wording remains in requirements or acceptance criteria, and every requirement carries at least one When/Then acceptance criterion.

  **Step 6 — Final Review and Recommend Next Step**
  - The post-save marker resolution now uses `AskUserQuestion`, offering the assumption you would otherwise make as the recommended first option (with a free-text option for open-ended markers), then folds each answer back into the relevant requirement and its acceptance criteria and re-saves. If no answer comes (an unattended `claude -p` run), the markers are left in place for the user — or `/awos:tech` — to resolve later.

  Step 5 (File Generation, write-first) is unchanged by this PR; its ordering — the file is saved before the final review — is the pre-existing contract this change relies on so that an unattended run always produces a complete deliverable.

  ## Why

  Before this change:
  - Error and boundary conditions appeared only in examples, not as a rule — requirements routinely omitted failure paths.
  - The self-review caught technical-language leakage but not vague, untestable wording like "user-friendly" or "as appropriate".
  - Nothing verified before saving that every requirement had at least one acceptance criterion.

  Keeping the interactive marker resolution in the post-save Step 6 (rather than ahead of the write) honors the framework contract that a spec file is always written even when questions go unanswered — a dismissable `AskUserQuestion` placed before the `Write` would end an unattended turn before the deliverable is produced.

  ## Tests

  Three new lint checks in `tests/lint-prompts.test.js` pin the contracts so they cannot be silently dropped:
  - **Definition of Done checklist presence** — anchors: `Definition of Done`, `Every requirement has at least one acceptance criterion`, `No vague wording remains`, and `offering the assumption you would otherwise make as the recommended first option`.
  - **Ambiguity self-review presence** — anchors: `vague or unmeasurable wording`, `"user-friendly"`, and `Make each one concrete in user-perceivable terms`.
  - **Boundary/error rules presence** — anchors: `boundary and error behavior the user sees` and `at least one acceptance criterion covering the failure path`.

### DIFF
--- a/commands/spec.md
+++ b/commands/spec.md
@@ -88,6 +88,7 @@ Your first goal is to determine the **topic** - the single, specific feature or 
 2.  **Functional Requirements (The "What"):**
     - Ask the user to describe what needs to be done from a user's perspective.
     - For every piece of information the user gives you, think like a tester and clarify ambiguities. If the user answers in technical terms, rewrite the information into plain, user-facing language before including it in the spec.
+    - Where applicable, capture the boundary and error behavior the user sees: what error message appears, what limits exist (file size, format, count), and what happens when the action fails.
     - If the user says: "The user needs to be able to upload a profile picture."
     - Probe with clarifying questions like: "Great. Let's break that down. What file formats should be allowed (e.g., JPG, PNG)? Is there a maximum file size? What should happen after the upload is successful? What specific error message should the user see if it fails?" — for any that stay unanswered, leave a `[NEEDS CLARIFICATION: …]` marker rather than stopping.
     - If information is missing, mark every unresolved detail with `[NEEDS CLARIFICATION: your specific question]` directly in the draft. Example: "The user should see an error message. [NEEDS CLARIFICATION: What should the exact text of the error message be?]"
@@ -96,8 +97,9 @@ Your first goal is to determine the **topic** - the single, specific feature or 
     - After clarifying a requirement, turn it into a concrete, testable acceptance criterion.
     - Acceptance criteria must read as manual QA test scripts that a non-developer could execute. Describe only what is visible on screen and what the user does — never reference internal system behavior.
     - Each acceptance criterion follows the same three-part shape as the example below: a precondition (Given), a user action (When), and a visible outcome (Then). Include Given only when the precondition affects the outcome.
-    - If any `[NEEDS CLARIFICATION: …]` markers remain on the parent requirement in §Functional Requirements, ask clarifying questions and resolve the markers before writing acceptance criteria.
-    - If a clarifying answer reveals a constraint or detail that belongs to the parent requirement (not just the acceptance criterion), update the requirement statement in §Functional Requirements before continuing. The requirement and its acceptance criteria must agree on level of detail.
+    - If any `[NEEDS CLARIFICATION: …]` markers remain on the parent requirement in **Functional Requirements**, ask clarifying questions and resolve the markers before writing acceptance criteria.
+    - If a clarifying answer reveals a constraint or detail that belongs to the parent requirement (not just the acceptance criterion), update the requirement statement in **Functional Requirements** before continuing. The requirement and its acceptance criteria must agree on level of detail.
+    - For requirements that capture boundary or error behavior, include at least one acceptance criterion covering the failure path (e.g., "When the user uploads a file larger than 5MB, then they see: 'File too large. Maximum size is 5MB.'").
     - Example Statement: "Okay, I've captured that. So a clear acceptance criterion would be: 'Given the user is on their profile page, when they upload a PNG file smaller than 5MB, then the new picture appears on their profile and a 'Success' message is shown.' Is that correct?"
 
 4.  **Scope and Boundaries:**
@@ -106,9 +108,15 @@ Your first goal is to determine the **topic** - the single, specific feature or 
     - Focus only on clarifying boundaries within the current **topic** itself.
     - Example: "To keep this focused on [your topic], what related aspects should we explicitly not include? For example, should we include [specific feature within this topic]?"
 
-### Step 4: Self-Review (Language Check)
+### Step 4: Self-Review (Language and Ambiguity Check)
 
 - Before presenting to the user, re-read the entire draft end-to-end. For every sentence, ask: "Would this make sense to someone who has never seen the codebase?" Replace any developer-facing language with plain, non-technical wording in the same language the user is using. Remove any references to internal system behavior, code, or architecture that slipped in.
+- Then re-read **Functional Requirements** — both the requirement statements and their acceptance criteria — for vague or unmeasurable wording: words like "fast", "user-friendly", or "as appropriate" that a tester could not verify. Make each one concrete in user-perceivable terms (e.g., "the search feels fast" becomes "search results appear within 2 seconds"), or, if the user has not decided the specific value yet, replace it with a `[NEEDS CLARIFICATION: …]` marker so it is resolved in Step 6, after the spec is saved. Quantify only where the user would notice the difference — do not force a number onto every sentence; narrative sections like **Overview and Rationale** may stay qualitative.
+
+**Definition of Done.** A self-review, not an approval gate — the file is still written at the end of the process. Confirm the draft meets both:
+
+1.  **No vague wording remains in requirements or acceptance criteria.** The check above already resolved each vague term — made it concrete in user-perceivable terms or converted it to a `[NEEDS CLARIFICATION: …]` marker; confirm none slipped through, and if one did, resolve it per the check above. Any term converted to a marker is resolved with the user in Step 6, or left in place in an unattended run.
+2.  **Every requirement has at least one acceptance criterion.** Confirm that each functional requirement in **Functional Requirements** carries at least one acceptance criterion in the When/Then shape (Given optional). If a requirement has none, write one for it before saving.
 
 ### Step 5: File Generation
 
@@ -118,5 +126,5 @@ Your first goal is to determine the **topic** - the single, specific feature or 
 
 ### Step 6: Final Review and Recommend Next Step
 
-1.  Present the saved specification and ask the user to review it for inaccuracies or missing details. Resolve any `[NEEDS CLARIFICATION: …]` markers with them, apply edits, and re-save. If no answer comes (e.g. an unattended `claude -p` run), leave the markers in place; the user — or `/awos:tech` — can resolve them later.
+1.  Present the saved specification and ask the user to review it for inaccuracies or missing details. Resolve each `[NEEDS CLARIFICATION: …]` marker with them via `AskUserQuestion`, offering the assumption you would otherwise make as the recommended first option (with a free-text option for open-ended markers); fold each answer back into the relevant requirement and its acceptance criteria, then re-save. If no answer comes (e.g. an unattended `claude -p` run), leave the markers in place; the user — or `/awos:tech` — can resolve them later.
 2.  Report the saved path and the next command: `/awos:tech`.
--- a/tests/lint-prompts.test.js
+++ b/tests/lint-prompts.test.js
@@ -946,3 +946,76 @@ test('context/<path> references in prompts are internally consistent', () => {
     );
   }
 });
+
+test('commands/spec.md has a pre-write Definition of Done checklist', () => {
+  // The spec command runs a Definition of Done self-review inside Step 4,
+  // before the file is written: confirm no vague wording remains and every
+  // requirement carries an acceptance criterion. It is a self-review, not an
+  // approval gate — the file is still written. Any `[NEEDS CLARIFICATION]`
+  // marker is resolved with the user post-save in Step 6 (offering the
+  // assumption as the recommended first option), or left in place in an
+  // unattended run; the Definition of Done itself never asks the user a
+  // question. The behavioral proof — no raw markers in the produced
+  // functional-spec.md, every requirement carries a criterion — lives in
+  // awos-qa. This lint only pins that the instruction text is present, so
+  // the contract can't be silently dropped from the prompt later.
+  const body = readUtf8(path.join(commandsDir, 'spec.md'));
+  assert.ok(
+    /Definition of Done/i.test(body),
+    'commands/spec.md must declare a "Definition of Done" pre-write checklist'
+  );
+  assert.ok(
+    body.includes('Every requirement has at least one acceptance criterion'),
+    'commands/spec.md Definition of Done must require every functional requirement to carry at least one acceptance criterion before saving'
+  );
+  assert.ok(
+    /No vague wording remains/i.test(body),
+    'commands/spec.md Definition of Done must gate on no vague wording remaining in requirements or acceptance criteria'
+  );
+  assert.ok(
+    body.includes(
+      'offering the assumption you would otherwise make as the recommended first option'
+    ),
+    'commands/spec.md Step 6 must resolve each [NEEDS CLARIFICATION] marker post-save via AskUserQuestion, offering the assumption you would otherwise make as the recommended first option'
+  );
+});
+
+test('commands/spec.md self-review checks for vague, unmeasurable wording', () => {
+  // The Step 4 self-review must hunt weasel words ("fast", "user-friendly",
+  // "as appropriate") in requirements and acceptance criteria, and either
+  // make them concrete in user-perceivable terms or convert them to a
+  // [NEEDS CLARIFICATION] marker that Step 6 resolves with the user
+  // post-save (or leaves in place in an unattended run). The behavioral
+  // proof — no unverifiable wording in the produced functional-spec.md —
+  // lives in awos-qa. This lint only pins that the instruction text is
+  // present, so the contract can't be silently dropped from the prompt later.
+  const body = readUtf8(path.join(commandsDir, 'spec.md'));
+  assert.ok(
+    /vague or unmeasurable wording/i.test(body),
+    'commands/spec.md self-review must scan requirements and acceptance criteria for vague or unmeasurable wording'
+  );
+  assert.ok(
+    /"user-friendly"/.test(body),
+    'commands/spec.md self-review must name concrete weasel-word examples (e.g. "user-friendly") so the model knows what to hunt'
+  );
+  assert.ok(
+    body.includes('Make each one concrete in user-perceivable terms'),
+    'commands/spec.md Step 4 self-review rule must instruct making each vague term concrete in user-perceivable terms (not technical metrics), or converting it to a [NEEDS CLARIFICATION] marker'
+  );
+});
+
+test('spec.md captures boundary/error behavior as rules in items 2 and 3', () => {
+  // both sentences must be present — item 2 tells the model what to
+  // elicit; item 3 closes the loop by requiring failure-path criteria for them.
+  const body = readUtf8(path.join(commandsDir, 'spec.md'));
+  assert.ok(
+    body.includes('boundary and error behavior the user sees'),
+    'spec.md Step 3 item 2 must contain the boundary/error elicitation rule'
+  );
+  assert.ok(
+    body.includes(
+      'at least one acceptance criterion covering the failure path'
+    ),
+    'spec.md Step 3 item 3 must require failure-path criteria for boundary/error requirements'
+  );
+});


tests/lint-prompts.test.js
```diff
@@ -907,3 +907,70 @@ test('context/<path> references in prompts are internally consistent', () => {
...
+  assert.ok(
+    /Definition of Done/i.test(body),
+    'commands/spec.md must declare a "Definition of Done" pre-write checklist'
+  );
+  assert.ok(
+    /\*\*Assumption:\*\*/.test(body),
+    'commands/spec.md Definition of Done must tell unattended runs to replace each unresolved [NEEDS CLARIFICATION] marker with a visible **Assumption:** and still save the file'
+  );
+  assert.ok(
+    /at least one acceptance criterion/i.test(body),
```
AlexanderMakarov: This assertion is meant to pin DoD item 2 ("Every requirement has at least one acceptance criterion"), but `/at least one acceptance criterion/i` also matches the Step 3 failure-path rule you added at `spec.md:99` ("include at least one acceptance criterion covering the failure path"). So if someone deleted the DoD completeness gate but left the Step 3 rule, this stays green — the item it names isn't actually covered. The other three assertions in this test (`Definition of Done`, `No vague wording remains`, `**Assumption:**`) are unique, so the section as a whole is pinned; it's just this one item that slips through. Anchoring on the unique bolded lead-in would fix it:

```js
assert.ok(
  body.includes('Every requirement has at least one acceptance criterion'),
  '...'
);
```

tests/lint-prompts.test.js
```diff
@@ -907,3 +907,70 @@ test('context/<path> references in prompts are internally consistent', () => {
...
+  assert.ok(
+    /vague or unmeasurable wording/i.test(body),
+    'commands/spec.md self-review must scan requirements and acceptance criteria for vague or unmeasurable wording'
+  );
+  assert.ok(
+    /"user-friendly"/.test(body),
+    'commands/spec.md self-review must name concrete weasel-word examples (e.g. "user-friendly") so the model knows what to hunt'
+  );
+  assert.ok(
+    /concrete in user-perceivable terms/i.test(body),
```
AlexanderMakarov: Minor, optional — same flavor as the DoD one but not load-bearing: `/concrete in user-perceivable terms/i` matches both the Step 4 pass (`spec.md:111`) and DoD item 1 (`spec.md:117`), so on its own it wouldn't catch Step 4 being removed. The sibling assertion `/vague or unmeasurable wording/i` *is* unique to Step 4, so the test still does its job. Leave it or tighten to a Step-4-only phrase (e.g. the `search results appear within 2 seconds` example) — your preference.

docs/optimization/spec-command-optimization.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Summary
+
+The proposed changes raise the **completeness and quality of the requirements** produced by the `/awos:spec` command. Points 3 and 6 are acknowledged as **problems** but are carved out as **separate, multi-command tasks**, because they require changes in other commands. Each optimization point must be validated with a **"quantitative before → quantitative after"** method, gated on **no regression** — **THIS REQUIRES A SEPARATE DISCUSSION.**
+
+---
+
+## Point 1 — Unresolved `[NEEDS CLARIFICATION]` markers reaching the saved file
+
+- **Problem:** The marker is only required to be closed on the parent requirement before acceptance criteria are written — `commands/spec.md:96`: _"If any `[NEEDS CLARIFICATION: …]` markers remain on the parent requirement in §Functional Requirements, ask clarifying questions and resolve the markers before writing acceptance criteria."_ Neither the self-review (Step 4, line 108) nor file generation (Step 6, lines 114–119) blocks saving a file with markers still open in other sections. The marker is introduced at `commands/spec.md:90`: _"mark every unresolved detail with `[NEEDS CLARIFICATION: your specific question]`"_.
```
AlexanderMakarov: The `commands/spec.md:NN` citations here are accurate against `main` as it stands, but this PR's own edits shift every one of them down (~+2 lines) and renumber the steps — so on merge they point at the wrong lines, and "file generation (Step 6, lines 114–119)" names a step that's now Step 7. Same root shows in Point 4 (line 30): "the self-review catches only leakage … weasel words [are] not present in `spec.md` today" reads in the present tense, but this PR is exactly what makes that false. If the doc stays, citing by section/step name and writing the problem statements in past tense ("before this change…") keeps it from contradicting the file it ships with.

commands/spec.md
```diff
@@ -103,15 +105,24 @@ Your first goal is to determine the **topic** - the single, specific feature or
...
 
-### Step 4: Self-Review (Language Check)
+### Step 4: Self-Review (Language and Ambiguity Check)
 
 - Before presenting to the user, re-read the entire draft end-to-end. For every sentence, ask: "Would this make sense to someone who has never seen the codebase?" Replace any developer-facing language with plain, non-technical wording in the same language the user is using. Remove any references to internal system behavior, code, or architecture that slipped in.
+- Then re-read §Functional Requirements — both the requirement statements and their acceptance criteria — for vague or unmeasurable wording: words like "fast", "user-friendly", or "as appropriate" that a tester could not verify. Make each one concrete in user-perceivable terms (e.g., "the search feels fast" becomes "search results appear within 2 seconds"), or, if the user has not decided the specific value yet, replace it with a `[NEEDS CLARIFICATION: …]` marker so it is resolved in Step 5. Quantify only where the user would notice the difference — do not force a number onto every sentence; narrative sections like §Overview and Rationale may stay qualitative.
```
AlexanderMakarov: >like §Overview and Rationale may stay qualitative.

Weak reference - I don't understand where name of section ends. The same for `re-read §Functional Requirements — both`.

Please use `[Link Text](#section-name)` formatting. Here and below in this file.

commands/spec.md
```diff
@@ -103,15 +105,24 @@ Your first goal is to determine the **topic** - the single, specific feature or
...
 
-### Step 5: Final Review
+### Step 5: Specification Definition of Done
+
+Before the final review and saving the file, check the draft against this list. It is a self-review, not a new approval gate — the file is still written at the end of the process.
+
+1.  **No vague wording remains in requirements or acceptance criteria.** Each requirement and criterion is concrete enough that a non-developer can tell unambiguously whether it passes or fails. If a vague term slipped through, resolve it as described in Step 4: make it concrete in user-perceivable terms, or convert it to a `[NEEDS CLARIFICATION: …]` marker — the marker is a temporary hand-off, not a final state; item 3 closes every marker before the file is written.
```
AlexanderMakarov: >No vague wording remains in requirements or acceptance criteria.

I guess it is excessive - big part of step 4 was dedicated to this check. If we make LLM to check itself in the same session (the same as asking person "Are you done?") it rarely works - LLM reads from it's own context that everything is completed already and do nothing (as the person answers "Yes" without checking, from memory). So this usually just a waste of tokens.

We may ask it to `Grep` in the specific file specific word instead - item 3 below.

commands/spec.md
```diff
@@ -103,15 +105,24 @@ Your first goal is to determine the **topic** - the single, specific feature or
...
+2.  **Every requirement has at least one acceptance criterion.** Confirm that each functional requirement in §Functional Requirements carries at least one acceptance criterion in the When/Then shape (Given optional). If a requirement has none, write one for it before saving.
+3.  **No unresolved clarification markers.** Scan the whole document for `[NEEDS CLARIFICATION: …]` markers, not only the section you edited last — this check runs last so it also catches markers created by items 1 and 2. In an interactive session, resolve each one with the user and fold the answer into the relevant requirement and its acceptance criteria. In an unattended session where no one can answer, **do not stop**: replace each marker with a visible `**Assumption:** …` line in plain language and save the file anyway, so the open decision is recorded rather than hidden. Example: `The user sees an error message. [NEEDS CLARIFICATION: exact wording?]` becomes `The user sees an error message. **Assumption:** the message reads "Something went wrong — please try again."`
```
AlexanderMakarov: >In an interactive session, resolve each one with the user

Please be more precise - like using `AskUserQuestion` tool with reasonable defaults - it would make LLM understand (and, therefore, implement) task better and be proactive for user (don't invent units for metrics).

docs/optimization/spec-command-optimization.md
```diff
@@ -0,0 +1,51 @@
+# Spec Command Optimization
```
AlexanderMakarov: TL;DR: better to move this file into separate PR. I don't see how it is used here.

AlexanderMakarov (changes requested):
This tightens `/awos:spec` in sensible places: it makes boundary/error behavior an explicit elicitation rule rather than a buried example, requires a failure-path acceptance criterion for those requirements, adds a weasel-word pass to the self-review, and closes the loop with a pre-write Definition of Done that sweeps `[NEEDS CLARIFICATION]` markers. The prompt edits read clean — the Step 4↔Step 5 cross-references line up, the marker-sweep "runs last so it also catches markers from items 1 and 2" ordering is thought through, the old-Step-5/6 → 6/7 renumbering is applied consistently, and the wording stays plain-declarative (one bold emphasis per rule) per the repo's emphasis convention. Nothing in the actual command blocks merge.

**What is `docs/optimization/spec-command-optimization.md` for?** It's a new pattern (no `docs/optimization/` today) and it isn't shipped by the installer, so this is purely about the repo's own docs. After merge, five of its seven points (1, 2, 4, 5, 7) are *implemented* — the doc then describes problems that no longer exist, in the present tense. The repo's own rule is "docs describe the present, not the history." The genuinely live content is Points 3 and 6 (the deferred cross-command work). Worth deciding: keep it as a durable record (then it wants section/step references instead of line numbers, and past-tense problem statements), or trim it to the deferred tasks and let the PR description carry the rest? Your call — I just don't want it to rot silently.

tests/lint-prompts.test.js
oleksii-shevchuk: Confirmed. `/at least one acceptance criterion/i` also matched the Step 3 failure-path rule at spec.md:99, so DoD item 2 was not pinned — removing line 118 left the test green. The assertion now reads `body.includes('Every requirement has at least one acceptance criterion')`, which occurs only at the DoD line. Verified by mutation: deleting DoD item 2 now turns the test red. The Step 3 failure-path rule stays independently pinned by the boundary test.

tests/lint-prompts.test.js
oleksii-shevchuk: Tightened. The assertion now anchors on the Step 4 rule sentence, `body.includes('Make each one concrete in user-perceivable terms')`, unique to line 111. The example phrase was not used as the anchor: it would leave the rule deletable while the test stayed green, whereas the rule sentence pins what the assertion message describes. Verified by mutation.

docs/optimization/spec-command-optimization.md
oleksii-shevchuk: Resolved by removing the document. The stale `spec.md:NN` citations and the present-tense "not present today" statement are removed with it.

commands/spec.md
oleksii-shevchuk: The section references are disambiguated: all five `§` tokens are replaced with the bold section name (e.g. `§Functional Requirements` → **Functional Requirements**), applied throughout the file.

`[Link Text](#section-name)` was not used because these references point to sections of the generated functional-spec.md and its template, not to headings within spec.md; an anchor in the prompt file would resolve to no target. Bold delimiting removes the ambiguity without a broken link.

commands/spec.md
oleksii-shevchuk: Item 1 now defers to Step 4 — a one-clause recap plus a cross-reference — instead of restating the check as an independent rule. A Grep was not adopted, for two reasons: the Definition of Done runs at Step 5, before file generation at Step 7, so no file exists on disk to grep — the whole checklist, including the marker sweep, operates on the in-context draft; and vagueness is not a fixed token set, so grepping the named example words would miss other wording.

commands/spec.md
oleksii-shevchuk: Done. The interactive branch of item 3 now resolves each marker with the user via `AskUserQuestion`, offering the assumption it would otherwise make as the recommended first option, with a free-text option for open-ended markers. It remains a confirmable choice rather than an auto-accept, and the unattended branch (`**do not stop**` / `**Assumption:**`) is unchanged, so the deliverable is still never gated on an answer. Metric units are not invented: the model proposes a value for the user to confirm, or records an explicit assumption when unattended.

docs/optimization/spec-command-optimization.md
oleksii-shevchuk: The document is removed from this PR. The deferred points (Non-Functional Metrics; Requirement-Priority Signal) will be addressed in their own coordinated PRs.

oleksii-shevchuk:
All review points are addressed in the latest revision — specifics are in each thread. No load-bearing contract changed: the Definition of Done stays a self-review, the file is always written, and the unattended assumption path is intact.

commands/spec.md
AlexanderMakarov: Verified at head — zero `§` tokens remain. And you're right about the anchor links: these are headings of the generated spec (`templates/functional-spec-template.md` lines 9 and 15), not headings in this file, so `(#section-name)` would have been a broken link. Bold is the right call.

commands/spec.md
AlexanderMakarov: The deferral reads well — item 1 is now a hand-off to Step 4 rather than a second full pass, and fair pushback on the Grep: at this step nothing is on disk yet, so there's nothing to grep. One heads-up: the checklist's position is affected by the #141 rebase (see the review summary) — items 1–2 can stay pre-write as they are; item 3 is the one that has to move.

commands/spec.md
AlexanderMakarov: The `AskUserQuestion` mechanics are what I was after — the assumption as the recommended first option is exactly right. Heads-up though: after rebasing onto #141, this interactive branch can't stay ahead of the file write — in `claude -p` a dismissed `AskUserQuestion` ends the turn before the `Write` ever runs, which is the exact failure #141 restructured this file around. Main's Step 6 now does post-save marker resolution; I'd move these mechanics there. Details in the review summary.

tests/lint-prompts.test.js
AlexanderMakarov: Confirmed at head — the exact substring occurs once, only at the DoD item 2 line. The mutation check is appreciated.

tests/lint-prompts.test.js
AlexanderMakarov: Confirmed — unique to the Step 4 rule sentence at head. And agreed the example phrase would have been the weaker anchor: the rule sentence is what the assertion message names.

AlexanderMakarov (commented):
Round 2 — everything from the first pass is addressed — thanks for the careful replies.

One thing now stands between this and merge, and it isn't a defect in this PR: main moved underneath it. [#141](https://github.com/provectus/awos/pull/141) restructured `commands/spec.md` around write-first delivery — the file is written at Step 5 (File Generation) *before* any final review, and Step 6 resolves `[NEEDS CLARIFICATION]` markers with the user *after* the save. Its [editor note](https://github.com/provectus/awos/blob/72b615a94a144ae8824b235fd76f5389816ace70/commands/spec.md#L43) explains why: in `claude -p`, a dismissed `AskUserQuestion` ends the turn, so a deliverable `Write` placed after a dismissable question never runs unattended. The branch is now conflicting in both files, and the resolution is semantic, not textual. Two decisions it forces:

1. **Where the Definition of Done lives.** Items 1–2 (vague-wording confirm, criterion coverage) are pure self-checks with no user interaction — they can stay ahead of File Generation as they are. Item 3's interactive branch, though, puts an `AskUserQuestion` ahead of the deliverable `Write` — exactly the structure #141 exists to prevent. I'd fold item 3's mechanics (offer the assumption you'd otherwise make as the recommended first option) into main's Step 6, which already does post-save marker resolution with the user. That keeps your improvement and main's ordering.

2. **What an unattended run leaves in the file.** Main's contract — codified in [`deliverable commands write their file on an unanswered question`](https://github.com/provectus/awos/blob/72b615a94a144ae8824b235fd76f5389816ace70/tests/lint-prompts.test.js#L407) — is that spec.md's unattended fallback is the `[NEEDS CLARIFICATION: …]` marker itself, left in the saved file for the user or `/awos:tech` to resolve later. This PR replaces markers with `**Assumption:** …` lines before saving. Both record the open decision, but they can't both be the story. I'd keep main's markers — that's the merged, test-pinned contract and the smaller delta. (I checked: nothing greps for the marker outside `commands/spec.md` today — `tech.md` has no marker consumer — so this is about keeping one documented hand-off, not protecting a live consumer. Which also means switching to `**Assumption:**` is viable if you think a recorded assumption is strictly better than an open marker; it just has to be one story, with #141's test comment and the INTERACTION rule updated in the same PR rather than the two drifting apart.)

commands/spec.md
oleksii-shevchuk: Addressed. Definition-of-Done items 1–2 (no vague wording remains; every requirement carries at least one acceptance criterion) remain pre-write self-checks in Step 4, with no user interaction. Item 3's interactive part is relocated to Step 6: post-save resolution of each `[NEEDS CLARIFICATION]` marker via `AskUserQuestion` offers the default assumption as the recommended first option (with a free-text option for open-ended markers), then folds the answer back into the requirement and its acceptance criteria. Main's write-first ordering is preserved.

commands/spec.md
oleksii-shevchuk: Addressed. The `**Assumption:**` fallback is removed. An unattended run now leaves the `[NEEDS CLARIFICATION]` marker in the saved file, per main's contract pinned by the `deliverable commands write their file on an unanswered question` test. Markers are no longer replaced with assumptions before saving, which aligns spec.md with product/roadmap and with Step 6.
Concretely: the `AskUserQuestion` marker-resolution now lives in post-save Step 6, after the file is written in Step 5 — no dismissable question precedes the `Write`, so the `claude -p` failure mode #141 restructured around is gone.

oleksii-shevchuk:
Round 2 addressed. 
Since main moved (#141, write-first), this is a semantic reconciliation rather than a textual merge; the branch is rebased onto `origin/main`.

  - The Step 4/5/6 skeleton and write-first File Generation match main.
  - Four improvements are preserved: boundary/error elicitation and a failure-path acceptance criterion (Step 3); the vague/unmeasurable-wording check and the Definition-of-Done sub-block (Step 4).
  - Interactive marker resolution is moved to post-save Step 6.
  - The `**Assumption:**` fallback is dropped in favor of main's marker contract.

Tests: the `deliverable…` contract is kept verbatim, a lint anchor for the Step 6 mechanic is added, and the stale comments are updated.

commands/spec.md
AlexanderMakarov: Confirmed at head — items 1–2 are pre-write self-checks in Step 4 with no user interaction, and item 3's interactive part now lives in post-save Step 6. That's the split I was after, and the write-first ordering holds.

commands/spec.md
AlexanderMakarov: Confirmed — the `AskUserQuestion` resolution now sits in Step 6 after the Step 5 write, and the unattended path leaves the marker in the saved file per main's contract. No dismissable question precedes the `Write` anymore, so the `claude -p` failure mode is closed.

AlexanderMakarov (approved):
Round 3 — this is all addressed — thanks for the careful semantic merge.

One housekeeping thing before you squash-merge — not a code issue, but the description lands in the commit history and currently describes a design that didn't ship. Three spots:

1. The "**Step 5 — Specification Definition of Done … new step**" section and the "Old Step 5 (Final Review) → Step 6; old Step 6 (File Generation) → Step 7" renumbering. At head there's no Step 7 and no standalone DoD step: the Definition of Done lives inside Step 4, Step 5 is File Generation, Step 6 is Final Review.
2. DoD item 3's "each marker is replaced with a visible `**Assumption:** …` line". That fallback was dropped — unattended runs now leave the `[NEEDS CLARIFICATION: …]` marker in the saved file, per main's contract.
3. The Tests section lists `**Assumption:**` as a Definition-of-Done test anchor; the test doesn't assert on it — its fourth anchor is `offering the assumption you would otherwise make as the recommended first option`.

Refresh those and I'm good to merge.

oleksii-shevchuk:
Thanks — description refreshed to match head. All three:

  1. **Removed** the "Step 5 — Definition of Done … new step" section and the "old Step 5→6 / old Step 6→7" renumbering. The description now reflects head: the Definition of Done lives inside **Step 4**, Step 5 is **File Generation**, Step 6 is post-save **Final Review**.
  2. **Removed** the `**Assumption:**` fallback wording. It now states that an unattended run leaves the `[NEEDS CLARIFICATION: …]` marker in the saved file, per main's contract.
  3. **Fixed** the Tests section: the DoD anchors now list the actual four — `Definition of Done`, `Every requirement has at least one acceptance criterion`, `No vague wording remains`, and `offering the assumption you would otherwise make as the recommended first option`. No `**Assumption:**` anchor.

  One correction beyond the three you flagged, so the credit is right: **Step 5 (write-first) is no longer listed as a change of this PR** — it's byte-identical to main (it shipped for spec.md in your #141, extending the write-first contract #131/#132 established for tasks/tech), so the description now frames it as the pre-existing ordering this change relies on. The diff touches only Steps 3, 4 and 6.

  Ready to merge.

PR #141, opened 2026-06-29
fix: write product/roadmap deliverable on an unanswered question

AlexanderMakarov:
## What & why

`/awos:product` and `/awos:roadmap` gated their deliverable file write behind interactive Q&A. In an unattended `claude -p` run the questions are silently dismissed, so neither command ever wrote its file — the draft was printed as the final assistant text and the turn ended, and `context/product/product-definition.md` / `context/product/roadmap.md` never landed on disk.

This is the same bug already fixed for `/awos:tasks` and `/awos:tech` in #131, which was never propagated to these two commands. Observed before the fix: roadmap prosed and never wrote in 6/6 `-p` runs; product prosed interactively.

## The fix

A dismissed `AskUserQuestion` in `-p` mode **ends the turn**, so any question asked *before* the write blocks the deliverable. The fix removes the pre-write questioning and makes the write unconditional, then refines interactively afterward — #131's "Option 2": keep the interactive UX, make the no-answer path still write.

- Add the INTERACTION rule to both commands: a skipped/unanswered question (e.g. in `claude -p`) is never a stop signal — fall back to a documented default/assumption and continue through the file write.
- **roadmap.md**: Creation Mode hands straight to Finalization; Finalization writes `roadmap.md` unconditionally, then surfaces it for review.
- **product.md**: Creation Mode drafts every section up front from the prompt and best-practice assumptions (no blocking question), Step 3 writes the file without waiting for approval, and a new Step 4 refines it with the user *after* it exists.
- Lock the contract with a Layer 1 lint test asserting the deliverable commands (product, roadmap, tasks, tech) carry the unattended-write fallback rule. Its comment notes that `spec.md` / `architecture.md` still have the same latent gate and are a deliberate follow-up.

## Verification

- Full suite green: `npm test` — 79 pass, 0 fail (includes the new lint test).
- `npx prettier . --check` clean.
- **Behavioral E2E** (new scenarios in awos-qa, run with `claude -p` against this branch): both pass 2/2 — the file is written and populated unattended. The first product attempt (INTERACTION rule only) failed the E2E because it still asked before writing; that's what drove the write-first restructure. See the companion PR: provectus/awos-qa#29.

## Scope note

Intentionally limited to product + roadmap (the reported regression). `spec.md` and `architecture.md` carry the same unfixed gate and are flagged in the test comment for a follow-up PR.

## Summary by CodeRabbit

* **New Features**
  * Product, roadmap, architecture, and spec drafting now continues when questions are skipped/unanswered, using documented defaults/assumptions and producing complete drafts upfront.
  * During unattended runs, deliverables are written immediately, then presented for review with later refinement possible by re-running the relevant command.

* **Bug Fixes**
  * Skipped/unanswered questions are no longer treated as stop signals.
  * Unresolved items are marked for clarification and persist if no response arrives.

* **Documentation**
  * Updated workflow guidance to ensure deliverable writing occurs before any dismissible prompts in unattended execution.

* **Tests**
  * Added a lint test verifying deliverable commands include “never a stop signal” and “including writing” guidance.

### DIFF
--- a/commands/product.md
+++ b/commands/product.md
@@ -66,7 +66,13 @@ First, check if the file `context/product/product-definition.md` exists.
 
 ### Step 2B: Creation Mode
 
-1.  **Brownfield detection.** Check whether the project already has source code by looking for common indicators (`src/`, `app/`, `lib/`, `package.json`, `requirements.txt`, `go.mod`, `Cargo.toml`, `pom.xml`, `Gemfile`, `build.gradle`, `*.csproj`, `Makefile`, `CMakeLists.txt`, `setup.py`, `pyproject.toml`, or similar). If any are found, ask the user whether to run codebase exploration using `AskUserQuestion` with two options: **Yes, explore the codebase** ("Use existing code as context for the product definition") and **No, start from scratch** ("Treat this as a new project — ignore existing code"). If the user chooses to skip or the question goes unanswered, proceed to step 2 as if no source code was found. Otherwise, run a comprehensive exploration before drafting:
+1.  **Brownfield decision.** Decide whether to use existing source code as context, in this order — and never block the write on the choice:
+
+    a. **Prompt intent first.** Read `<user_prompt>`. If it clearly asks to use or explore the existing codebase (e.g. "explore existing code", "use the current codebase", "brownfield"), the decision is **explore**. If it clearly opts out (e.g. "without brownfield detection", "from scratch", "ignore existing code", "greenfield"), the decision is **skip**. Interpret intent with natural-language understanding, not substring matching — "don't explore the codebase" is a **skip**, not an explore. When the prompt states a preference, act on it and do not ask.
+
+    b. **Silent prompt.** If `<user_prompt>` says nothing either way, check whether the project already has source code by looking for common indicators (`src/`, `app/`, `lib/`, `package.json`, `requirements.txt`, `go.mod`, `Cargo.toml`, `pom.xml`, `Gemfile`, `build.gradle`, `*.csproj`, `Makefile`, `CMakeLists.txt`, `setup.py`, `pyproject.toml`, or similar). If none are found, the decision is **skip**. If any are found, ask the user with `AskUserQuestion` — **Yes, explore the codebase** ("Use existing code as context for the product definition") and **No, start from scratch** ("Treat this as a new project — ignore existing code"). This question is reached only in an interactive run whose prompt expressed no preference; an unattended run states its choice in the prompt.
+
+2.  **If the decision is explore,** run a comprehensive exploration before drafting:
 
     a. Launch an `Explore` agent focused on the product domain:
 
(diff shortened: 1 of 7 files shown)


commands/product.md
```diff
@@ -63,7 +66,7 @@ First, check if the file `context/product/product-definition.md` exists.
...
+1.  **Brownfield detection.** Check whether the project already has source code by looking for common indicators (`src/`, `app/`, `lib/`, `package.json`, `requirements.txt`, `go.mod`, `Cargo.toml`, `pom.xml`, `Gemfile`, `build.gradle`, `*.csproj`, `Makefile`, `CMakeLists.txt`, `setup.py`, `pyproject.toml`, or similar). If any are found, ask the user whether to run codebase exploration using `AskUserQuestion` with two options: **Yes, explore the codebase** ("Use existing code as context for the product definition") and **No, start from scratch** ("Treat this as a new project — ignore existing code"). If the user chooses to skip or the question goes unanswered, proceed to step 2 as if no source code was found. Otherwise, run a comprehensive exploration before drafting:
```
dustyo-O: strange, for other steps we maintain default behaviour on skip, here we just skip step 2 on unanswered question, isn't that contradictory?

tests/lint-prompts.test.js
```diff
@@ -404,6 +404,43 @@ test('every core command declares an INTERACTION section', () => {
...
+  // specification"). They are a deliberate follow-up — add them here once the
+  // same fallback rule is ported to them.
+  const deliverableCommands = [
+    'product.md',
+    'roadmap.md',
+    'tasks.md',
+    'tech.md',
+  ];
+  const missing = [];
+  for (const c of deliverableCommands) {
```
dustyo-O: perf: 1-letter variables I would rather to be avoided with global claude settings

dustyo-O:
did you consider for ci to implement some args instead of building a pile a crutches trying to detect we are in ci?

dustyo-O (approved):

commands/product.md
AlexanderMakarov: Fair point — the "skip on unanswered" clause was hiding a real gap rather than resolving it. A dismissed `AskUserQuestion` in `claude -p` ends the turn, so that question (and the triage after it) sit in front of the write, and a brownfield `-p` run would produce nothing.

Reworked in 170a4a2: the brownfield decision now comes from the prompt — "explore existing code" / "without brownfield detection" set the choice and skip the question; we only ask when the prompt is silent **and** source is detected (an interactive-only case). Exploration findings are written to `brownfield.md` before triage, and the accept/reject triage moves to Step 4, after the definition is saved — so the deliverable always lands and there's no special "unanswered" branch to contradict the other steps.

tests/lint-prompts.test.js
AlexanderMakarov: Renamed `c` → `command` in ed66ba9.

AlexanderMakarov:
Agreed — no CI detection and no flag. The brownfield decision now reads the user's free-form prompt ("explore existing code" → explore, "without brownfield detection" → skip), so the command doesn't ask when intent is stated; `AskUserQuestion` stays only for the interactive, prompt-silent case. Same idea as how `/awos:tasks` already reads "skip tests" intent from the prompt. Done in 170a4a2 (re: the thread on `commands/product.md`).

commands/product.md
AlexanderMakarov: Here's everything that landed in response to this review — all on `fix/product-roadmap-unattended-write`, rebased on latest `main`, mergeable with CI green.

**Brownfield decision is now prompt-driven — `170a4a2` (product.md)**
The opt-in `AskUserQuestion` and the finding triage no longer sit in front of the write. The decision is read from the prompt first ("explore existing code" → explore, "without brownfield detection" → skip); only a prompt that is silent *and* a project with source indicators reaches the interactive opt-in question. Exploration writes `brownfield.md` before the draft, and accept/reject triage moves to Step 4, after the definition is saved. A brownfield `-p` run always produces the deliverable, and the "skip on unanswered" clause this thread flagged is gone.

**roadmap.md — `170a4a2`:** same move — capabilities appended before the roadmap is written; triage in Finalization, after the write.

**architecture.md — `e50093d`:** the same two blockers removed (the brownfield tech triage and the "do not proceed until the section is confirmed" gate). Drafts up front from the explored stack + best-practice defaults, writes, then triages. The brownfield.md cleanup step is unchanged.

**spec.md — `942c8c7`:** completes the set. Reuses its existing `[NEEDS CLARIFICATION: …]` marker as the fallback — drafts from context, writes `functional-spec.md` unconditionally (short-name derived from the topic, no question), the approval gate is removed, and review moves after the write.

**Consistency & tests**
- One bold-emphasis rule per file (this + CodeRabbit's note): the structural write stays bold, the fallback bullet is plain. Each command carries a short editor-only note explaining why the prose rule alone isn't sufficient.
- Lint: renamed the 1-letter `c` → `command` (`ed66ba9`); the deliverable-write contract now covers all six doc-generating commands (product, roadmap, architecture, spec, tasks, tech) — the "follow-up" note is dropped. 94/94 Layer-1 tests pass.
- Behavioral: 8 `claude -p` E2E scenarios in provectus/awos-qa#29 (greenfield, explore-intent, skip-intent, roadmap capabilities, architecture brownfield, spec) all pass 2/2 against this branch — each deliverable lands unattended.

On the "args, not CI-crutches" point from the other thread: there's no flag and no `-p` detection — intent comes from the free-form prompt, the same way `/awos:tasks` reads "skip tests".

PR #145, opened 2026-07-13
feat(audit): prevention-coverage dimension — enforced/instructed/absent tiers over existing checks

kmakarychev-dev:
## What

The audit measured **state** ("is the repo in good shape?") but not **stability** ("will it stay that way under continued AI code generation?"). In AI-driven development any property not encoded in a mechanical gate or agent-visible context is one regeneration away from regressing — this PR makes the audit measure that.

New `prevention-coverage` dimension of **8 clusters**, each grouping existing checks by failure mode and scored by a pair:

- **Enforcement** (`PRV-01…08`, `detected`, weight 3) — a mechanism that runs mechanically blocks recurrence: pre-commit/husky/lefthook gate, CI check step, server-side bot (Dependabot/Renovate), or agent hook. Config nothing runs = WARN ("configured, not gated").
- **Instruction** (`PRV-11…18`, `judgment`, weight 2) — the rule is written where agents see it (CLAUDE.md chain, AGENTS.md, rules, skills). Rubrics require quoted passages.

IDs pair at +10 (`PRV-0n ↔ PRV-1n`, codes 3100–3107 / 3110–3117); pairs join via a new `cluster` key, and `covers_checks` on the enforcement half names the source checks each cluster guards (lint-enforced cross-references).

A linkage pass (`prevention.ts`) derives per-cluster **tiers** — `enforced` > `instructed` > `absent` (`pending` while judgments are outstanding) — annotates covered FAIL/WARN checks in `audit.json`, and surfaces **unguarded passes**: checks passing today with nothing preventing regression. The block is recomputed by `audit-core` and every `aggregate` (never preserved/patched), so pending tiers finalize automatically after `patch-judgment`. PRV check records are self-describing (`cluster`/`covers_checks` stamped by `buildCheck`), so no standards access is needed post-audit.

## Surfaces

- **Reports**: a Prevention Matrix section (md + HTML, tier badges, at-risk lists, fragility footer) plus per-check tier annotations on failing covered checks.
- **Org mode**: `computePreventionGaps` rolls up per-cluster tier counts as `prevention_gaps`; PRV checks are excluded from `org_gaps` to avoid double-listing.
- **Orchestration**: judgment count grows 5 → 13; the 8 instruction rubrics share one instruction-file corpus and ride a single subagent (SKILL.md + repo-auditor guidance updated). `report-context` exposes the block and annotations; failing checks in `absent` clusters get two-part recommendations (fix + add guardrail).

## Scoring impact

+40 applicable weight — audit totals are **not comparable with pre-2.3 runs**. `standards_version` bumped to 2026.07; `dist/` rebuilt and committed. Plugin version stays 2.2.0 — release-drafter bumps it from the `minor` label at release time.

## Testing

- 54 new tests across layers: detector fixtures (PASS/WARN/FAIL per gate), tier state machine, end-to-end flow (audit-core → patch-judgment → report-context → aggregate idempotence), render (md/HTML/escaping/org), rollup, standards-schema cross-reference lints (`covers_checks` must resolve, pair/weight contracts), and Layer-1 raw-text lints.
- Full suite green: 93 lint + 1300 engine + 42 installer + 5 fixtures + 20 harness + 11 tools.
- Verified end-to-end on a real repo (hopstoproject): plausible tiers for a Gradle+JS monorepo, byte-identical deterministic output across runs, correct pending→final transitions, org rollup with no PRV leakage into `org_gaps`.

## Known follow-ups (not in this PR)

- `ARCH_TOOL_RX` doesn't recognize Konsist (Kotlin architecture linting) — teams using it would FAIL PRV-05 undeservedly.
- PRV-08 doesn't recognize Danger/reviewdog "docs must change with code" rules.
- Cosmetic: cluster titles capitalize only the first word ("Ai file integrity").
- Judgment variance (5 → 13 verdicts) worth watching via `qa/compare_audit_runs.ts` on early real runs.

## Summary by CodeRabbit

* **New Features**
  * Added **prevention-coverage** analysis with **prevention tiers**, **covered/at-risk** annotations, and **unguarded-pass** reporting.
  * Expanded AI-readiness audits to run **13 judgment checks** (including prevention-instruction checks) in the guidance flow.
  * Added **Prevention Matrix** (single-repo) and **Prevention Gaps (Org)** tables (portfolio rollups).
* **Documentation**
  * Updated audit standards, report format, orchestration, and prevention-gap authoring guidance.
* **Chores**
  * Bumped AWOS plugin/marketplace manifest versions to **2.3.0**.
* **Tests**
  * Added validation for prevention detectors, linkage/aggregation, rendering, and prevention-coverage schema rules.


plugins/awos/skills/ai-readiness-audit/dimensions/prevention-coverage.md
```diff
@@ -0,0 +1,214 @@
...
+- **Category:** 3100
+
+### PRV-02: Dependency risk automation
+
+- **What:** Dependency risk is mechanically managed — vulnerability scanning in CI or an automated update bot
+- **How:**
+  1. Search CI workflow files for a vulnerability scanner (`npm|yarn|pnpm audit`, `pip-audit`, `safety`, `snyk`, `trivy`, `grype`, `osv-scanner`, `dependency-check`).
+  2. Check for update-bot config: `.github/dependabot.yml`, `renovate.json` and variants. A bot counts as enforcement — it runs server-side without anyone invoking it.
+- **Pass:** CI scanner step or update-bot config present
+- **Warn:** Only a lockfile-maintenance config with no scanner and no CI audit
```
AlexanderMakarov: This Warn tier can't be emitted — `detectDependencyRiskAutomation` returns only PASS (any Dependabot/Renovate config, including a lockfile-maintenance-only `renovate.json`, which is in `RENOVATE_PATHS`) or FAIL. And the lockfile-only case is a real config, not a strawman: Renovate's `lockFileMaintenance` runs independently of dependency updates — it refreshes lockfiles only, leaving dependency declarations unchanged, with no scanning (https://docs.renovatebot.com/configuration-options/#lockfilemaintenance). So exactly the state your WARN line describes gets scored full `enforced` at weight 3, and the documented WARN never happens.

PRV-02 is the only binary enforcement detector — every other one has the config-present-but-ungated WARN. `standards.toml`'s PRV-02 definition and the detector comment both say a bot counts as full enforcement, so I'd resolve it in the docs: drop this WARN line (and the uniform PASS/WARN/FAIL wording in `standards.md`). If you do want lockfile-only configs to grade partial, then it's the detector that needs the WARN branch instead — either way the two shouldn't disagree.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,566 @@
...
+  PRECOMMIT_FORMATTER_RX,
+} from './software_best_practices.ts';
+import { detectAgentSafetyHooks } from './security.ts';
+
+// ---------------------------------------------------------------------------
+// prevention-coverage enforcement detectors — categories 3100–3107
+// (PRV-01…PRV-08). One per cluster; the instruction halves (3110–3117) are
+// judgment categories with no detector entry.
+//
+// Uniform grading rule: PASS = an actively running mechanism (invocation
```
AlexanderMakarov: Small thing on this header: two of the eight detectors below don't follow it — PRV-02 (`detectDependencyRiskAutomation`) has no WARN at all, and PRV-06 (`detectTestCoverageGate`) grades on a coverage threshold rather than config-existence. Worth softening to something like "default grading rule (PRV-02 and PRV-06 deviate — see their functions)" so the header doesn't contradict its own code.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,566 @@
...
+    '.eslintrc*',
+    'nx.json',
+  ]);
+  const boundaryConfig = eslintConfigs.find((f) => {
+    const content = readTextSafe(f);
+    return content !== null && ESLINT_BOUNDARIES_RX.test(content);
+  });
+  if (boundaryConfig) {
+    const eslintGated =
+      gateMatches(surfaces, /\beslint\b|\bnpm run lint\b|\byarn lint\b/i)
```
AlexanderMakarov: This gated-eslint check uses `/\beslint\b|\bnpm run lint\b|\byarn lint\b/i` — no `pnpm`. Your own `LINT_GATE_RX` (~line 292) includes `pnpm lint`/`pnpm run lint`, and `pnpm lint` is a real invocation — pnpm aliases any package.json script as a subcommand, so `pnpm lint` is shorthand for `pnpm run lint` (https://pnpm.io/cli/run). Since this detector supports Nx — where `pnpm lint` / `pnpm nx lint` is the usual gate — a repo gating eslint-boundaries only through `pnpm lint` scores WARN instead of PASS. Adding `pnpm` to the alternation lines it up with `LINT_GATE_RX`.

plugins/awos/skills/ai-readiness-audit/references/standards.toml
```diff
@@ -1995,6 +1996,345 @@ url = "https://owasp.org/www-project-software-component-verification-standard/"
...
+dimension = "prevention-coverage"
+weight = 3
+method = "detected"
+cluster = "secrets-hygiene"
+covers_checks = ["AS-05", "AS-12", "AS-13", "AS-14"]
+definition = "A secret-scanning gate (gitleaks, trufflehog, detect-secrets, git-secrets, ggshield, secretlint) runs in pre-commit or CI, mechanically blocking committed credentials"
+applies_when = "always"
+sources = ["audit"]
+reliability_default = "maximal"
+source = "AWOS Prevention Coverage Standard"
```
AlexanderMakarov: One architectural thing I'd flag: all 16 PRV records self-source — `source = "AWOS Prevention Coverage Standard"` with a `url` back into this repo's own `standards.md`. That breaks the convention the rest of the file holds — AS-05/AS-12 → OWASP ASVS, SCS-01 → OWASP SCVS, QA-01/ARCH-01 → Fowler, AIS-01 → OWASP LLM Top 10, DOC-01 → Diátaxis, plus NIST SSDF/DORA. A rubric whose only cited authority is itself is the bias those citations exist to prevent.

Two asks for these (and future) metrics:

1. **A trustworthy external source per metric.** Each cluster already names externally-sourced covered checks — inherit that authority per cluster (secrets → OWASP ASVS; supply-chain → OWASP SCVS; testing/architecture/style → Fowler; ai-file-integrity → OWASP LLM Top 10; docs → Diátaxis). For the *enforcement* claim specifically, anchor on a shift-left / automated-controls standard — NIST SSDF (SP 800-218), already cited elsewhere here, plus SLSA for supply-chain. The instruction half (PRV-11…18) is the one place external literature is genuinely thin; Anthropic's Claude Code best-practices is the honest anchor there.

2. **A numeric value wherever the metric isn't binary.** Anything with a WARN/partial tier is a *graded* metric, so the cited source should supply the threshold that draws the line and the rubric should pin it — e.g. PRV-06 grades on a coverage gate but names no number; something like the Google Testing Blog coverage guidance or DORA gives a defensible band. A partial tier with no cited numeric cutoff is unfalsifiable.

`standards-refresh` is built for exactly this pass (authoritative-source replacement + link verification) and would also flag the circular self-URL.

plugins/awos/skills/ai-readiness-audit/dimensions/prevention-coverage.md
```diff
@@ -0,0 +1,214 @@
+---
+name: prevention-coverage
+title: Prevention Coverage
+description: Whether the good state the other dimensions measure is protected against regression — per failure-mode cluster, is there a mechanical enforcement gate (pre-commit/CI/hook/scanner) or at least an agent-visible written rule?
```
AlexanderMakarov, line 4: A top-level design question, not a blocker — worth deciding deliberately rather than by default: does prevention need to be its own **scored dimension**, or could the same signal ride as a **derived lens** over the existing checks?

The signal is real and *orthogonal* — "is the good state enforced against regression" is a different question from "is the state good," and it cuts across every domain. That's a solid argument for a dedicated dimension plus the Prevention Matrix as a first-class view. Two alternatives worth weighing out loud:

- **Zero-weight lens over existing checks.** `annotateCoveredChecks` already stamps each covered check with its tier, so the matrix could be a pure report layer with no new scored categories. Counter: a zero-weight lens doesn't move the score, so it doesn't push teams to add gates — which is the point. If prevention should count, it needs weight, and a dimension is the clean home for that weight.
- **Distribute the pairs into their home dimensions** (secrets-hygiene → application-security, etc.). Keeps related things together, but blurs each dimension's subtotal — mixing "how good is the state" with "how well is it enforced," two different questions — and scatters the cross-cutting matrix.

I'd keep it a separate dimension (orthogonal axis + clean matrix + weight that actually counts). The one real cost to name: the same concern is now represented twice — AS-05 ("are secrets committed") and PRV-01/11 ("is there a gate") — defensible as two distinct failures, but it's why PRV had to be special-cased out of `org_gaps`. Mostly I'd want the choice stated as a choice.

workshur (approved):

AlexanderMakarov (commented):
Solid and well-tested. A few things:

- **PRV-02 (inline — the one decision):** its spec documents a WARN tier the detector can't emit, so a lockfile-only Renovate config scores full `enforced` at weight 3. Drop the WARN line, or implement the WARN branch.
- **Sourcing (inline — structural):** all 16 PRV records self-cite "AWOS Prevention Coverage Standard" back into this repo, breaking the external-authority convention the rest of the file holds (OWASP/Fowler/NIST SSDF/DORA/Diátaxis). Two asks — a trustworthy external source per metric, and a numeric cutoff wherever a metric isn't binary. Note that WARN tier with no cited threshold is unfalsifiable. `standards-refresh` skill is the home for it.
- **Plugin version:** the bump was reverted (`ed5c3c44`, "release-drafter owns the version"), so plugin.json + marketplace.json stay 2.2.0. But release-drafter owns the *npm release* version (CLAUDE.md:164, PR labels), not the *plugin* version — CLAUDE.md:123 says a plugin-behavior change bumps both plugin files manually. This is a big behavior change (+40 weight, scores not comparable), so the plugin version should move. If release-drafter is meant to own the plugin version too, then CLAUDE.md:123 is stale and should be updated; right now they contradict.
- **Dimension design (inline):** own scored dimension vs a weighted lens over existing checks vs distributed into home dimensions — a design question, not a blocker; note is on the dimension spec.
- Minor (inline): "Uniform grading rule" header overstates (PRV-02/PRV-06 deviate); eslint-gate probe drops `pnpm` while `LINT_GATE_RX` includes it; org_gaps-exclusion is tested against pre-filtered input, not through `rollup_input.ts` where it's enforced.

(CodeRabbit was mid-review when I posted — reconcile before submitting.)

plugins/awos/skills/ai-readiness-audit/dimensions/prevention-coverage.md
kmakarychev-dev: Fixed in 47eedc2 — went with the detector-side resolution since scoring a lockfile-only config as full enforcement did oversell it. `detectDependencyRiskAutomation` now classifies each Renovate config: `lockFileMaintenance` as the only update-driving key (besides `$schema`) → WARN with a `lockfile-maintenance-only config` evidence line; any config with real update drivers (`extends`, `packageRules`, manager blocks) or any Dependabot config stays PASS. Unparseable configs fall back to a conservative regex. The dimension doc's Warn line now describes exactly this case, and standards.md's uniform WARN wording names the PRV-02 deviation. Tests added for both the WARN and the `extends`-still-PASS cases.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
kmakarychev-dev: Fixed in 58ed802 — the header now reads "Default grading rule" and explicitly names both deviations: PRV-02 (WARNs on a lockfile-maintenance-only bot rather than config-without-gate — its axis changed with the fix from the other thread) and PRV-06 (grades on coverage-threshold presence rather than config existence).

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
kmakarychev-dev: Fixed in e0f9592 — extracted the alternation into a named `ESLINT_GATE_RX` (`eslint | npm run lint | pnpm (run )?lint | yarn lint`) aligned with `LINT_GATE_RX`'s package-manager coverage, and added a test for the exact case you described: `eslint-plugin-boundaries` config gated only via `pnpm lint` in CI now scores PASS.

plugins/awos/skills/ai-readiness-audit/references/standards.toml
kmakarychev-dev: Fixed in b144f2d — took your per-cluster inheritance approach for all 16 records: enforcement halves now cite NIST SSDF SP 800-218 (secrets + SAST gates), OWASP SCVS (dependency automation), Martin Fowler CI/CD / Software Architecture Guide (style + boundary gates), Google's Code Coverage Best Practices (test gate), OWASP Top 10 for LLM Applications (agent surface), and Diátaxis (docs); all 8 instruction halves cite Anthropic's Claude Code best practices, per your point that it's the honest anchor there. Every URL is reused verbatim from already-link-verified records elsewhere in standards.toml, so no new linkcheck surface.

On the numeric ask: PRV-06's definition now cites Google's 60/75/90 bands and states explicitly that only the gate's *presence* is graded — the threshold value is not scored (AWOS heuristic, matching what the detector actually parses). I went with the honest-disclosure route rather than pinning a number the detector doesn't check, to avoid re-creating the doc/detector mismatch from the PRV-02 thread. The tier model and 3/2 weights stay declared as AWOS calibration in standards.md. Agreed a `/standards-refresh` pass should police this going forward — it will now find real sources to verify.

plugins/awos/skills/ai-readiness-audit/dimensions/prevention-coverage.md
kmakarychev-dev: Documented in 6377584 — agreed with your conclusion, and the choice is now stated as a choice in the dimension doc's intro: a "Why a separate scored dimension" paragraph names both rejected alternatives (zero-weight lens: doesn't move the score, and moving the score is the point; distribution into home dimensions: blurs "how good is the state" with "how well is it protected" in one subtotal and scatters the matrix) and the accepted cost you flagged — the same concern represented twice (AS-05 as state, PRV-01/PRV-11 as prevention), which is exactly why PRV is special-cased out of `org_gaps` in favor of `prevention_gaps`.

kmakarychev-dev:
Re the versioning point in the review: you're right, and the `ed5c3c4` revert rationale was wrong — fixed in 5967596.

What the history shows: release-drafter owns only the **npm installer version** (`publish.yml` rewrites `package.json` from the release tag; no workflow touches the plugin manifests), while the plugin version has always moved manually in PRs (2.1.0 in #112, 2.2.0 in #139 — the same commit that added the version-pin lint). So the pin's real contract is "plugin.json + marketplace.json + the pinned literal move together in one deliberate commit," and the lint's failure message ("release version is managed by release-drafter, not bumped per change") was the misleading text that caused the bad revert.

5967596 does four things: restores the 2.3.0 bump in both manifests (warranted — +40 applicable weight, non-comparable totals, and `engineVersion()` stamps this value into `audit.json` provenance and the report footer, so 2.2.0 would mislabel every audit the new engine produces); moves the lint pin to an `EXPECTED_PLUGIN_VERSION = '2.3.0'` constant; rewords the lint message to state the three-file contract; and replaces CLAUDE.md's ambiguous line with an explicit description of the two independent version lines.

Kept it at **2.3.0** rather than 3.0.0 on the house scale: #139's full deterministic-engine rework was 2.1.0 → 2.2.0, and the score-comparability break is already flagged by `standards_version` 2026.07 in standards.md.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,610 @@
...
+ * declares `lockFileMaintenance` and no other update-driving keys. Presets
+ * (`extends`), `packageRules`, or manager blocks all mean real dependency
+ * updates happen. Unparseable configs fall back to a conservative regex:
+ * mentions lockFileMaintenance and neither extends nor packageRules.
+ */
+function isLockfileMaintenanceOnly(content: string): boolean {
+  try {
+    const cfg = JSON.parse(content) as Record<string, unknown>;
+    if (cfg.lockFileMaintenance === undefined) return false;
+    const drivers = Object.keys(cfg).filter(
```
AlexanderMakarov: Nice fix, but there's a gap in how it decides a config is "lockfile-only". The docstring says the tell is a real update driver — `extends`, `packageRules`, a manager block — yet the code treats *any* key besides `lockFileMaintenance` and `$schema` as one. Renovate configs almost always carry housekeeping keys like `timezone`, `schedule`, or `labels`, and those don't drive updates — but they're enough to convince the check that updates are happening, so a genuinely lockfile-only config slips past the new WARN and scores full enforcement.

The two cases you tested — `{lockFileMaintenance}` alone, and one with `extends` — are both still right. It's the common in-between that lands wrong: lockfile maintenance plus a schedule or a couple of labels. Narrowing the driver test to the keys the docstring actually names would line them back up. Non-blocking.

AlexanderMakarov (approved):
Re-review — this all looks addressed. The PRV-02 WARN branch, the grading-rule header, the pnpm gate, the external sourcing, the dimension-vs-lens writeup, and the version-line split are all in, and they're tested. Happy to approve.

The PRV-02 fix does leave one small gap that's worth a look — flagged it inline, but it's not a blocker.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
kmakarychev-dev: Fixed in 939f385 — the driver test now ignores a named `RENOVATE_HOUSEKEEPING_KEYS` set (timezone, schedule, labels/addLabels, assignees/reviewers, PR/branch limits, dependency dashboard, semantic-commit keys, onboarding) instead of treating every extra key as a driver, so your in-between case — `lockFileMaintenance` plus a schedule and labels — now grades WARN. Anything outside the set (extends, packageRules, manager blocks, and deliberately any *unknown* key) still counts as an update driver, so the WARN can only under-trigger rather than falsely demote a real update bot. Added tests for the housekeeping-keys WARN and a `packageRules`-still-PASS case.

plugins/awos/skills/ai-readiness-audit/dimensions/prevention-coverage.md
AlexanderMakarov: Confirmed at the tip: the WARN branch exists (`prevention_coverage.ts:301`), the doc's Warn line now describes exactly the detector's behavior, and standards.md names the deviation. Tests cover both directions. Resolving.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
AlexanderMakarov: Header at the tip names both deviations as asked. Resolving. (Minor, non-blocking: the dimension doc still opens with "The uniform grading rule for enforcement checks" without naming the deviations — the per-check Pass/Warn/Fail lines below are accurate, so fine to fold into a follow-up.)

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
AlexanderMakarov: Verified — `ESLINT_GATE_RX` covers `pnpm lint`/`pnpm run lint` and the test pins the exact scenario. Resolving. (`pnpm nx lint` is still outside both regexes; fine as a future widening if it ever bites.)

plugins/awos/skills/ai-readiness-audit/references/standards.toml
AlexanderMakarov: All 16 records verified against the external anchors, self-citation is gone, and PRV-06's presence-only disclosure is the honest version of the numeric ask — pinning a number the detector doesn't parse would recreate the doc/detector mismatch from the other thread. Resolving; `/standards-refresh` polices the links from here.

plugins/awos/skills/ai-readiness-audit/dimensions/prevention-coverage.md
AlexanderMakarov: Paragraph covers both alternatives and the accepted cost. Resolving.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
AlexanderMakarov: Verified — housekeeping set matches the docstring, unknown keys fail toward PASS as designed, tests pin both directions. Resolving. One handoff: top-level `enabled: false` was an unknown key here, so a disabled config got promoted to PASS — that's now fixed in 4f8ce8c via the CodeRabbit thread on this function.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,636 @@
...
+      const content = readTextSafe(f);
+      if (content !== null) {
+        surfaces.push({
+          file: relative(repoPath, f),
+          kind: 'agent-hooks',
+          content,
+        });
+      }
+    }
+  }
```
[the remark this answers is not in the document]
AlexanderMakarov: Valid in principle — an unregistered script under `.claude/hooks` does count as a gate surface today. Full hook-command parsing is out of proportion to this detector family's documented precision tier (grep-heuristic, SCS-06 parity), and scripts in those two dirs are conventionally registered. Deferring the parse; a cheaper follow-up is to gate hook-script surfaces on the presence of a `hooks` key in `.claude/settings*.json`.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,636 @@
...
+      'dependency risk mechanically managed (CI vulnerability scan and/or update bot)',
+      ...ciHits.slice(0, 5).map((h) => `ci scanner: ${h}`),
+      ...updateBots.map((c) => `update bot: ${c}`),
+    ]);
+  }
+  if (lockfileOnlyBots.length > 0) {
+    return makeResult('WARN', lockfileOnlyBots.length, [
+      'Renovate only maintains lockfiles — no dependency updates and no vulnerability scanning',
+      ...lockfileOnlyBots.map((c) => `lockfile-maintenance-only config: ${c}`),
+    ]);
```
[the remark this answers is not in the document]
AlexanderMakarov: Confirmed and fixed in 4f8ce8c. `{"enabled": false}` landed in `updateBots` (no `lockFileMaintenance` key, and `enabled` was an 'unknown key = driver' under the housekeeping-set rule), and an empty dependabot.yml passed on presence — a disabled bot scoring weight-3 enforcement is the exact false-PASS this dimension exists to prevent. Now: Renovate configs are classified disabled/lockfile-only/update-bot with top-level `enabled: false` checked first, and a Dependabot config only counts with an `updates:` block declaring a `package-ecosystem` entry. Tests cover all three cases.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,636 @@
...
+
+export function detectCodeStyleGated(
+  repoPath: string,
+  _params?: unknown
+): ReturnType<typeof makeResult> {
+  const surfaces = gateSurfaces(repoPath);
+  const hits = [
+    ...gateMatches(surfaces, LINT_GATE_RX, ['pre-commit', 'ci']),
+    ...gateMatches(surfaces, PRECOMMIT_FORMATTER_RX, ['pre-commit']),
+  ];
```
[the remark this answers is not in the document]
AlexanderMakarov: Agreed and fixed in 4f8ce8c — and the alternation was near-redundant anyway: a pre-commit config with real lint/format hooks matches the linter names directly as a pre-commit surface, so `pre-commit run` only ever fired in the case it got wrong. Removed, with a test pinning the secrets-only pre-commit case to FAIL.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,636 @@
...
+  // ArchUnit: a build-file dependency or an Arch test class means the rules
+  // run with the test suite.
+  for (const buildGlob of ['build.gradle', 'build.gradle.kts', 'pom.xml']) {
+    for (const f of iterFiles(repoPath, [buildGlob])) {
+      const content = readTextSafe(f);
+      if (content !== null && ARCHUNIT_RX.test(content)) {
+        return makeResult('PASS', 1, [
+          `ArchUnit dependency found in ${relative(repoPath, f)} — boundary rules run with the test suite`,
+        ]);
+      }
```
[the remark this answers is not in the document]
AlexanderMakarov: Valid — dependency-only PASS is inconsistent with PRV-06's 'no CI test gate → FAIL' stance. Deferring to a follow-up: require an ArchUnit rule/test-class grep (`ArchRule`/`@ArchTest`/`@AnalyzeClasses`) before PASS, dependency-without-rules → WARN. The CI-runs-tests cross-check can ride the same change.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -0,0 +1,636 @@
...
+
+  const ciHits = gateMatches(gateSurfaces(repoPath), AGENT_FILE_CI_RX, ['ci']);
+  if (ciHits.length > 0) {
+    return makeResult('PASS', ciHits.length, [
+      'CI checks agent configuration files',
+      ...ciHits.slice(0, 5).map((h) => `ci check: ${h}`),
+    ]);
+  }
+  // WARN (hooks exist but guard nothing) or FAIL from the delegate stands.
+  return delegated;
```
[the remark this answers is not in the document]
AlexanderMakarov: Both halves confirmed. The empty-hooks PASS is the preexisting AIS-07 delegate (`security.ts` — untouched by this PR) and should be fixed there, where PRV-07 will inherit it. The `AGENT_FILE_CI_RX` half is new here and sharper — `paths-ignore: ['CLAUDE.md']` currently PASSes as 'CI checks agent configuration files'. Tracking both as a follow-up: require nonempty hook config in the delegate; constrain the CI-mention route to command contexts (`run:`/`entry:` lines).

plugins/awos/skills/ai-readiness-audit/references/standards.toml
```diff
@@ -1995,6 +1996,346 @@ url = "https://owasp.org/www-project-software-component-verification-standard/"
...
+cluster = "docs-freshness"
+covers_checks = ["DOC-01", "DOC-02", "DOC-03", "DOC-04", "DOC-07"]
+definition = "Documentation is mechanically checked — a link checker, markdown lint, or strict docs build (lychee, markdown-link-check, linkinator, markdownlint, vale, mkdocs --strict) runs in CI or pre-commit"
+applies_when = "always"
+sources = ["audit"]
+reliability_default = "maximal"
+source = "Diátaxis — quality (completeness)"
+url = "https://diataxis.fr/quality/"
+date = "2024"
+last_verified = "2026-07-06"
```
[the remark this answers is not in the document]
AlexanderMakarov: Fair on the wording: the TOML definition is honest but the cluster label, standards.md row, and the detector's 'stale docs are not mechanically caught' FAIL line promise drift detection the mechanism doesn't do. Follow-up: reword to 'docs validation' semantics (evidence strings + standards.md row + dimension doc); a real drift mechanism is out of scope for this heuristic tier.

plugins/awos/skills/ai-readiness-audit/tests/org-rollup.test.ts
```diff
@@ -686,3 +686,104 @@ test('org_rollup: org_connections is present and non-empty for repos without tec
...
+    },
+  ]);
+  // rollup_input filters PRV checks before they reach rollup(); this guards
+  // the rollup-side contract that org_gaps only carries what it was given.
+  assert.ok(result.org_gaps!.every((g) => !g.check_id.startsWith('PRV-')));
+  assert.ok(
+    result.prevention_gaps,
+    'the PRV signal arrives via prevention_gaps'
+  );
+});
```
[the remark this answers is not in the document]
AlexanderMakarov: Correct — the filter is in `readPerRepoAudit` (`rollup_input.ts:114-117`) and nothing imports `rollup_input` from tests, so the exclusion contract is untested (the assertion here is vacuously true). Follow-up: a `readPerRepoAudit` fixture test with a real PRV check, asserting `checks` excludes it while `prevention` stays populated.

AlexanderMakarov:
Re https://github.com/provectus/awos/pull/145#issuecomment-4980703144 (versioning): end state checks out — both manifests and the lint pin agree on 2.3.0, the lint message now states the actual three-file contract, and CLAUDE.md documents the two independent version lines. Good analysis on the publish.yml scope — that was the missing fact.

Also verified all six "Fixed in <sha>" replies against the tip and resolved those threads; the CodeRabbit findings are dispositioned in-thread (two fixed in 4f8ce8c, five deferred as follow-ups).

PR #147, opened 2026-07-14
fix(implement): gate on not-user-reviewed marker and verify subagent claims

AlexanderMakarov:
## What

Backlog source: [Effort-Profit matrix, row 62](https://docs.superhuman.com/d/_dCoNoiEfeSz#Effort-Profit-matrix_tujgsglY/r62&view=center).

Two trust gates added to `commands/implement.md`:

1. **Unreviewed-plan gate.** Step 1 now checks the top of the target spec's `tasks.md` for the `<!-- not-user-reviewed -->` marker that `/awos:tasks` writes when it saves a draft plan (and removes only after user review). If the marker is present, the command tells the user the plan is an unreviewed draft, points at re-running `/awos:tasks` to finish the review, and stops — reviewing a plan happens only in `/awos:tasks`.
2. **Subagent claim verification.** Step 4 previously said to assume a subagent success signal means the task completed. It now treats the report as a claim, not a fact — aligned with the flow templates' wording: spot-check one or two of the files the report names, confirm the definition-of-success verification commands were run, and on a mismatch take the failure path (no `[x]`, stop the loop, surface the discrepancy). The orchestrator still never edits code itself; a gap found in the spot-check is a failure to surface, not something to fix.

## Why

Without the gate, an unreviewed draft `tasks.md` executed immediately — task selection never consulted the marker, making the review step in `/awos:tasks` skippable by accident. And blind trust in subagent success signals let unverified (or hallucinated) completions advance the plan and flip checkboxes.

## Tests

Layer-1 lint test locks the contract: the literal marker string as the join key with `commands/tasks.md`, the claim-not-fact framing in Step 4, and the absence of the old assume-success instruction. `npm run test:lint`: 93/93 pass; prettier clean.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance to pause or proceed when an implementation plan is marked as unreviewed.
  * Strengthened completion checks by requiring verification of reported changes and successful validation commands.

* **Tests**
  * Added coverage ensuring draft-plan handling and completion verification guidance remain enforced.

### DIFF
--- a/commands/implement.md
+++ b/commands/implement.md
@@ -41,7 +41,8 @@ Follow this process precisely. Steps 2–5 form the per-task loop: repeat them f
 1.  Analyze `<user_prompt>`. If it names a specific task, set scope to that single task in the spec it belongs to. If it names a spec (without a specific task), set the target spec from the prompt and set scope to "every incomplete (`[ ]`) task in that spec".
 2.  Otherwise (no prompt): scan `context/spec/` in order, find the first directory whose `tasks.md` has an incomplete item (`[ ]`), select it as the target spec, and set scope to "every incomplete task in that spec".
 3.  If no target can be determined (ambiguous prompt, or all tasks are done), tell the user and stop.
-4.  Load the static spec context once, in parallel:
+4.  Check the top of the target spec's `tasks.md` for the `<!-- not-user-reviewed -->` marker. `/awos:tasks` writes it when it saves a draft plan and removes it only after the user has reviewed the plan — if it is still present, this plan was never reviewed. Ask via `AskUserQuestion` how to proceed: "Stop and review the plan first" (recommended — re-run `/awos:tasks` to finish the review) or "Proceed with the unreviewed draft". On stop, end the run. On proceed, continue but leave the marker in place — proceeding accepts the risk for this run only; it is not a review.
+5.  Load the static spec context once, in parallel:
     - `[target-spec-directory]/functional-spec.md`
     - `[target-spec-directory]/technical-considerations.md`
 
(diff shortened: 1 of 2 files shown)


commands/implement.md
```diff
@@ -41,7 +41,8 @@ Follow this process precisely. Steps 2–5 form the per-task loop: repeat them f
...
 2.  Otherwise (no prompt): scan `context/spec/` in order, find the first directory whose `tasks.md` has an incomplete item (`[ ]`), select it as the target spec, and set scope to "every incomplete task in that spec".
 3.  If no target can be determined (ambiguous prompt, or all tasks are done), tell the user and stop.
-4.  Load the static spec context once, in parallel:
+4.  Check the top of the target spec's `tasks.md` for the `<!-- not-user-reviewed -->` marker. `/awos:tasks` writes it when it saves a draft plan and removes it only after the user has reviewed the plan — if it is still present, this plan was never reviewed. Ask via `AskUserQuestion` how to proceed: "Stop and review the plan first" (recommended — re-run `/awos:tasks` to finish the review) or "Proceed with the unreviewed draft". On stop, end the run. On proceed, continue but leave the marker in place — proceeding accepts the risk for this run only; it is not a review.
```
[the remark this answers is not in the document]
AlexanderMakarov: The inline tool name is deliberate and matches the repo convention. The guideline you cite governs where the tool default is declared — core `commands/*.md` vs the `claude/commands/*.md` wrappers — not whether PROCESS steps may name the tool. Sibling commands name `AskUserQuestion` inline at the decision point throughout: `product.md` (Steps 2b, 4), `roadmap.md`, `architecture.md`, `tasks.md` (three steps), `spec.md`, `verify.md`. The explicit name is also load-bearing here: a plain-prose question ends a non-interactive `claude -p` turn before the stop/proceed decision is captured (`tasks.md` states this failure mode verbatim for its own review gate). Keeping as is.

dustyo-O: too complicated, it's enough to stop here and ask to launch awos:tasks

it shouldn't be the issue, because the case seems to be quite rare.

other option -> just launch user approval in that case, but it requires copy-paste from tasks command, which I would rather to avoid

dustyo-O (approved):

commands/implement.md
AlexanderMakarov: Simplified to an unconditional stop: on finding the marker, implement now tells the user the plan is an unreviewed draft, points at re-running `/awos:tasks`, and ends the run. The AskUserQuestion branch and the proceed-with-draft marker semantics are removed. This also closes the headless hole — a question with a "proceed" default would have let drafts execute silently, which is the case the marker exists to block. Agreed on not duplicating the review flow here; `/awos:tasks` stays the only place a plan gets approved.

PR #152, opened 2026-07-20
feat(prompts): bake the verification-before-completion reflex into agent prompts

AlexanderMakarov:
Closes the "Bake the verification-before-completion reflex into agent prompts" backlog item.

## Problem

The existing rule — *"a subagent's report is a claim, not a fact"* (`commands/implement.md` Step 4, both flow templates) — covers the **orchestrator distrusting a subagent**. Nothing constrained what an agent claimed about **its own** work, and `templates/agent-template.md`, the ancestor of every hired specialist, carried no verification discipline at all: 18 lines, neither rule. So the leverage point where the reflex would be inherited by every agent had nothing to inherit.

## What changed

Two rules, borrowed from the superpowers `verification-before-completion` skill:

1. **A completion claim cites its evidence** — run the check that proves the behavior, report its actual output; never claim something works ("done", "should work", "probably") without fresh command output from the same run. The form is picked by fit, without assuming a specific tool exists: tests or the command that exercises the change; for anything a user sees, the real UI driven through the project's browser-automation tooling with a screenshot to `docs/screenshots/`; for APIs, data, and business logic, `curl`, shell, a CLI invocation, log or database inspection, or a configured MCP tool.
2. **A new test is proven with RED validation** — revert the change it covers, watch the test fail, restore the tree exactly, watch it pass. A test that never failed guards nothing.

| File | Change |
| --- | --- |
| `templates/agent-template.md` | Both rules, with the sanctioned evidence forms named. The payload — every hired specialist inherits them. |
| `commands/implement.md` | A `<completion_evidence>` block composed into each delegation prompt, so a subagent constrains its own report before the orchestrator's spot-check ever sees it. |
| `plugins/awos/templates/implement-feature-template.md` | Orchestrator spot-check of one testing-slice test in Step 6, honoring `<!-- skip-tests: true -->`. |
| `tests/lint-prompts.test.js` | Assertions for each contract; stale comment removed (see below). |
| `tests/README.md` | Layer 1 contract list updated. |

**An opt-out of tests does not opt out of evidence.** `verify.md:41` already settled this tension for the framework — *"skip-tests suppresses test suites, not look-and-feel"* — and the reflex follows it. Rule 1 is unconditional: under an opt-out the evidence becomes a render, CLI, or MCP check instead of a test run. Rule 2 applies only when the work has an agent write a test, so a user or project that opted out leaves it inert rather than unmet — and no agent writes a test just to satisfy it. `commands/implement.md` had **no** `skip-tests` awareness before this PR (unlike `verify.md`, `fix-bug-template.md`, `delivery-flow-template.md`, and `flow.md`); its `<completion_evidence>` block now drops the RED-validation clause when the spec's `tasks.md` carries the marker, and keeps the evidence requirement.

**Vocabulary is joined, not coined.** `commands/tasks.md` already emits *"Write tests with RED validation (must fail before implementation is confirmed done)"* in its testing slice, and `verify.md:39-40` already names the evidence forms. The new text reuses both rather than inventing parallel terms, and a lint assertion pins the literal "RED validation" wording in `tasks.md` so a rename there cannot silently strand the references.

**Division of labour is explicit.** `fix-bug-template.md` deliberately keeps the revert/restore maneuver orchestrator-side. Here the subagent proves the tests it writes and the flow orchestrator independently spot-checks one — stated on both sides so the maneuver is not performed twice on the same tree, with a "restore the tree exactly" guard against a failed stash leaving a broken working tree.

**One stale contract removed.** `tests/lint-prompts.test.js` carried a comment stating a verification block was *intentionally not* asserted and that teams wanting one should add it via wrapper customization. This change reverses that decision, so the comment is deleted rather than annotated, and `<completion_evidence>` folds into the existing required-XML-block list.

## Deliberately unchanged

- **`commands/verify.md`** — its Step 3 already requires evidence for every acceptance criterion and forbids marking one without confirming it. Adding the rule there would restate it, which the repo's prompt guidance warns against.
- **`plugins/awos/templates/fix-bug-template.md`** — already carries the full red-proof, and its existing lint pin is untouched.
- **Step 7 (verify) was rejected** as a home for the feature-template paragraph: `verify.md:37` declares `/awos:verify` "a look-and-feel + spec-freshness check, not a test runner", so a test proof there would contradict it head-on. Step 6 is the structural analog of fix-bug's regression-test stage.

## Verification

- `npm run test:lint` — **125 pass, 0 fail**. (Run with a real Node toolchain on PATH; a plain `npm run` can resolve `node` to a Bun shim that cannot execute `node:test`.)
- `npx prettier . --check` — clean.
- **Every new and rewritten assertion proven red→green individually**: each contract was broken in place, the targeted run confirmed to fail with the contract-naming message, then restored and confirmed to pass. An assertion that never failed proves nothing — including the ones in this PR.
- No version bumps (release-drafter owns the npm line; no plugin behavior change warranting the three-file plugin bump). No `dist/` rebuild — no engine `.ts` touched.

## Follow-up, not in this PR

`plugins/awos/templates/fix-bug-template.md` delegates its fix via `**[Agent: name]**` with no completion-evidence injection, unlike `commands/implement.md`. Real asymmetry surfaced during review, but fixing it reaches past this backlog item — worth its own change once we decide whether `agent-template.md` inheritance alone is the intended rationale for both flows.

## Summary by CodeRabbit

* **Improvements**
  * Added stronger completion verification requirements for implementation tasks, including fresh command output as evidence.
  * Added RED validation guidance for new tests, ensuring tests fail before the related change and pass afterward.
  * Added post-implementation spot checks to confirm tests genuinely cover newly added behavior.
  * Documented support for opting out of test validation when `<!-- skip-tests: true -->` is specified.

* **Tests**
  * Expanded prompt linting to verify completion evidence, RED validation, and test opt-out behavior.


AlexanderMakarov:
Tested by https://github.com/provectus/awos-qa/pull/41

tests/lint-prompts.test.js
```diff
@@ -957,6 +956,75 @@ test('fix-bug template reads remote links, sweeps all surfaces, and verifies sub
...
+    implement.includes('<!-- skip-tests: true -->'),
+    'commands/implement.md <completion_evidence> block must honor the <!-- skip-tests: true --> marker — drop the RED-validation clause under an opt-out while keeping the evidence requirement'
+  );
+
+  const featureTemplate = readUtf8(
+    path.join(pluginTemplatesDir, 'implement-feature-template.md')
+  );
+  assert.ok(
+    /RED validation/.test(featureTemplate) &&
+      /revert|stash/i.test(featureTemplate),
```
dustyo-O, line 1019: strange |

dustyo-O (approved):

tests/lint-prompts.test.js
AlexanderMakarov: `|` is regex alternation — `/revert|stash/i` matches either "revert" or "stash", because the template phrases the maneuver as "revert or stash the implicated hunks". The assertion just checks the template still names one of them.

PR #153, opened 2026-07-20
feat(hire): install Claude Code hooks from the awos-recruitment registry

AndreyNenashev:
## Summary

- `/awos:hire` now installs Claude Code hooks from the awos-recruitment registry alongside agents and skills.

> [!IMPORTANT]
> **Depends on [awos-recruitment#75](https://github.com/provectus/awos-recruitment/pull/75).** The `hook` CLI verb, the `type="hook"` registry search, and the first hook (`docs-that-work-gate`) all ship there. Do not merge this PR until #75 is merged and the `@provectusinc/awos-recruitment` CLI is version-bumped and published — otherwise `/awos:hire` instructs an install command users can't run yet.

## Changes

- Commit de15d4e: feat(hire): install Claude Code hooks from the awos-recruitment registry
- Commit abed81e: review follow-ups — hooks get a consent gate separate from the passive skills/MCPs/agents confirmation (dedicated `AskUserQuestion` naming the auto-running-shell-script nature); after the atomic CLI install, the real `HOOK.md` + entrypoint script are read back and summarized, with rollback offered on mismatch; lint assertion for Step 3 hook discovery pinned to unique text.

## Summary by CodeRabbit

- **New Features**
  - `/awos:hire` now discovers, searches for, and installs registry-provided hooks alongside other components.
  - Added separate consent for hooks, including clear executable-behavior warnings.
  - Added post-install verification, mismatch rollback options, and hook coverage reporting.

- **Documentation**
  - Clarified that hooks come exclusively from the registry and are never authored automatically.

- **Tests**
  - Added validation covering hook discovery, installation, consent, verification, and reporting.


commands/hire.md
```diff
@@ -87,15 +88,28 @@ Follow this process precisely.
...
+3.  If the `awos-recruitment` MCP server is not available or returns errors, tell the user it is unavailable and that you will proceed with generating agent files using general configuration. Note that they can prepare custom skills and agents in `.claude/skills/` and `.claude/agents/`. Skip to **Step 6**.
+4.  Gather all found skills, MCPs, agents, and hooks from the search results.
+5.  Show the user what was found and confirm installation before proceeding.
 
     | Role             | Found Skills                  | Found MCPs | Found Agents       |
     | ---------------- | ----------------------------- | ---------- | ------------------ |
     | `python-backend` | `fastapi-expert`              | —          | —                  |
     | `aws-infra`      | `terraform-pro`, `aws-deploy` | `aws-mcp`  | `aws-infra-expert` |
 
+    Found hooks get their own table with behavior-level detail — the user consents to what each hook does, not just its name:
```
AlexanderMakarov: This is the one place a user can meaningfully consent before an auto-running shell script lands in their settings — the CLI does no confirmation of its own, and the registry doesn't vet script contents (validation is metadata/layout + exec-bit only). Two adjustments worth making:

- **Separate the hook confirmation from the blanket one.** Hooks currently ride the same "confirm installation before proceeding" that also covers skills, MCPs, and agents — which are passive prompt text. A user could approve the whole batch without registering that they just accepted executable behavior on a lifecycle event. Ask for hooks specifically ("these install shell scripts that run automatically on the listed events — install them?"), distinct from the passive components.
- **Surface the actual hook, not just a metadata one-liner.** "What it runs" is authored in the hook's `HOOK.md` and the registry never checks it against the real `.sh`, so it can understate what the script does. Point the user at the actual script or its `HOOK.md` (the bundle ships it) before they confirm, so consent is against the real behavior.

tests/lint-prompts.test.js
```diff
@@ -1194,6 +1194,31 @@ test('hire.md QA Complement Rule is search-first and not tool-hardcoded', () =>
...
+  // registry ships, never fabricate hook commands, and must roster the
+  // post-install state. See docs: design spec 2026-07-17-hire-hooks.
+  const body = readUtf8(path.join(commandsDir, 'hire.md'));
+  assert.ok(
+    body.includes('npx @provectusinc/awos-recruitment hook ') &&
+      body.includes('bunx @provectusinc/awos-recruitment hook '),
+    'commands/hire.md must install hooks via the awos-recruitment `hook` verb in both npx and bunx forms'
+  );
+  assert.ok(
+    body.includes('.claude/settings.json'),
```
AlexanderMakarov: `body.includes('.claude/settings.json')` passes as long as that string appears *anywhere* in `hire.md` — and it now appears in three places: the Step 3 discovery bullet, Step 5's install line ("writes hook entries into … `.claude/settings.json`"), and the Step 8 roster rules. So this assertion stays green even if the Step 3 discovery bullet it's named for were deleted, which is exactly the contract its failure message claims to guard. Pin it to something unique to Step 3 — e.g. `settings.local.json` (which appears only in Step 3 and Step 8) together with the discovery-specific phrase `A missing or unparseable file means no existing hooks`, or match the whole discovery sentence.

AlexanderMakarov (commented):
The prompt change is clean in isolation. But hooks aren't just a fourth flavor of skill: this wires `/awos:hire` to install **auto-running shell scripts**, and that changes what the review has to weigh. Two things before this lands:

**1. It's coupled to [awos-recruitment#75](https://github.com/provectus/awos-recruitment/pull/75) — don't merge ahead of it.** The `npx/bunx @provectusinc/awos-recruitment hook …` verb, the `type="hook"` registry search, and the `docs-that-work-gate` hook all live in [awos-recruitment#75](https://github.com/provectus/awos-recruitment/pull/75), which is still under review with open issues, and the `hook` CLI verb isn't usable until that CLI is version-bumped and published. Merging #153 first ships a `/awos:hire` that instructs an install command users can't run yet. Please sequence #153 behind [awos-recruitment#75](https://github.com/provectus/awos-recruitment/pull/75)'s publish (and reference it in the PR).

**2. This command is the *only* consent gate for installing an executable.** On the registry side, hook validation is metadata/layout-only — it checks the frontmatter schema and the entrypoint's exec bit but never inspects what the `.sh` actually does; the sole safety barrier is maintainer PR review of the registry. And the CLI installs silently: no prompt, no dry-run — it drops an executable into `.claude/hooks/` and wires a `PreToolUse` entry into `.claude/settings.json` that then runs before every Bash call. So the "confirm installation" step in `/awos:hire` is the entire trust boundary a user has. The hook table (Event / What it runs) is the right instinct, but I'd strengthen it — see the inline on the consent step.

The rest is minor: a weak test assertion (inline), and the registry-unavailable fallback still points users to custom skills/agents but not hooks (small, and it already omits MCPs, so optional).

commands/hire.md
AndreyNenashev: Split into two gates in abed81e: skills/MCPs/agents keep the existing confirmation, and hooks get their own dedicated `AskUserQuestion` that names the executable nature ("install shell scripts that run automatically on the listed events"), plus an explicit statement that the "What it runs" column is registry metadata not vetted against the script.

On surfacing the actual script before confirming: not literally possible with today's CLI — the MCP search returns only `HOOK.md` frontmatter, and the `hook` verb's download + `.claude/hooks/` copy + settings injection are one atomic operation with no dry-run/preview. So consent is two-phase instead: the pre-install gate states the metadata caveat, and immediately after install the command reads back the installed `HOOK.md` + entrypoint script, summarizes what it actually does (commands run, files touched, what it can block), and offers keep/roll-back if the behavior exceeds what was advertised (rollback = delete the hook dir + remove the settings entry the CLI wrote). True pre-install inspection needs a preview/dry-run verb on the awos-recruitment CLI — worth raising on awos-recruitment#75.

tests/lint-prompts.test.js
AndreyNenashev: Fixed in abed81e — pinned to `settings.local.json` together with the discovery-unique phrase `A missing or unparseable file means no existing hooks` (the phrase appears only in the Step 3 bullet, so deleting the bullet now fails the test). Also added assertions for the two new contracts from the consent rework: the two-gate confirmation and the post-install HOOK.md/script read-back.

AndreyNenashev:
Addressed in abed81e:

1. **Sequencing** — the PR body now carries a "Depends on awos-recruitment#75" callout; this PR holds until [#75](https://github.com/provectus/awos-recruitment/pull/75) merges and the CLI is version-bumped and published.
2. **Consent gate** — split into two gates plus a post-install read-back of the actual script; details in the inline thread, including why true pre-install script inspection needs a preview verb on the awos-recruitment CLI (the `hook` verb installs atomically, no dry-run).
3. **Test assertion** — pinned to Step 3-unique text; inline thread.
4. **Fallback** — took the optional one too: the registry-unavailable path now also mentions hand-wiring hooks in `.claude/settings.json`.

commands/hire.md
```diff
@@ -178,6 +207,11 @@ Generated by `/awos:hire` on YYYY-MM-DD. Re-run `/awos:hire` to refresh — this
 | Name | Description | Skills |
 | ---- | ----------- | ------ |
 
+## Installed Hooks
+
+| Name | Event | Description |
+| ---- | ----- | ----------- |
```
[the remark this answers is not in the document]
AndreyNenashev: Adopted in 7e80c3d, with one widening: `Name` had the same provenance gap as `Description` (settings entries are anonymous — a name is only derivable when the command path matches the `.claude/hooks/<name>/` convention). The header is now `Name | Event | Command | Description` with explicit derivation rules: every cell is either taken from disk (settings entry, or the installed hook's `HOOK.md` frontmatter for `Description`) or the literal `—` — never summarized from the command string. Lint pins the exact header row.

tests/lint-prompts.test.js
```diff
@@ -1194,6 +1194,46 @@ test('hire.md QA Complement Rule is search-first and not tool-hardcoded', () =>
...
+    'commands/hire.md must state that hooks come from the registry only — hire never authors hook entries or commands'
+  );
+  assert.ok(
+    body.includes('two separate gates') &&
+      body.includes('shell script that runs automatically'),
+    'commands/hire.md Step 4 must gate hook consent separately from the passive skills/MCPs/agents confirmation, naming that hooks install auto-running shell scripts'
+  );
+  assert.ok(
+    body.includes('second half of the hook consent') &&
+      body.includes('.claude/hooks/<name>/HOOK.md'),
```
AlexanderMakarov, line 1232: This assertion's message promises the read-back offers *rollback on mismatch*, but the two substrings only prove the read-back sentence exists and mentions `HOOK.md`. The rollback path — delete `.claude/hooks/<name>/` and undo the settings entry when the script contradicts approved metadata (Step 5.4) — could be dropped from the prompt and this stays green, the same class of gap as last round. A third conjunct pins it: `body.includes('roll it back')` (unique to that paragraph); optionally `body.includes('entrypoint script')` for the "and entrypoint script" half.

tests/lint-prompts.test.js
```diff
@@ -1194,6 +1194,46 @@ test('hire.md QA Complement Rule is search-first and not tool-hardcoded', () =>
...
+    body.includes('.claude/settings.local.json') &&
+      body.includes('A missing or unparseable file means no existing hooks'),
+    'commands/hire.md Step 3 must discover existing hooks from the project settings files (.claude/settings.json / .claude/settings.local.json), treating a missing or unparseable file as no existing hooks'
+  );
+  assert.ok(
+    body.includes('## Installed Hooks'),
+    'commands/hire.md Step 8 coverage-report structure must contain the "## Installed Hooks" section'
+  );
+  assert.ok(
+    body.includes('| Name | Event | Command | Description |'),
```
AlexanderMakarov, line 1218: Same shape, lower stakes: the message says Name/Description must fall back to `—` rather than be invented, but the substring only pins the header row. The never-invent rule lives in the Step 8 rule block (`never inferred or invented`, and `Do not summarize the command string into a description`) and nothing asserts it — those rules could be gutted while the header row stays intact. If you want the message to match what's verified, add `body.includes('never inferred or invented')`.

commands/hire.md
AlexanderMakarov: The two-phase consent — a pre-install gate naming the metadata caveat, then a post-install read-back of the real script with keep/roll-back — resolves this well. Agreed that true pre-install inspection needs a preview/dry-run verb on the recruitment CLI. Thanks.

tests/lint-prompts.test.js
AlexanderMakarov: The pin to `settings.local.json` plus the discovery-unique phrase fixes the one I flagged. While re-reading, two of the sibling assertions you added in the same commit have the same message-overclaims-substring gap — left those inline.

AlexanderMakarov (commented):
Everything from the last round is addressed — thanks for the thorough turnaround; the two-phase hook consent (dedicated gate + post-install read-back) is a solid resolution.

Two small things on the new lint assertions (inline) — both the same shape as the gap from last round: the failure message names a contract the substring doesn't actually pin, so the clause it's guarding could be deleted and the test would stay green. Non-blocking.

The sequencing concern from last round is cleared: [awos-recruitment#75](https://github.com/provectus/awos-recruitment/pull/75) — the `hook` verb, the `type="hook"` search, and `docs-that-work-gate` — merged 2026-07-21. Since the recruitment CLI runs from the repo directly, that puts the `hook` verb `/awos:hire` calls in place, so this is no longer waiting on the dependency.

**Interaction with #146 (awos-containment guard).** #146 ships a PreToolUse guard whose `PROTECTED_INTREE` includes `.claude/settings.json` precisely because a hook/plugin registration there is "durable execution." On a project running that guard, `/awos:hire`'s hook install writes `settings.json` via the recruitment CLI's `npx` subprocess, which the guard can't observe — so it slips through — but a *tool*-mediated touch would be blocked: the Step 5.4 rollback that "removes the settings entry the CLI just added," if done via `Edit`, hits the guard and can't complete (there's no CLI uninstall verb). Worth deciding the interaction if both land — e.g. `/awos:hire` detecting containment and offering the `AWOS_CONTAINMENT_OFF`/whitelist path, or an uninstall verb so rollback doesn't need a tool write. Flagged the same on #146.

AlexanderMakarov:
Two things are still open, neither blocking and both already on their threads — flagging only so they don't get lost, not re-raising them:

- The two lint assertions I noted last round still overclaim: the failure message names a contract the substring doesn't pin, so the guarded clause could be deleted with the test staying green. Cosmetic, and your call whether it's worth a follow-up.
- The interaction with #146 (the awos-containment guard) is the one thing I'd genuinely settle before *both* land — and it's now the live topic of the Slack thread, so it's being worked in the right place. Nothing to change in this PR for it; just don't merge the two into a state where a contained project can install a hook via the CLI subprocess but can't roll it back through a tool write. That's a sequencing/coordination call between the two PRs, not a fix here.

I guess #146 is not a blocker anymore.

AlexanderMakarov (commented):
Two things are still open, neither blocking and both already on their threads — flagging only so they don't get lost, not re-raising them:

- The two lint assertions I noted last round still overclaim: the failure message names a contract the substring doesn't pin, so the guarded clause could be deleted with the test staying green. Cosmetic, and your call whether it's worth a follow-up.
- The interaction with #146 (the awos-containment guard) is the one thing I'd genuinely settle before *both* land — and it's now the live topic of the Slack thread, so it's being worked in the right place. Nothing to change in this PR for it; just don't merge the two into a state where a contained project can install a hook via the CLI subprocess but can't roll it back through a tool write. That's a sequencing/coordination call between the two PRs, not a fix here.

I guess #146 is not a blocker anymore.

AlexanderMakarov (approved):

PR #155, opened 2026-07-22
feat(flow): pre-seed the spec interview via the flow's fetch stage

dustyo-O:
## What & why

Roadmap item **"Pre-seed the spec interview via the flow's fetch stage (no new questions in `/awos:spec`)"**.

Cuts `/awos:spec`'s cold interview from 5–10 questions to 2–4 by pre-seeding context — **without** adding an "any source material?" question to `spec.md` (that ask is net-negative when there are no docs). The asking happens **once at `/awos:flow` generation time**; the fetching happens **per run** in the generated command.

## Changes

- **`plugins/awos/commands/flow.md`** — §1 interview now captures *where the ticket's surrounding context lives* (epic/parent ticket, tracker remote links & attachments, docs connector, meeting-notes/chat source) and which §7 transport reaches each. Generator version `2.4.2 → 2.4.3`.
- **`plugins/awos/templates/delivery-flow-template.md`** — new §1 field **"Surrounding context to pre-seed the spec"** so the decision is recorded for the generated fetch stage to consume.
- **`plugins/awos/templates/implement-feature-template.md`** — fetch stage (Step 1) now pulls the ticket's epic/parent description, remote links, and attachments, and **lists unreachable links** instead of silently skipping — mirroring the existing `fix-bug-template.md` remote-links pull. Step 4 already forwards the normalized bundle to `/awos:spec`, so the richer material flows through for free.
- **`commands/spec.md`** — Step 2 silently **consumes ticket IDs / URLs / file paths already in `<user_prompt>`** before interviewing, folding them into the known-info extraction. No new `AskUserQuestion`. Covers both the flow handoff and a user pasting a link into a direct `/awos:spec` run.
- **`tests/lint-prompts.test.js`** — new Layer-1 test locking all three edits (flow §1 capture, per-run fetch, spec consume-rule) plus the no-new-question guard. `EXPECTED_PLUGIN_VERSION → 2.4.3`.
- **`plugin.json` / `marketplace.json`** — plugin version `2.4.2 → 2.4.3` (behavior changed → generated commands regenerate on the next `/awos:flow` re-run).

## Verification

- `npm run test:lint` — **146 pass, 0 fail** (includes the new test).
- `npx prettier@3.8.4 . --check` — clean (repo-pinned version).

Prompt-only change; no engine rebuild needed.

## Summary by CodeRabbit

* **New Features**
  * Feature workflows now pre-seed generated specs with surrounding ticket context, including parent/epic framing, linked issues, attachments, documentation references, and conversation sources.
  * Linked-but-unreachable references are now carried forward for later review rather than being skipped.
  * Spec interview steps now automatically consider source references already provided in the original prompt.

* **Documentation**
  * Updated workflow templates and command guidance to reflect the expanded context capture and handling.

* **Chores**
  * Bumped the AWOS plugin version to **2.4.3**.


tests/lint-prompts.test.js
```diff
@@ -956,6 +956,74 @@ test('fix-bug template reads remote links, sweeps all surfaces, and verifies sub
...
+  // per run (mirroring fix-bug's remote-links pull); and spec.md silently
+  // consumes any references already in its prompt. Lock all three so a
+  // future edit can't quietly drop the pre-seed and reintroduce the cold
+  // interview, or "fix" it by adding the forbidden question.
+
+  // flow.md §1 must ask, once at generation time, where the ticket's
+  // surrounding context lives and which §7 transport reaches each.
+  const flow = readUtf8(path.join(pluginCommandsDir, 'flow.md'));
+  assert.ok(
+    /where the ticket's surrounding context lives/i.test(flow),
```
AlexanderMakarov: Two of the new assertions pin a section header but their messages promise the source list behind it:

1. This one (`/where the ticket's surrounding context lives/i`) — the message says it captures "epic/parent, remote links, docs connector, meeting notes", but only the header sentence is matched. `epic/parent` and `remote links` are re-checked by the implement-feature assertion below, but **docs connector** and **meeting notes** are pinned nowhere — delete the four-source enumeration behind the header and this stays green. AND a second substring from the list, e.g. `/docs connector|Confluence\/Notion/i` (and/or `/meeting-notes/i`).
2. Same shape at line 991 (`/Surrounding context to pre-seed the spec/i`): the message claims "each with its §7 transport", but only the field header is matched — the transport requirement could be dropped with the test still green. A `/§7 transport/i` conjunct closes it.

Both low-severity — the field text and the implement-feature fetch anchor back the behavior — but since the point of this test is to lock the pre-seed against a quiet regression, worth making the anchors match what the messages assert.

tests/lint-prompts.test.js
```diff
@@ -956,6 +956,74 @@ test('fix-bug template reads remote links, sweeps all surfaces, and verifies sub
...
+  const spec = readUtf8(path.join(commandsDir, 'spec.md'));
+  assert.ok(
+    /Consume source material already in the prompt/i.test(spec) &&
+      /ticket IDs, URLs, or file paths/i.test(spec),
+    'commands/spec.md Step 2 must consume ticket IDs / URLs / file paths already in <user_prompt> before the interview, folding them into the known-info extraction'
+  );
+  assert.ok(
+    /without a new question/i.test(spec) && /net-negative/i.test(spec),
+    'commands/spec.md must pre-seed WITHOUT adding an "any source material?" question — the ask is net-negative when the prompt carries no references'
+  );
```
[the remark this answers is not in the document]
AlexanderMakarov: Agreed — the guard checks the prohibition sentence is *present*, not that the forbidden question is *absent*, so a reintroduced "any source material?" ask stays green as long as the prohibition prose also survives; the message asserts a behavior the regex doesn't test.

One trap in the fix: a naive `assert.ok(!/any source material/i.test(spec))` fails the *current* file, because the prohibition quotes the phrase verbatim (`Do **not** add an "any source material?" question`). A count-based pin avoids that — assert the phrase occurs exactly once and that the one occurrence is the prohibition:

```js
assert.equal((spec.match(/any source material/gi) || []).length, 1,
  'spec.md must name "any source material" only in the prohibition, never as an added ask');
assert.ok(/Do \*\*not\*\* add an "any source material\?" question/.test(spec),
  'the single mention must be the prohibition');
```

Worth stating plainly that a substring linter can only catch a *literal* reintroduced ask, not prove no runtime question is asked — so this tightens the gap rather than closing it, which is a reason to add the negative, not to treat the guard as complete.

AlexanderMakarov (commented):
The change is clean and reads merge-ready. The 4 coupled edits line up into a consistent capture → record → fetch → consume contract: flow.md §1 captures where the surrounding context lives (once, at generation time), the delivery-flow template records it, implement-feature's fetch stage pulls it per run and lists unreachable links, and spec.md consumes prompt references with no new question. The double-fetch case is handled explicitly (spec.md reads the flow's bundle rather than re-fetching), the "mirrors fix-bug's remote-links pull" claim checks out.

The only thing worth a look is test strength — the new lint lock is good, but a couple of its anchors verify a heading or a prohibition sentence rather than the contract their failure message names, so a future edit could regress the guarded behavior with the test still green. CodeRabbit already flagged the main one (the no-new-question guard); I've added the fix detail on that thread and one sibling instance inline. All non-blocking — this is good to go once you've weighed them.

tests/lint-prompts.test.js
dustyo-O: Fixed in 1efb587. Both anchors now match what their messages assert:
- flow.md §1 anchor also pins `/documentation connector|Confluence\/Notion/i` and `/meeting-notes/i` — the two sources that weren't re-checked elsewhere (epic/parent + remote links stay covered by the implement-feature anchor).
- delivery-flow-template §1 anchor now also requires `/each with its §7 transport/i`.

tests/lint-prompts.test.js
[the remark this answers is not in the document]
dustyo-O: Applied your count-based pin verbatim in 1efb587 — assert `any source material` occurs exactly once and that the one occurrence is the prohibition. Kept the note that a substring linter only catches a literal reintroduced ask, not a proof that no runtime question is asked, so this tightens the gap rather than closing it. Thanks for the trap callout — a naive negative check would indeed have failed the current file.

tests/lint-prompts.test.js
```diff
@@ -956,6 +956,95 @@ test('fix-bug template reads remote links, sweeps all surfaces, and verifies sub
...
+
+  // The decision record must carry the field so the generated fetch stage
+  // has a recorded source list to pull from.
+  const dfTemplate = readUtf8(
+    path.join(pluginTemplatesDir, 'delivery-flow-template.md')
+  );
+  assert.ok(
+    /Surrounding context to pre-seed the spec/i.test(dfTemplate) &&
+      /each with its §7 transport/i.test(dfTemplate),
+    'delivery-flow-template.md §1 must record the surrounding-context sources AND that each carries its §7 transport, the fetch stage pulls to pre-seed the spec interview'
```
AlexanderMakarov: Pulling the parenthetical out into its own conjunct left the sentence stranded — `…sources AND that each carries its §7 transport, the fetch stage pulls to pre-seed the spec interview` doesn't parse, and this string is exactly what someone stares at when the anchor trips. Maybe:

```js
'delivery-flow-template.md §1 must record the surrounding-context sources the fetch stage pulls to pre-seed the spec interview, each with its §7 transport';
```

Same contract, both halves still named.

AlexanderMakarov (commented):
Everything from the last round is addressed — thanks for the quick turnaround, and for keeping the note that a substring pin catches a literal reintroduced ask rather than proving no runtime question is asked. That framing is the honest one.

I checked the three new anchors against the files they pin rather than trusting the green run: `documentation connector (Confluence/Notion …)` and `meeting-notes` are both in `plugins/awos/commands/flow.md:95`, `each with its §7 transport` is in `plugins/awos/templates/delivery-flow-template.md:32`, and `commands/spec.md` carries the phrase exactly once, in the prohibition at line 89. So none of the new pins is vacuously green, and the count-plus-prohibition pair does close the hole it claims to — the prohibition itself contributes the one occurrence, so a reintroduced ask always pushes the count to two.

One leftover from the edit: the `delivery-flow-template` assertion message lost its grammar (inline). Not blocking.

tests/lint-prompts.test.js
dustyo-O: Fixed in 29fcb9c — reworded to your suggestion: `delivery-flow-template.md §1 must record the surrounding-context sources the fetch stage pulls to pre-seed the spec interview, each with its §7 transport`. Both halves named, and the sentence parses. Thanks for catching it — that string is exactly what someone stares at when the anchor trips.

AlexanderMakarov (approved):

PR #161, opened 2026-07-24
fix(audit): ARCH-05 accepts digit-led filename tokens (migrations, 2fa_auth)

AndreyNenashev:
## What happened

Running the audit on a standard Python service produced an ARCH-05 **FAIL** (0 of 3 pts) for "consistent file naming conventions", even though the project's naming is fully consistent. The only 3 files flagged were Alembic database migrations that use the near-universal `NNN_description` ordering convention.

From `context/audits/2026-07-24_13-25-00/code-architecture.json`:

```json
{
  "check_id": "ARCH-05",
  "status": "FAIL",
  "value": 0.9769230769230769,
  "evidence": [
    "inconsistent file naming: 3 file(s) depart from the dominant snake_case convention (all-or-nothing — AWOS's own standard)",
    "dominant convention: snake_case (127/130 = 98% compatible)",
    "  snake_case: 127 compatible file(s)",
    "  kebab-case: 77 compatible file(s)",
    "  camelCase: 77 compatible file(s)"
  ],
  "weight_awarded": 0, "weight_max": 3
}
```

The 3 "departing" files:

```
migrations/versions/001_create_document_metadata_table.py
migrations/versions/002_create_api_keys_table.py
migrations/versions/003_create_document_scope_links_table.py
```

## Root cause

`compatibleConventions(name)` in `detectors/code_architecture.ts` classifies a file's name-stem, and every snake/kebab/camel pattern was anchored on `^[a-z]` — the stem had to start with a lowercase **letter**:

```ts
if (/^[a-z][a-z0-9]*$/.test(name)) return ['snake_case','kebab-case','camelCase'];
if (/^[a-z][a-z0-9]*(_[a-z0-9]+)+$/.test(name)) return ['snake_case'];
if (/^[a-z][a-z0-9]*(-[a-z0-9]+)+$/.test(name)) return ['kebab-case'];
if (/^[A-Z][A-Za-z0-9]*$/.test(name)) return ['PascalCase'];
if (/^[a-z][A-Za-z0-9]*[A-Z][A-Za-z0-9]*$/.test(name)) return ['camelCase'];
return []; // 'other'
```

`001_create_document_metadata_table` starts with a digit, matches nothing, and is bucketed as `'other'` → not counted toward the dominant `snake_case`. Because ARCH-05 is all-or-nothing (`dominantCount === total` required for PASS), those 3 files fail the whole check despite 98% compatibility.

## Why it's wrong

- `NNN_snake_case` is the standard Alembic/Django/Flyway migration convention — the numeric prefix encodes **apply order**, not a naming style. The remainder (`create_document_metadata_table`) is textbook snake_case.
- These files are never imported by module name; frameworks load them by file path (`importlib`) and resolve ordering from internal `revision`/`down_revision` ids, so a digit-leading filename is correct and intentional.
- The detector already exempts other legitimate "different axis" names (test files, and stems `index`/`__init__`/`main`/`app`/`setup`/`config`) but not numeric ordering prefixes — so it penalized a convention more standardized than most it accepts.
- The same `^[a-z]` anchoring also miscounted legitimate **digit-led descriptive tokens**: `2fa_auth` (two-factor auth), `s3_bucket`, `oauth2_client`, `utf8_decoder`, `md5_hash`, `i18n_utils` — all valid snake_case where a token simply starts with a digit.
- Net effect: any repo with DB migrations, numbered scripts, or digit-led tokens got a guaranteed false FAIL on a maximal-reliability, 3-pt check.

## How it's resolved

Fixed at the root — the classifier — rather than special-casing migration paths:

- **Tokens may be lowercase-alphanumeric and may start with a digit.** The separator-delimited (`snake_case`, `kebab-case`) and single-token classes now accept `[a-z0-9]+` tokens, so `001_create_users`, `2fa_auth`, `s3_bucket`, and `01-intro` all classify by their actual separator/casing shape.
- **`camelCase`/`PascalCase` stay letter-led** — a case-based convention is letter-led by definition, so those patterns are unchanged.
- This is a general fix: it covers migration ordering prefixes, numbered scripts, and digit-led descriptive tokens in one rule, with no path heuristics.
- **Genuine convention clashes still FAIL** — mixed casing/separators (e.g. a lone `myController.py` among snake files) remain violations.

Also corrected a stale header comment that still described the retired graded `90%/70%` curve; ARCH-05 has been all-or-nothing since the 2026-07-06 standards refresh (`PASS` at 100% compatible, `FAIL` below, no WARN band). No scoring semantics, weight, or `standards.toml` changes.

## Tests

Added ARCH-05 regression cases in `detectors/code_architecture_arch05.test.ts`:

1. Numbered migration files (`00N_*.py`) among snake sources → **PASS**
2. Digit-led descriptive tokens (`2fa_auth`, `s3_bucket`, `oauth2_client`) → **PASS**
3. A genuine camelCase deviation alongside a numbered file → still **FAIL** (the relaxation doesn't mask real inconsistency)

Full engine suite: **1324 pass / 0 fail**. Regenerated the committed `dist/` engine bundle (`npm run build:audit-engine`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated naming-convention scoring so the dominant convention now only passes at full compatibility.
  * Improved detection for filenames with numeric prefixes and digit-led snake_case tokens to avoid incorrect violations.
  * Continued to report failures when mixed naming conventions are present.

* **Tests**
  * Expanded ARCH-05 coverage for numeric ordering prefixes, digit-led tokens, and mixed naming violations.


plugins/awos/skills/ai-readiness-audit/detectors/code_architecture_arch05.test.ts
```diff
@@ -144,6 +144,77 @@ test('ARCH-05 classifies dotted role qualifiers (.d.ts, .stories.tsx) by their s
   }
 });
 
+test('ARCH-05 treats a leading numeric ordering prefix as sort order, not a naming style (numbered files PASS)', () => {
```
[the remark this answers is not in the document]
AndreyNenashev: Verified against the current file — this is a false positive, so no change.

`prettier --check` passes cleanly on this file (exit 0), and piping it through `prettier` returns **byte-identical** output, so it's already in Prettier's canonical form. Prettier's `printWidth` (80) is a *soft* wrap target — it does not break string literals. A `test('<long descriptive name>', () => {` line therefore keeps its long string argument on one line **by design**; that's exactly what Prettier emits.

Manually wrapping these three declarations would actually *introduce* a violation: `prettier --check` would then report the file as needing formatting and the CI format gate (`prettier . --check`) would fail. The gate is green as-is, so there's nothing to fix here.

plugins/awos/skills/ai-readiness-audit/detectors/code_architecture.ts
```diff
@@ -451,19 +455,20 @@ type NamingConvention =
...
 
 /**
- * Conventions a filename stem is compatible with. Single-token lowercase
- * names (`utils`, `api`) carry no separator evidence, so they are compatible
- * with EVERY lowercase convention (snake, kebab, camel) rather than being
- * pinned to one — otherwise they would skew dominance away from the
- * convention the multi-word names actually follow. Returns an empty array
- * for names that fit no recognised convention ('other').
+ * Conventions a filename stem is compatible with. Tokens are lowercase-
+ * alphanumeric and may start with a digit (`2fa`, `s3`, `oauth2`, `001`).
```
AlexanderMakarov: The relaxation only affects stems whose *first character* is a digit. In this example list, `s3` and `oauth2` start with a letter — they already matched the old `^[a-z][a-z0-9]*` anchor (letter, then digits/letters allowed), so they classify the same before and after. Only `2fa` and `001` actually demonstrate what changed. Worth trimming the parenthetical to genuinely digit-first examples so the docstring shows the real behavior.

plugins/awos/skills/ai-readiness-audit/detectors/code_architecture_arch05.test.ts
```diff
@@ -144,6 +144,77 @@ test('ARCH-05 classifies dotted role qualifiers (.d.ts, .stories.tsx) by their s
...
+      res.status,
+      'PASS',
+      `numbered files share the snake_case convention once the ordering prefix is stripped; got ${res.status}: ${JSON.stringify(res.evidence)}`
+    );
+  } finally {
+    rmSync(repo, { recursive: true, force: true });
+  }
+});
+
+test('ARCH-05 accepts digit-led tokens as valid snake_case (2fa_auth, s3_bucket, oauth2_client)', () => {
```
AlexanderMakarov: Same conflation here: the test is titled "digit-led tokens … (2fa_auth, s3_bucket, oauth2_client)" and the comment just below (175–178) calls `s3`/`oauth2` digit-led — but only `2fa_auth` is. `s3_bucket` and `oauth2_client` start with a letter and already passed under the old regex, so two of the three cases assert pre-existing behavior rather than the new branch. Fine as regression guards, but the naming implies they were fixed here when they weren't — either relabel, or swap them for genuinely digit-first names (`001_foo`, another `N…_…`) so the test actually exercises the relaxation.

plugins/awos/skills/ai-readiness-audit/detectors/code_architecture_arch05.test.ts
```diff
@@ -144,6 +144,77 @@ test('ARCH-05 classifies dotted role qualifiers (.d.ts, .stories.tsx) by their s
...
+      'app/auth_handler.py': 'x = 1\n',
+      'app/data_model.py': 'x = 1\n',
+      'migrations/versions/001_create_users.py': 'x = 1\n',
+      'migrations/versions/002_add_index.py': 'x = 1\n',
+    });
+    const res = detect(repo);
+    assert.equal(
+      res.status,
+      'PASS',
+      `numbered files share the snake_case convention once the ordering prefix is stripped; got ${res.status}: ${JSON.stringify(res.evidence)}`
```
AlexanderMakarov: Small one: the assert message says the numbered files pass "once the ordering prefix is stripped," but nothing is stripped — `001_create_users` matches the snake regex whole now that `001` is a valid leading token. Could send a maintainer looking for stripping logic that doesn't exist. "…because a digit-led token is valid snake_case" matches what actually happens.

AlexanderMakarov (commented):
Solid catch and fix — relaxes the ARCH-05 classifier so a stem whose *first character* is a digit (Alembic `001_…` migrations, `2fa_auth`) is no longer bucketed as `'other'` and forced to FAIL. The logic holds up: separators are still required, evaluation order is preserved, and the rebuilt `dist/cli.js` matches the source (clean rebuild, no stale bundle). Since the change only moves names out of `'other'`, it can relax a FAIL to PASS but never introduce a new FAIL — nicely contained. Keeping camel/Pascal letter-led is the right call, too: a digit-led filename isn't even valid in most languages that use camelCase file names (a Java class file can't start with a digit), so there's nothing to gain by relaxing those.

One accuracy point worth fixing before merge (inline): the "digit-led" framing overreaches. Of the examples in the docstring and the test, only `2fa`/`001` (and `2fa_auth`) actually start with a digit — `s3`, `oauth2`, `s3_bucket`, `oauth2_client` all start with a *letter* and already matched the old `^[a-z][a-z0-9]*` anchor, so they classified identically before and after this change. Presenting them as names this change fixes is misleading in shipped docstrings and tests. Plus one tiny wording nit in a test message.

Bigger picture, not for this PR: this whole check is language-agnostic — it derives one dominant convention across the entire repo. Conventions are really per-language (Java class files are PascalCase, Python is snake_case), so a polyglot repo that's internally consistent *within* each language can still FAIL here, and edge cases like digit-led names are symptoms of that same flat model. If ARCH-05 keeps throwing false FAILs, a language-aware detector — expected convention per language / file group — would resolve the class of problem at the root rather than one regex at a time. Out of scope for this PR; the narrow fix here is the right call for the immediate bug. Just flagging the direction.

CodeRabbit's Prettier thread doesn't hold — replied there.

plugins/awos/skills/ai-readiness-audit/detectors/code_architecture.ts
AndreyNenashev: Fixed in 87318b8 — trimmed the parenthetical to the genuinely digit-first examples (`2fa`, `001`). `s3` still appears in the single-token sentence below, where it illustrates an alphanumeric token rather than digit-led support. The section header comment above had the same overreach, so its examples now attach to the clauses they actually demonstrate: non-leading digits (`s3_bucket`, `oauth2_client`) vs leading (`2fa_auth`, `001_create_users`).

plugins/awos/skills/ai-readiness-audit/detectors/code_architecture_arch05.test.ts
AndreyNenashev: Went with relabel rather than swap, in 87318b8: the title now reads "digit-led (2fa_auth) and letter-led (s3_bucket, oauth2_client)", and the comment states outright that the letter-led pair classified as snake_case before the relaxation and stays as regression guards for non-leading digits — the regexes were rewritten, so pinning that shape down has value. The relaxation itself is exercised by `2fa_auth` here plus `001_`/`002_` in the ordering-prefix test. The assert message had the same conflation and is de-conflated too.

plugins/awos/skills/ai-readiness-audit/detectors/code_architecture_arch05.test.ts
AndreyNenashev: Fixed in 87318b8 with your wording — the message now reads "numbered files pass because a digit-led token is valid snake_case". The same phantom mechanism appeared in two sibling comments (test 1's "the stem after the prefix", test 3's "Stripping the ordering prefix must not mask…"); both reworded, since nothing in the detector strips anything.

AlexanderMakarov (approved):
Everything from the last round is addressed — thanks for the quick turnaround, including the sibling wording spots I did not call out individually. The digit-led vs letter-led distinction reads accurately now across the docstring and tests. Nothing new on my end; the language-aware-detector direction I raised stays a non-blocking future note, not for this PR. LGTM.

PR #166, opened 2026-07-29
docs: document the brownfield integration path

dustyo-O:
## What

Documents the **brownfield** (existing-codebase) adoption path to parity with the greenfield Quick Start. The brownfield *behaviour* was shipped across prior PRs (detection in `/awos:product`, capability/tech-stack passes in `/awos:roadmap` and `/awos:architecture`, external-source gathering) but lived only in the command prompts and `CLAUDE.md` — absent from every user-facing doc. This closes that gap.

Roadmap item: **Documented brownfield integration path** (Brownfield, High).

## Contents

- **`docs/brownfield-adoption.md`** *(new)* — end-to-end guide: the AI-readiness audit as Step 0, the brownfield-aware `product → roadmap → architecture` flow, the `brownfield.md` / `context/sources/` artifact lifecycle, opt-in/opt-out, and a greenfield-vs-brownfield comparison table.
- **`README.md`** — the "running on an existing codebase?" callout now links to the guide.
- **`docs/commands/product.md` · `roadmap.md` · `architecture.md`** — a "Brownfield mode" section in each, cross-linking the guide.
- **`plugins/awos/README.md`** — positions `/awos:ai-readiness-audit` as the recommended first step of brownfield adoption, noting its scoring is self-contained (independent of `/awos:hire`).

## Scope

Documentation only — no behaviour or engine changes, so no new test contracts (the lint layer already guards the command prompts these docs describe). Lint: **146 pass / 0 fail**; Prettier clean; internal links resolve.

## Summary by CodeRabbit

* **Documentation**
  * Added a Brownfield Adoption Guide for using AWOS with existing codebases.
  * Documented brownfield behavior for product discovery, roadmap creation, and architecture planning.
  * Clarified capability inventory, technology discovery, external documentation imports, cleanup, and opt-in/opt-out workflows.
  * Updated Quick Start and installation guidance with the recommended readiness audit and adoption sequence.

* **Enhancements**
  * Improved roadmap planning to distinguish completed capabilities from partial, scaffolded, and planned work.
  * Enabled architecture proposals to be reviewed after they are saved.


dustyo-O:
@coderabbitai review

dustyo-O:
Addressed all 5 findings in `4cb0c7d`.

- **Audit baseline (Major)** — reworded so the *methodology* is what's self-contained (engine, independent of `/awos:hire`'s agents); the *score* reflects repo state and is expected to rise with adoption. Removes the self-contradictory "scores the same" phrasing.
- **Capabilities completion status (Major)** — only fully-implemented capabilities are treated as done; partial/scaffolded and planned (TODO/FIXME) work stays visible as upcoming roadmap items. Applied in `brownfield-adoption.md` and `docs/commands/roadmap.md`.
- **Secrets (Major)** — retain only non-secret source metadata under `context/`; credentials belong in env vars / a secret manager. Applied in `brownfield-adoption.md` and `docs/commands/architecture.md`.
- **Auto-detection claim (Minor)** — narrowed: only `/awos:product` detects; the later commands build on its staged `brownfield.md`.
- **Fence language (Minor)** — tagged the audit command block as `text` (the new file's only fence).

Lint 146 / 0, Prettier clean.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+If the plugin is not installed, the command tells you so and continues without external sources — the code-based product definition is still produced.
+
+## Step 2 — `/awos:roadmap`: start from what already exists
+
+When `context/product/brownfield.md` is present, `/awos:roadmap` runs a focused `Explore` pass to inventory **existing capabilities** — features that are fully implemented (routes, UI, tests), partially built or scaffolded, and planned (TODOs, FIXMEs). It appends these under a `## Capabilities` heading in `brownfield.md`.
+
+The roadmap is then anchored in reality: only **fully-implemented** capabilities are treated as done, while **partially-built or scaffolded** work and **planned** items (TODOs, FIXMEs) stay visible as upcoming roadmap items rather than being marked complete. New phases describe what comes next instead of re-planning work the codebase has already finished. As with product, the capability findings are triaged with you before the roadmap is finalized.
```
AlexanderMakarov: This distinction isn't in the prompt. [`commands/roadmap.md:84`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/roadmap.md#L83-L85) is the only line that says what the roadmap does with the findings, and it folds all of them into done:

> `Use the full set of capabilities (from brownfield.md) to anchor the roadmap: existing capabilities are noted as already done, and new phases focus on what comes next.`

The three-way split does exist, but only in the Explore agent's *reporting* instructions at [`commands/roadmap.md:68-70`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/roadmap.md#L67-L71) — nothing downstream acts on the difference. Your own side-by-side table at `:113` ("Existing capabilities marked done") describes the actual behavior, so the doc disagrees with itself here too. Same sentence is in `docs/commands/roadmap.md:29`.

This came from CodeRabbit's capability-status finding, which asked the *docs* to assert the distinction — so the fix ended up documenting something the framework doesn't do. Given that, I'd rather see it go the other way: add the rule to `commands/roadmap.md:84` (one line — "only capabilities reported as fully implemented are noted as done; partial/scaffolded and TODO/FIXME items become upcoming roadmap items") and keep the prose you've already written. It's the better product behavior and it's a smaller edit than unpicking the claim from three places. If you'd rather keep this PR docs-only, the alternative is reverting `:79`/`:113`/`roadmap.md:29` to match the prompt and filing the prompt change separately.

While you're in this sentence: "triaged with you before the roadmap is finalized" is backwards — [`commands/roadmap.md:82`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/roadmap.md#L81-L83) puts triage in Step 3: Finalization "after the roadmap is saved — so exploration never blocks the write", and [`:129-130`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/roadmap.md#L128-L131) writes, then triages, then re-saves. You describe that ordering correctly for product at `:56`; worth matching it.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+| `context/product/brownfield.md` | `/awos:product`             | `/awos:roadmap`, `/awos:architecture` | Deleted by `/awos:architecture` after absorption.                                 |
+| `context/sources/sources.md`    | `/awos:product` (via skill) | `/awos:product`, retrieval pass       | Removed by `/awos:architecture` once absorbed; kept + referenced if still useful. |
+| `context/audits/<timestamp>/`   | `/awos:ai-readiness-audit`  | You (review + recommendations)        | Permanent history; each run is an independent snapshot.                           |
+
+## Opting in and opting out
+
+- **Force brownfield:** include intent in your prompt — `/awos:product explore the existing codebase`.
+- **Force greenfield on a repo with code:** `/awos:product start from scratch, ignore existing code`.
+- **Skip external sources:** answer **No** when asked, or decline; a `## Status: none` marker records the choice so you are not asked again.
+- **Unattended runs** (`claude -p …`): every brownfield prompt has a safe default (explore, no external sources) and never blocks the deliverable, so foundation setup completes without a human present.
```
AlexanderMakarov: "never blocks the deliverable, so foundation setup completes without a human present" is the sentence someone will rely on to put `/awos:product` in CI, and the prompt documents two ways it doesn't hold.

[`commands/product.md:98`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L97-L99) and [`:99`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L98-L100) both end with `If the skill triggers an editor restart, stop here.` — that stop is inside Step 2B, and the write is Step 3 ([`commands/product.md:139`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L138-L140)), so on the restart path no `product-definition.md` gets written at all. And the prompt authors flagged the general case themselves, at [`commands/product.md:42`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L41-L43):

> `In claude -p a dismissed AskUserQuestion ends the turn, so a deliverable Write placed after such a question never runs unattended. The fix is structural — keep the Write ahead of any dismissable question, then refine afterward.`

Both Step 2B questions ([`:73`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L72-L74), `:99`) sit ahead of the Step 3 write, so that note applies here. Worth knowing the platform side reinforces it: the [headless docs](https://code.claude.com/docs/en/headless) note that under `--permission-mode dontAsk` — the locked-down mode for CI — "`AskUserQuestion` … [is] denied even when an allow rule matches", so in exactly the setup this bullet is written for, the question doesn't get a default, it gets refused.

The first half of your sentence is right — `:73` does default to Yes-explore and `:99` to No when unanswered. I'd keep that and replace the guarantee with what's actually true: an *unanswered* question falls back to its default and the run continues; a *dismissed or denied* one ends the turn; and external-source setup that adds an MCP server stops for a restart before the definition is written, so you re-run to finish.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+Brownfield projects usually have knowledge that lives outside the code — a Confluence or Notion wiki, Jira or Linear tickets, Slack history, email threads. When `/awos:product` creates `brownfield.md`, it also offers to import that knowledge:
+
+> Do you have external documentation (wikis, tickets, chats, email) you'd like to import into the project context?
+
+If you opt in, the command invokes the `awos:configure-external-sources` skill (part of the AWOS plugin), which:
+
+- identifies each source and its category (documentation, tickets, communication),
+- guides you through connecting the right tool — an MCP server, a CLI like `gh`, or manual paste,
+- records the configuration in `context/sources/sources.md`, and
+- once connected, retrieves product-relevant content (requirements, goals, audience, pain points) and folds it into the same triage flow as the code findings.
```
AlexanderMakarov: This bullet is under "the command invokes the skill … which:", but retrieval is the one thing the skill says it doesn't do — [`configure-external-sources/SKILL.md:13`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L12-L14):

> `Your only output is the structured source manifest at context/sources/sources.md. You do not retrieve content from the sources — the calling command handles retrieval after you finish.`

`/awos:product` does it itself, in its own substep — [`commands/product.md:101`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L100-L102), one Explore agent per `Access: mcp` / `Access: cli` source, `AskUserQuestion` paste for `Access: manual`. Worth splitting: end the skill's bullets at "records the configuration in `context/sources/sources.md`", then a separate sentence for the retrieval that `/awos:product` runs afterwards. `docs/commands/product.md:30` has the same misattribution.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+The roadmap is then anchored in reality: only **fully-implemented** capabilities are treated as done, while **partially-built or scaffolded** work and **planned** items (TODOs, FIXMEs) stay visible as upcoming roadmap items rather than being marked complete. New phases describe what comes next instead of re-planning work the codebase has already finished. As with product, the capability findings are triaged with you before the roadmap is finalized.
+
+## Step 3 — `/awos:architecture`: adopt the existing stack, then clean up
+
+With `brownfield.md` present, `/awos:architecture` runs an `Explore` pass to discover the **existing technology stack** — languages, frameworks and versions, databases and ORMs, infrastructure (Docker, cloud configs, deployment scripts), external services, and testing/build/CI tooling. It appends these under a `## Technology` heading, and uses them as the **defaults** for each architecture decision instead of proposing technologies from scratch. You still confirm or change every decision.
```
AlexanderMakarov: Two overstatements in this one sentence. "You still confirm or change every decision" — the file is written first, without waiting for you ([`commands/architecture.md:132`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/architecture.md#L131-L133): "**Write the file without waiting for approval** — an architecture is reversible…"), and the triage afterwards is scoped to the brownfield subset ([`:133`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/architecture.md#L132-L134): "If brownfield technology findings seeded any defaults, triage them with the user now"). And "uses them as the defaults for each architecture decision instead of proposing technologies from scratch" is only true where findings exist — [`commands/architecture.md:114`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/architecture.md#L113-L115): "When brownfield or documentation findings provided a known technology, use it as the default; **otherwise pick a sensible best-practice default and label it as an assumption**."

You already explain this write-then-refine shape well for `/awos:product` at `:56` — reusing that framing here would fix both. Same claim in `docs/commands/architecture.md:27`.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
+# Adopting AWOS on an Existing Codebase (Brownfield)
+
+Most AWOS documentation assumes a **greenfield** project — you start from a blank slate and describe what you want to build. But AWOS works just as well on a **brownfield** project: an existing codebase with real source, history, and often external documentation (wikis, tickets, chat). This guide is the brownfield counterpart to the [Quick Start](../README.md#quick-start) — same commands, same document-centric workflow, with the extra awareness AWOS brings to code that already exists.
+
+You do not run a different set of commands for brownfield. The same foundation commands (`/awos:product`, `/awos:roadmap`, `/awos:architecture`) **auto-detect** your existing code and fold it into the context they build. This guide explains what that detection does, what it produces, and how to steer it.
```
AlexanderMakarov: Only `/awos:product` detects. The other two just check for the file: [`commands/roadmap.md:59`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/roadmap.md#L58-L60) — "Check if `context/product/brownfield.md` exists (produced by `/awos:product` when it detects an existing codebase)" — and [`commands/architecture.md:57`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/architecture.md#L56-L58) is identical in shape. Neither scans for source indicators.

This is CodeRabbit's auto-detection finding, and you fixed it in `README.md:24` — the phrasing there ("`/awos:product` auto-detects your code and stages what it finds, and the later foundation commands build on that") is exactly right and worth reusing here. Your table at `:12-14` and the Step 2/3 headings already get it right, so `:5` is the odd one out.

Related, and not stated anywhere yet: because detection is per-`/awos:product`, choosing **No, start from scratch** at Step 1 means no `brownfield.md`, so Steps 2 and 3 do zero brownfield work either. That's worth a line in "Opting in and opting out" — the opt-out is global, not per-step.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+| ------------------------------- | --------------------------- | ------------------------------------- | --------------------------------------------------------------------------------- |
+| `context/product/brownfield.md` | `/awos:product`             | `/awos:roadmap`, `/awos:architecture` | Deleted by `/awos:architecture` after absorption.                                 |
+| `context/sources/sources.md`    | `/awos:product` (via skill) | `/awos:product`, retrieval pass       | Removed by `/awos:architecture` once absorbed; kept + referenced if still useful. |
+| `context/audits/<timestamp>/`   | `/awos:ai-readiness-audit`  | You (review + recommendations)        | Permanent history; each run is an independent snapshot.                           |
+
+## Opting in and opting out
+
+- **Force brownfield:** include intent in your prompt — `/awos:product explore the existing codebase`.
+- **Force greenfield on a repo with code:** `/awos:product start from scratch, ignore existing code`.
+- **Skip external sources:** answer **No** when asked, or decline; a `## Status: none` marker records the choice so you are not asked again.
```
AlexanderMakarov: Answering **No** doesn't write anything. [`commands/product.md:99`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L98-L100) is just "If no, skip to substep 4" — no marker. "or decline" is the accurate half: the marker gets written when you pick *None* ([`SKILL.md:36`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L35-L37)), decline the data-egress confirmation ([`:49`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L48-L50)), or every source drops out at verification ([`:77`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L76-L78)). And you're right that an existing marker suppresses the question ([`commands/product.md:96`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L95-L97)).

The part I'd want documented is the footgun in the same branch. `commands/product.md:99`, handling a missing plugin: "if the skill is not found (plugin absent), inform the user they can install it from the marketplace … **Write `context/sources/sources.md` with `## Status: none`** and continue". So someone who answered **Yes** but hadn't installed the plugin gets the "user previously declined" marker, and `:96` never asks again — including after they install it. The recovery is to delete `context/sources/sources.md`, which isn't written down anywhere.

`:73` describes exactly this plugin-absent path and would be the natural place to mention it. (The cleaner fix is prompt-side — don't write the decline marker when the reason was a missing plugin — but that's a `commands/product.md` change, not this PR's job.)

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+## The brownfield path at a glance
+
+| Step  | Command                       | What's different for brownfield                                                             |
+| ----- | ----------------------------- | ------------------------------------------------------------------------------------------- |
+| **0** | `/awos:ai-readiness-audit`    | _(Recommended first.)_ Scores how AI-ready the codebase is and what to improve.             |
+| **1** | `/awos:product`               | Detects existing source, explores it, and drafts the product definition from what it finds. |
+| **2** | `/awos:roadmap`               | Assesses already-built capabilities so the roadmap starts from reality, not zero.           |
+| **3** | `/awos:architecture`          | Discovers the existing tech stack and uses it as the default architecture.                  |
+| **4** | `/awos:hire`, `/awos:flow`    | Identical to greenfield — set up specialist agents and the delivery flow.                   |
```
AlexanderMakarov: Two things about the `/awos:flow` row, and the second is a selling point you're giving away.

`/awos:flow` ships only in the plugin — at `a98d515` there's no `commands/flow.md`, no `claude/commands/flow.md` and no `docs/commands/flow.md`, only `plugins/awos/commands/flow.md`, and [`README.md:38`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/README.md#L37-L39) labels it "Available from AWOS plugin now." You're careful to include `/plugin install` for the audit at `:25`, so a reader who skipped Step 0 arrives at `:120` without `/awos:flow`. Worth marking it as plugin-provided.

And it isn't identical to greenfield — [`plugins/awos/commands/flow.md:28`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/commands/flow.md#L27-L29) lists `context/sources/sources.md` as an optional input, "when present with `## Status: configured`, provides already-configured transports for ticket trackers, documentation platforms, and communication tools; **avoids re-probing those services**" — reinforced at [`:69`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/commands/flow.md#L68-L70) ("Adopt those transports directly rather than re-probing") and [`:81`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/commands/flow.md#L80-L82). Since `:88` says `sources.md` can survive `/awos:architecture`, that's a real brownfield benefit that carries past the foundation — which also makes `:120` ("no brownfield-specific behavior" from here on) not quite true.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+This step also **cleans up the temporary brownfield artifacts**:
+
+- `context/product/brownfield.md` is **deleted** — its findings have been absorbed into the product definition, roadmap, and architecture documents by now.
+- `context/sources/` is **removed once its contents are fully absorbed**. If durable **non-secret** configuration remains useful (source URLs, tool names, scopes to reach a wiki again later), `context/sources/sources.md` is **kept** and referenced from `product-definition.md` so downstream commands can find it. Credentials and tokens never belong under `context/` — supply those through environment variables or a secret manager, and keep only the pointers to them here.
+
+After this step, your `context/` looks the same as a greenfield project's — a clean set of permanent documents (`product-definition.md`, `roadmap.md`, `architecture.md`) with no brownfield scaffolding left behind.
```
AlexanderMakarov: "no brownfield scaffolding left behind" contradicts `:88` two lines up, where `sources.md` is kept and referenced when durable config is still useful — and per [`commands/architecture.md:152`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/architecture.md#L151-L153) that's the normal branch rather than an edge case. `:98` also has `context/audits/<timestamp>/` as permanent. Suggest "…apart from a retained `context/sources/sources.md` when durable source config is still useful, and your audit history under `context/audits/`."

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+With `brownfield.md` present, `/awos:architecture` runs an `Explore` pass to discover the **existing technology stack** — languages, frameworks and versions, databases and ORMs, infrastructure (Docker, cloud configs, deployment scripts), external services, and testing/build/CI tooling. It appends these under a `## Technology` heading, and uses them as the **defaults** for each architecture decision instead of proposing technologies from scratch. You still confirm or change every decision.
+
+This step also **cleans up the temporary brownfield artifacts**:
+
+- `context/product/brownfield.md` is **deleted** — its findings have been absorbed into the product definition, roadmap, and architecture documents by now.
+- `context/sources/` is **removed once its contents are fully absorbed**. If durable **non-secret** configuration remains useful (source URLs, tool names, scopes to reach a wiki again later), `context/sources/sources.md` is **kept** and referenced from `product-definition.md` so downstream commands can find it. Credentials and tokens never belong under `context/` — supply those through environment variables or a secret manager, and keep only the pointers to them here.
```
AlexanderMakarov: The keep criteria dropped one of the prompt's three items — [`commands/architecture.md:152`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/architecture.md#L151-L153):

> `If useful information remains in context/sources/sources.md that hasn't been captured elsewhere (source URLs, user-pasted content, configuration details), keep the file and add a reference to it…`

User-pasted content is precisely the non-configuration case — it's how `Access: manual` sources get in at all ([`commands/product.md:101`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L100-L102)) — so framing retention as "durable non-secret *configuration*" tells a reader to expect pasted wiki content to be deleted when the prompt says keep it.

The credentials sentence itself is fine and true (the `sources.md` schema has no credential field), and I know it came out of CodeRabbit's secrets finding — I'd keep it. Just worth not presenting it in `docs/commands/architecture.md:27` as something the *command* enforces, since no prompt states it.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+This file is a **staging area, not a deliverable.** Its findings feed the drafted product definition, and after the definition is saved you **triage them** — each finding is presented for **Accept** or **Reject** (with room for free-text corrections). Rejected findings are removed from both the file and the saved definition. The file itself is temporary and is cleaned up later, at the architecture step.
+
+### External documentation sources
+
+Brownfield projects usually have knowledge that lives outside the code — a Confluence or Notion wiki, Jira or Linear tickets, Slack history, email threads. When `/awos:product` creates `brownfield.md`, it also offers to import that knowledge:
+
+> Do you have external documentation (wikis, tickets, chats, email) you'd like to import into the project context?
+
+If you opt in, the command invokes the `awos:configure-external-sources` skill (part of the AWOS plugin), which:
```
AlexanderMakarov, line 64: The bullets walk the whole external-sources flow but skip its one blocking gate — [`configure-external-sources/SKILL.md:49`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L48-L50):

> `Use AskUserQuestion to confirm: "Retrieving content from these sources will send data to the LLM provider's API. This data may include sensitive or personal information (PII in tickets, internal discussions in wikis, private messages in chats). Do you wish to proceed?" with options **Yes, proceed** and **No, skip external sources**. If skipped, write context/sources/sources.md with ## Status: none and stop.`

For this doc's audience — a team about to point AWOS at a corporate Confluence or Slack — that's the most consequential prompt in the flow, and a genuine opt-out that "Opting in and opting out" should list. It stands out more because you do spend two sentences on credential hygiene at `:88`. One bullet plus one line in the opt-out section.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+- identifies each source and its category (documentation, tickets, communication),
+- guides you through connecting the right tool — an MCP server, a CLI like `gh`, or manual paste,
+- records the configuration in `context/sources/sources.md`, and
+- once connected, retrieves product-relevant content (requirements, goals, audience, pain points) and folds it into the same triage flow as the code findings.
+
+**Editor-restart handling.** Some MCP servers only become available after the editor restarts. If setup requires that, the skill writes a partial config (`## Status: restart-pending`) and stops. Restart your editor and re-run `/awos:product` — the skill detects the pending state and resumes automatically where it left off. No progress is lost.
```
AlexanderMakarov: Two details narrower than the skill. The trigger isn't "if setup requires a restart" — [`SKILL.md:65`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L64-L66) is "**If any MCP servers were added** during Step 4, write … `## Status: restart-pending`", i.e. any MCP addition. And `restart-pending` isn't the only resumable state: [`SKILL.md:21-22`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/configure-external-sources/SKILL.md#L20-L23) defines `verifying` and `verified` too, and [`commands/product.md:98`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L97-L99) re-invokes on all three. Something like "if any MCP server was added, the skill saves a partial config and stops; re-running resumes from wherever it left off" covers both.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+## Step 1 — `/awos:product`: detection, exploration, and external sources
+
+In **Creation Mode** (no `product-definition.md` yet), `/awos:product` decides whether to treat the project as brownfield, then builds context accordingly.
+
+### How detection works
+
+The command chooses whether to explore existing code in this order:
+
+1. **Your prompt wins.** If you say something like "explore the existing codebase" or "brownfield", it explores. If you say "start from scratch", "greenfield", or "ignore existing code", it skips detection and treats the project as new.
+2. **Otherwise it looks.** If your prompt says nothing either way, the command scans for common source indicators — `src/`, `app/`, `lib/`, `package.json`, `requirements.txt`, `go.mod`, `Cargo.toml`, `pom.xml`, `Gemfile`, `build.gradle`, `*.csproj`, `pyproject.toml`, and similar. If none are found, it proceeds as greenfield. If any are found, it asks you to confirm: **Yes, explore the codebase** or **No, start from scratch** (defaulting to explore if the question goes unanswered).
```
AlexanderMakarov: The indicator list is missing `Makefile`, `CMakeLists.txt` and `setup.py` from the fifteen in [`commands/product.md:73`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/commands/product.md#L72-L74). "and similar" keeps it from being wrong, but those three are exactly the cases where a C/C++ or older-Python reader would conclude their repo won't trip detection. Three words.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+
+- **A baseline.** It tells you where the codebase stands _before_ AWOS, so you can measure improvement later.
+- **A punch list.** `recommendations.md` surfaces the highest-leverage gaps (missing `CLAUDE.md`, thin tests, no agent configs) that make the rest of the AWOS flow more effective.
+- **A shared vocabulary.** The scores give the team a concrete, agreed-upon picture of the starting point.
+
+**The audit is a fixed measuring stick, so run it first without hesitation.** Its _methodology_ is self-contained — the score is computed by an engine, not by the specialist agents `/awos:hire` later installs — so it measures the repository the same way before and after adoption. The _score itself_ reflects the state of the repo and is expected to change as you improve it: a first run on an untouched brownfield repo will score the AI-tooling and spec-driven-development dimensions low, and that low score _is_ the honest baseline. Adopt AWOS (foundation, `/awos:hire`, specs), re-run the same audit, and those dimensions climb — the measuring stick didn't move, the codebase did. The audit is read-only and each run is an independent snapshot, so re-running costs nothing.
```
AlexanderMakarov: "re-running costs nothing" undersells the cost — an audit is a multi-turn LLM flow with connector fetches and, in org mode, subagent fan-out; [`ai-readiness-audit/SKILL.md:118`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/skills/ai-readiness-audit/SKILL.md#L117-L119) is explicitly about minimizing engine calls because "in single-repo mode wall time is dominated by the number of serial turns". The property you want is safety, not free: "re-running is safe — it never modifies your source and never reads a previous audit" says the true thing, and the rest of this paragraph is accurate.

plugins/awos/README.md
```diff
@@ -18,6 +18,8 @@ To register the marketplace manually:
 /plugin marketplace add provectus/awos
 ```
 
+> **Adopting AWOS on an existing codebase?** The audit is the recommended first step of the brownfield path — run it before the foundation commands to get a baseline score and a punch list, then re-run it after adoption to see the delta. Its scoring is self-contained (it does not depend on the specialist agents `/awos:hire` installs), so it works the same on an untouched repo. See the [Brownfield Adoption Guide](../../docs/brownfield-adoption.md).
```
AlexanderMakarov: "re-run it after adoption to see the delta" reads like a feature, and there isn't one — [same file at `:86`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/plugins/awos/README.md#L85-L87) says every audit "stands alone; earlier directories are history, never input". The guide gets this right at `:35`. Suggest "compare the two reports" so the two new pieces of prose in this PR agree.

README.md
```diff
@@ -21,9 +21,7 @@ npx @provectusinc/awos
...
 
-> **Running on an existing codebase?** Start with an AI readiness audit to understand how AI-friendly your project is. Install the plugin with `/plugin install awos@awos-marketplace`, then run `/awos:ai-readiness-audit` to get a scored assessment with actionable recommendations for improvement. [Learn more](plugins/awos/README.md)
->
-> AWOS integrates seamlessly with existing projects — the setup commands auto-detect your codebase and use it as context, so you won't start from a blank slate.
+> **Running on an existing codebase?** AWOS integrates seamlessly with brownfield projects — `/awos:product` auto-detects your code and stages what it finds, and the later foundation commands build on that so you won't start from a blank slate. The recommended first step is an AI-readiness audit: install the plugin with `/plugin install awos@awos-marketplace`, then run `/awos:ai-readiness-audit` for a scored assessment with actionable recommendations. See the **[Brownfield Adoption Guide](docs/brownfield-adoption.md)** for the full path — how detection works, what it produces, and how to steer it.
```
AlexanderMakarov: Preference rather than a defect, so take or leave it — but this rewrite puts the brownfield mechanism ahead of the audit recommendation and ends on "how detection works", and that's the ordering I asked to reverse [on #137](https://github.com/provectus/awos/pull/137#discussion_r3419722670): keep the audit explanation leading, put brownfield exploration after it, and pitch seamless integration rather than explaining how it works inside.

To be fair to the rewrite: it's much shorter than the version that prompted that, it does keep "The recommended first step is an AI-readiness audit", and both install commands survive. But if you swap the two clauses — audit first, then "AWOS integrates seamlessly with brownfield projects, `/awos:product` auto-detects your code and the later commands build on that" — and end the guide link on the outcome rather than the mechanism, it lands the same content in the shape I was after.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,120 @@
...
+Steps 1–4 are the same foundation commands every project runs. The only brownfield-specific addition is **Step 0**, the audit — and the fact that Steps 1–3 read your code instead of starting from a blank page.
+
+## Step 0 — Measure AI-readiness first (the audit)
+
+Before you describe the product, it is worth knowing how ready the existing codebase is for agent-driven development. The AWOS plugin ships an **AI-readiness audit** that scores a repository across a dozen dimensions — code architecture, quality assurance, security, documentation, spec-driven-development health, AI tooling — and produces a report with prioritized, actionable recommendations.
+
+```
+/plugin install awos@awos-marketplace
+/awos:ai-readiness-audit
+```
```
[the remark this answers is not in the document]
AlexanderMakarov: `text` is the right call over `sh` here — the block is two slash commands, not shell. [MD040](https://github.com/DavidAnson/markdownlint/blob/main/doc/md040.md) is satisfied by any specifier by default ("The default value is `[]` which means any language specifier is valid"), and this repo pins no `allowed_languages` — no markdownlint config at all, and CI runs prettier rather than markdownlint. Nothing further needed; this can be resolved.

AlexanderMakarov (changes requested):
Fills a real gap — the brownfield path has been shippable for a while with nothing user-facing describing it, and the artifact-lifecycle table at `:94-98` is the part I'd keep exactly as written; every cell in it checks out.

Accuracy is what I'd want resolved before merge, since describing the prompts is this doc's whole job. Six load-bearing claims don't match the prompts — all inline, all a sentence each, none needing code. Two of them (`:79` and `:88`) came out of fixing CodeRabbit's points and landed past what the framework actually does, so those want a deliberate decision rather than another reword.

Three things that aren't tied to a line:

**The `docs/commands/*.md` sections duplicate the guide, and one has already drifted inside this PR.** `docs/commands/roadmap.md:29` repeats the guide's incorrect roadmap claim, so one prompt fact now has two files to fix; `docs/commands/architecture.md:27` restates all of guide `:83-88` in a single 120-word line, carrying two of the defects with it. I'd make those three sections pointer-shaped — a couple of sentences of orientation plus the guide link — and let the guide be the one place the behavior is described.

**You're right about not shipping tests here, and it's worth saying so since it's the kind of thing that gets asked reflexively.** `tests/lint-prompts.test.js` reads nothing under `docs/`, so there's no docs→prompt contract today — but it does already pin the prompt side of nearly everything this guide quotes, and CLAUDE.md's same-PR-test rule enumerates framework structural contracts, not prose pages. Decisive bit: a lint asserting each quoted string appears in the prompt would have caught none of the findings above, because they're all semantic — a doc asserting a rule the prompt doesn't contain, or crediting work to the wrong actor. Cutting the duplication above beats testing it.

**Discoverability.** [`README.md:60`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/README.md#L60) and [`:72`](https://github.com/provectus/awos/blob/a98d515e023febf1069727205f53c02b1340e701/README.md#L72) give `docs/rationale.md` and `docs/testing-strategies.md` a dedicated top-level `## ` section each, and there's no `docs/` index, so those sections are the index. This guide is now the second-longest thing in `docs/` and is reachable only from inside the Step 1 blockquote. A three-line `## Brownfield Adoption` section next to `## Testing Strategies` would match how the others are surfaced — keeping the Step 1 link as the in-flow pointer.

docs/brownfield-adoption.md
dustyo-O: Good call — went the prompt route: roadmap.md now only marks fully-implemented capabilities as done. Also fixed the triage-timing sentence (saved first, triaged after).

docs/brownfield-adoption.md
dustyo-O: Reworded — an unanswered question falls back to its default and continues; a dismissed/denied one ends the turn; an MCP-add stops for a restart before the write. Added the CI caveat.

docs/brownfield-adoption.md
dustyo-O: Split it — the skill only writes the config, product does the retrieval. Fixed the same misattribution in the command doc.

docs/brownfield-adoption.md
dustyo-O: Fixed both — defaults only where a finding exists (else a best-practice default labeled an assumption), and reused the write-then-refine framing. Same in the architecture doc.

docs/brownfield-adoption.md
dustyo-O: Reused the README wording — only product detects. Added the global-opt-out note to the opt-in/out section too.

docs/brownfield-adoption.md
dustyo-O: Fixed — answering No writes nothing now; the marker only lands on None / declined egress / all-sources-drop. Documented the plugin-absent footgun and the delete-to-recover step.

docs/brownfield-adoption.md
dustyo-O: Marked flow as plugin-provided and noted the sources.md reuse. Softened the 'no brownfield behavior from here' line too, since that carries past the foundation.

docs/brownfield-adoption.md
dustyo-O: Reworded to acknowledge the retained sources.md and audit history — no longer claims a fully clean slate.

docs/brownfield-adoption.md
dustyo-O: Widened the keep criteria to include user-pasted content. Kept the credentials line, just stopped presenting it as command-enforced in the architecture doc.

docs/brownfield-adoption.md
dustyo-O: Added the data-egress confirmation as its own bullet and called it out in the opt-out section.

docs/brownfield-adoption.md
dustyo-O: Reworded — trigger is any MCP added, and it resumes from whatever state it saved (pending restart or mid-verification).

docs/brownfield-adoption.md
dustyo-O: Added Makefile, CMakeLists.txt and setup.py.

docs/brownfield-adoption.md
dustyo-O: Switched to safety framing — re-running is safe: never modifies source, never reads a prior audit.

plugins/awos/README.md
dustyo-O: Changed to 'compare the two reports' — no delta feature, agreed.

README.md
dustyo-O: Swapped the clauses — audit leads now, brownfield integration after, and it ends on the outcome rather than the mechanism.

dustyo-O:
Thanks, really useful review — pushed in 8f67659.

Addressed all the inline accuracy points. On the roadmap completion-status one I took the prompt route you suggested (added the rule to `commands/roadmap.md`) rather than reverting, so the doc claim is now true.

Two structural bits:
- The three `docs/commands/*.md` brownfield sections are now pointer-shaped, so the guide is the single source of truth.
- Added a top-level `## Brownfield Adoption` section next to Testing Strategies for discoverability.

Lint 146/0, Prettier clean.

docs/brownfield-adoption.md
```diff
@@ -0,0 +1,122 @@
...
+
+After this step, your `context/` holds the same permanent documents as a greenfield project's (`product-definition.md`, `roadmap.md`, `architecture.md`) — the temporary `brownfield.md` is gone, apart from a retained `sources.md` when durable source config is still useful, plus your audit history under `context/audits/`.
+
+## Artifact lifecycle
+
+| Artifact                        | Created by                  | Consumed by                           | Fate                                                                              |
+| ------------------------------- | --------------------------- | ------------------------------------- | --------------------------------------------------------------------------------- |
+| `context/product/brownfield.md` | `/awos:product`             | `/awos:roadmap`, `/awos:architecture` | Deleted by `/awos:architecture` after absorption.                                 |
+| `context/sources/sources.md`    | `/awos:product` (via skill) | `/awos:product`, retrieval pass       | Removed by `/awos:architecture` once absorbed; kept + referenced if still useful. |
```
AlexanderMakarov: The `sources.md` row undersells its consumers — and I own part of this, since I told you last round every cell in this table checked out. This one doesn't.

`/awos:roadmap` and `/awos:architecture` each gate on `## Status: configured` and run their own retrieval pass ([`commands/roadmap.md:86`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/roadmap.md#L86), [`commands/architecture.md:85`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/architecture.md#L85)), and `/awos:flow` reads it for transports ([`plugins/awos/commands/flow.md:28`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/plugins/awos/commands/flow.md#L28)) — the reuse you added at `:15` and `:122`.

As written the row reads as if the file goes cold after `/awos:product`, which undercuts both the retention rationale at `:89` and the flow payoff you just documented. There's also something in it a reader would want to know: a wiki connected once gets queried again by roadmap and by architecture, so that single egress consent covers three passes, not one.

`/awos:product`, `/awos:roadmap`, `/awos:architecture`, `/awos:flow` — each running its own retrieval pass.

README.md
```diff
@@ -75,6 +73,12 @@ The **`awos`** framework is flexible and non-prescriptive when it comes to testi
 
 [Explore testing strategies and customization options](docs/testing-strategies.md)
 
+## Brownfield Adoption
+
+Adopting **`awos`** on an existing codebase runs the same commands as greenfield, with extra awareness of code that already exists — plus a recommended AI-readiness audit up front. The foundation commands detect your source, inventory what's already built, and adopt your current stack as the architecture defaults.
```
AlexanderMakarov: Small, but it's the exact claim we spent last round narrowing: "The foundation commands detect your source" puts detection back in the plural. Only `/awos:product` detects ([`commands/product.md:73`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/product.md#L73)); the other two key off whether `brownfield.md` exists. `:24` in this same commit gets it right.

It reads acceptably as a distributive list if you squint, so take or leave — but one clause settles it: "`/awos:product` detects your source; the later foundation commands inventory what's already built and adopt your current stack as the architecture defaults."

docs/brownfield-adoption.md
AlexanderMakarov: The detection narrowing landed, but the replacement clause overshot it. `:5` now says roadmap and architecture "build on that staged context rather than scanning the codebase themselves" — and they both scan:

- [`commands/roadmap.md:66`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/roadmap.md#L66) — "Explore this codebase and assess what's already built"
- [`commands/architecture.md:65`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/architecture.md#L65) — "Explore this codebase and document the existing technology stack"

The guide contradicts itself three times over: `:18` ("Steps 1–3 read your code"), and `:78` / `:84` both describing those Explore passes.

The true statement is narrower than either version — they don't re-detect, but they do explore. Something like: "…build on that staged context — they don't re-detect, but each runs its own focused `Explore` pass (capabilities, tech stack) seeded with the staged findings."

`README:78` went the other way on the same claim; separate comment there.

docs/brownfield-adoption.md
AlexanderMakarov: Trigger and resume wording are both right now. One thing sitting next to them that I missed last round: the recovery action itself isn't what the skill tells the user to do.

[`SKILL.md:67`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/plugins/awos/skills/configure-external-sources/SKILL.md#L67) prints: exit the session (Ctrl+C twice or `/exit`) and resume it with the `claude --resume <id>` from the exit message, then run `/mcp` if any server needs interactive auth. The guide says "Restart your editor and re-run `/awos:product`" — so a CLI user restarts an IDE they aren't in, loses the session id the skill's own message points at, and the auth step is gone entirely. Both halves the guide drops are real: `--resume` takes a session id ([CLI reference](https://code.claude.com/docs/en/cli-reference)), and `/mcp` is what authenticates a server needing OAuth ([MCP docs](https://code.claude.com/docs/en/mcp) — "Use `/mcp` to authenticate with remote servers that require OAuth 2.0 authentication").

In fairness `commands/product.md` says "editor restart" too, so the guide is faithfully copying one prompt while diverging from the message the user actually reads — I'd follow the skill's text.

Rider on the parenthetical: [`SKILL.md:19-21`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/plugins/awos/skills/configure-external-sources/SKILL.md#L19-L21) routes three saved states, not two — `restart-pending` and `verifying` both resume at Step 6, `verified` resumes at Step 7 (scope collection).

docs/brownfield-adoption.md
AlexanderMakarov: The guide got it — `:84` has the write-first-then-triage framing now. The pointer doc didn't come along: [`docs/commands/architecture.md:27`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/docs/commands/architecture.md#L27) still says "uses those findings as the **defaults** for the architecture decisions they cover (you confirm or change each one…)", which reads as approval before the write. [`commands/architecture.md:132`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/architecture.md#L132) writes "**without waiting for approval**", and the triage happens after.

Same fix as the guide's: "(brownfield-seeded defaults are triaged with you after the document is saved; where no finding exists, it falls back to a best-practice default labeled as an assumption)".

docs/brownfield-adoption.md
AlexanderMakarov: Both landed — `:15` and `:122` mark flow as plugin-provided and name the `sources.md` reuse.

One leftover two lines below: `:18` is now the odd one out. It still says "The only brownfield-specific addition is **Step 0** … and the fact that Steps 1–3 read your code", which was true before you added the Step 4 carryover to the table above it. Appending "…plus `/awos:flow` reusing any retained source config" keeps the paragraph from contradicting the table it's summarizing.

docs/brownfield-adoption.md
AlexanderMakarov: All three fixed. One refinement on the footgun — it's wider than plugin-absent.

[`commands/product.md:99`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/commands/product.md#L99) writes the marker after *any* failed skill invocation: "if the skill is not found (plugin absent), inform the user … if the call was rejected for another reason, report the error message to the user. Write `context/sources/sources.md` with `## Status: none`". So a denied permission prompt earns the same permanent suppression a missing plugin does.

"if you answer **Yes** and the skill call fails for any reason (plugin missing, call denied)" covers both, and the delete-to-recover step you already added handles the way out.

docs/brownfield-adoption.md
AlexanderMakarov: Gate's there, and calling it out in the opt-out section too was the right move.

Take or leave: the bullet sits third, after "guides you through connecting the right tool", but the skill runs it second — [Step 3 privacy gate](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/plugins/awos/skills/configure-external-sources/SKILL.md#L49) comes before [Step 4 tool setup](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/plugins/awos/skills/configure-external-sources/SKILL.md#L51). Your "before anything is fetched" is accurate so nothing is wrong — it just has the reader planning an MCP install before learning the consent question comes first. Swapping the two bullets matches the order they'll meet them.

AlexanderMakarov (commented):
Everything from the last round is addressed — and taking the prompt route on the roadmap completion rule rather than reverting the doc was the right call; it makes the claim true instead of merely careful.

Four things left, three of them the fix commit's own doing.

The guide's opening paragraph and the new README section moved in opposite directions on the same claim. Narrowing detection to `/awos:product` was correct, but the opening now says `/awos:roadmap` and `/awos:architecture` "build on that staged context rather than scanning the codebase themselves" — they both scan, each with its own `Explore` pass, and the guide's own Step 2 and Step 3 sections say exactly that. The README section swings the other way and puts detection back in the plural. Both are one-clause fixes.

The artifact-lifecycle table needs one cell widened — and this is me correcting myself, since I told you last round every cell in it checked out.

`docs/commands/architecture.md` still reads as a pre-save approval gate in its Brownfield mode section. The guide picked up the write-then-refine framing; the pointer doc didn't come along.

Two for here rather than a line:

**The audit needs Node and the guide never says so.** Step 0 tells the reader to run it "without hesitation", but [`ai-readiness-audit/SKILL.md:79`](https://github.com/provectus/awos/blob/8f676597eb1090e31f6f23e59dd5e6f3e036577e/plugins/awos/skills/ai-readiness-audit/SKILL.md#L79) preflights `command -v node` and stops without it. This guide's whole audience is existing Java/Python/Go/.NET repos, which is exactly where a Node runtime is least likely to already be sitting — a one-sentence prerequisite next to the install/run block would save the first-run failure.

**Step 0 also points at a list that is already a row short.** It sends readers to the plugin README "for the full dimension list", and that table is missing Prevention Coverage, which `standards.toml`'s `dimension_order` carries. That gap arrived from main rather than from you, so I'd leave it out of this PR — flagging it only so the referral doesn't quietly inherit it.

Last, small: the description still says "Documentation only — no behavior or engine changes". `commands/roadmap.md` is a prompt change and you called it out in the thread; worth putting in the body so the release label matches what actually ships.

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — the `sources.md` artifact row now lists all four consumers (`/awos:product`, `/awos:roadmap`, `/awos:architecture`, `/awos:flow`), each running its own retrieval pass, so it no longer reads as going cold after product.

README.md
dustyo-O: Fixed in 00698f5 — reworded to your suggestion: "`/awos:product` detects your source; the later foundation commands inventory what's already built and adopt your current stack as the architecture defaults."

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — `:5` now reads "…build on that staged context — they don't re-detect, but each runs its own focused `Explore` pass (capabilities, tech stack) seeded with the staged findings," so it no longer contradicts :18/:78/:84.

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — the restart paragraph now follows the skill's actual message: exit (Ctrl+C twice or `/exit`) → `claude --resume <id>` → `/mcp` for interactive auth → re-run `/awos:product`, and it names all three saved states (restart-pending/verifying resume at tool verification, verified resumes at scope collection). Also aligned the "Unattended runs" note away from "editor restart."

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — `docs/commands/architecture.md:27` now reads "(brownfield-seeded defaults are triaged with you after the document is saved; where no finding exists, it falls back to a best-practice default labeled as an assumption)," matching the write-first framing in the guide.

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — `:18` now appends "…plus `/awos:flow` reusing any retained source config," so the paragraph matches the Step 4 carryover in the table above it.

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — the footgun now reads "if you answer **Yes** and the skill call fails for any reason (plugin missing, call denied)," covering the denied-permission case too.

docs/brownfield-adoption.md
dustyo-O: Fixed in 00698f5 — swapped the bullets so the data-egress consent gate precedes tool setup, matching skill Step 3 → Step 4.

docs/brownfield-adoption.md
AlexanderMakarov: `:27` reads right now. My fault for pointing you at one line: `docs/commands/architecture.md:20`, seven lines up, now says the opposite of it.

Two things wrong in that one sentence, and both predate this PR — the prompt flipped to write-first in [`72b615a`](https://github.com/provectus/awos/commit/72b615a) (#141) and this bullet, which dates to #99, never followed. Your `:27` edit is what puts the contradiction in the same file.

1. "You confirm or change each decision before moving on" promises a gate before the write. [`commands/architecture.md:112`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/architecture.md#L111-L113) forbids exactly that — "never blocking on a question before the write" — and [`:132`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/architecture.md#L131-L133) writes "without waiting for approval". The only triage is [`:133`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/architecture.md#L132-L134), after the write, and it covers brownfield-seeded defaults rather than every decision.
2. "works through the architecture template section by section" reads as incremental too. `:112` drafts every area in one pass.

Something like: "The agent drafts every architectural area in one pass and saves `architecture.md` without waiting for approval; you then review the saved document and, on brownfield, accept or reject the findings that seeded the defaults."

`docs/commands/product.md:30` already has that shape. It's pre-existing, so leaving it for a follow-up is entirely fair — but it's two lines in a file you already have open.

docs/brownfield-adoption.md
AlexanderMakarov: The condition is right now. The recovery I gave you last round isn't — deleting `context/sources/sources.md` won't get you asked again on the project where this bites.

The question fires only from [`commands/product.md:99`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/product.md#L98-L100), inside Step 2B Creation Mode. In this footgun's own scenario the run doesn't stop — the failure branch says "continue to substep 4", so it reaches [`:139`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/product.md#L138-L140) and writes `product-definition.md`. The next `/awos:product` then hits [`:52-55`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/product.md#L51-L56) and routes to Update Mode, whose steps only jump to template sections — never to that substep.

Deleting the file is still necessary, just not sufficient: [`SKILL.md:23`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/plugins/awos/skills/configure-external-sources/SKILL.md#L22-L24) stops immediately if `## Status: none` is still there. What works is delete, then run `/awos:configure-external-sources` yourself — plugin skills are user-invocable as `/plugin:skill` by default ([skills docs](https://code.claude.com/docs/en/skills)). Worth naming the ceiling too: `/awos:product` won't re-read the sources afterwards, so what you configure lands via `/awos:roadmap` and `/awos:architecture`.

So roughly: "…won't ask again. Deleting `context/sources/sources.md` doesn't bring the question back on its own — the run already wrote `product-definition.md`, so the next `/awos:product` enters Update Mode. To recover, delete the file and run `/awos:configure-external-sources` yourself; `/awos:roadmap` and `/awos:architecture` will pick the sources up."

docs/brownfield-adoption.md
AlexanderMakarov: The CI caveat landed. There's a fourth case it doesn't cover, and it's the one that bites hardest, because the lifecycle table promises the opposite.

`:98` says `brownfield.md` is "Deleted by `/awos:architecture` after absorption", unconditionally. Unattended it isn't. [`commands/architecture.md:133`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/architecture.md#L132-L134) triages the brownfield-seeded defaults with `AskUserQuestion`, and the cleanup that deletes the file is [Step 5](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/architecture.md#L150-L152), two steps later. In `-p` a dismissed question ends the turn before either runs — the exact structural trap your own editor note at [`commands/roadmap.md:35`](https://github.com/provectus/awos/blob/00698f56aeec48748526eaffccda27e47c3ae455/commands/roadmap.md#L34-L36) describes: "a deliverable Write placed after such a question never runs unattended."

The `architecture-reads-brownfield` E2E hits it — the cleanup check fails headless and `brownfield.md` is still on disk afterwards. It fails identically on stock `main`, so this isn't something you introduced. But this PR is the doc that promises the lifecycle, so it's the right place to scope the promise to interactive runs.

Adding it to this bullet would cover it: "…and `/awos:architecture`'s brownfield triage sits ahead of its cleanup step, so unattended the run ends before `brownfield.md` is deleted — re-run interactively, or delete it yourself."

AlexanderMakarov (approved):
All eight from the last round landed — thanks for the turnaround.

Two things only became visible once they did, and both are on me: I pointed at a single line where the sentence beside it needed the same fix. Replies on those two threads. Neither is a regression from this commit and neither should hold the PR — approving; take them here or in a follow-up, your call.

A third came out of running the brownfield scenarios: the artifact lifecycle's "deleted by `/awos:architecture`" doesn't hold in `claude -p`. Reply on the unattended-runs thread. It reproduces on stock `main`, so it's a caveat this guide should carry rather than a bug for you to fix here.

The scope line wants a small correction too. "Documentation only — no behaviour or engine changes, so no new test contracts" stopped being true when we took the prompt route on the completion-status point and changed `commands/roadmap.md` step 2d. awos-qa#46 adds the behavioural coverage for it — worth linking from here.

One for a tracked follow-up rather than this PR. The guide now says "session reload" while every prompt still says "editor restart" — `plugins/awos/skills/configure-external-sources/SKILL.md:5` and `:20`, `commands/product.md:98` and `:99`, `CLAUDE.md:94`. The divergence is deliberate and I asked for it, since the guide should follow the message users actually read. But it leaves one state with two names across the repo. Landing the new vocabulary in the prompts is the other half of that change.

While in there: `commands/product.md:99`'s "Write `context/sources/sources.md` with `## Status: none`" is a standalone sentence, not bound to the failure branch it follows. The guide reads it as failure-only, which is plainly the intent. A literal reader could apply it on the success path and clobber a manifest the skill just wrote — worth an "On failure," at the front.

docs/brownfield-adoption.md
dustyo-O: Fixed in d47f407 — scoped the promise to interactive runs. Added the case to the Unattended runs bullet (`/awos:architecture`'s brownfield triage runs ahead of its cleanup step, so unattended the run ends before `brownfield.md` is deleted — re-run interactively or delete it yourself) and qualified the lifecycle table's Fate cell the same way.

docs/brownfield-adoption.md
dustyo-O: Fixed in d47f407 — rewrote `docs/commands/architecture.md:20` to the write-first shape: the agent drafts every architectural area in one pass and saves `architecture.md` without waiting for approval, then you review the saved document. Matches `commands/architecture.md:112`/`:132`; the brownfield accept/reject stays at `:27`.

docs/brownfield-adoption.md
dustyo-O: Fixed in d47f407 — the footgun note now says deleting `sources.md` isn't sufficient (the run already wrote `product-definition.md`, so the next `/awos:product` enters Update Mode). Recovery is delete the file and run `/awos:configure-external-sources` yourself; `/awos:roadmap` and `/awos:architecture` pick the sources up.

PR #168, opened 2026-07-30
fix(audit): state observations, not conclusions, in detector evidence

### LINKED ISSUE #156: audit: detector evidence strings assert conclusions instead of observations
**Effort:** ~1–2 days

**Profit:** **high** — every FAIL/WARN evidence line in every report is affected, so users see the difference on essentially every real-project run. The current conclusion-style wording is also hard for a reader to check: it asserts a capability-level fact ("Claude Code cannot run the application without human involvement") that the reader must disprove by knowing the repo, instead of a file-level observation they can verify in seconds. Fixing it converts the least-trustworthy sentences in the report into the most-trustworthy ones.

## Problem

Detector evidence strings assert the *definition-level conclusion* instead of the *observation actually made*. Example from `detectors/ai_development_tooling.ts` (AI-07, `detectCanRunApp`):

> no run mechanism found — no Makefile, docker-compose, package.json start/dev script, build-tool wrapper (mvnw/gradlew), manage.py, or Procfile; **Claude Code cannot run the application without human involvement**

The detector only proved "no conventional entrypoint file at the repo root". The bolded clause is a claim about agent capability that is false for, e.g., a Maven repo whose CLAUDE.md documents `scripts/backend-start.sh` — exactly the case reported in [Slack feedback](https://provectus.slack.com/archives/C093YCJ6RRU/p1784808778900659?thread_ts=1783932694.550419&cid=C093YCJ6RRU):

> `(so every verification loop depends on a human running Maven/npm by hand.)` - this is not correct argumentation

Because these strings ship with `reliability_default = "maximal"`, every proxy-vs-definition divergence becomes a confidently stated falsehood in a client-facing report.

## Fix

Sweep all detectors (and computed metrics that emit evidence) so that:

- Evidence text states **what was checked and what was found** (globs/patterns searched, files matched, counts), never the capability-level conclusion.
- The capability-level *definition* stays where it already lives — the `standards.toml` `definition` field, surfaced via the report hint column.
- FAIL evidence for list-based detectors enumerates the signals looked for, phrased as "checked for X, Y, Z at repo root; none present" rather than "the project cannot do W".

## Definition of done

- [ ] Every detector FAIL/WARN evidence string in `detectors/*.ts` and `metrics/*.ts` describes the observation (what was searched, where, what matched), with no capability/consequence claims ("cannot", "depends on a human", "nothing prevents").
- [ ] A Layer-1-style lint or engine test greps evidence-producing strings for the banned conclusion phrasings and fails on regression.
- [ ] Engine tests updated where they pin old evidence text; `npm run build:audit-engine` re-run and `dist/` committed.
- [ ] Plugin version bumped (three-file discipline) since report output changes.
- [ ] Spot-check: re-run the audit on a Maven repo without root entrypoint files — AI-07 evidence reads as an observation and is factually accurate.

AlexanderMakarov:
Closes #156.

## The problem

Detector evidence strings asserted the definition-level conclusion instead of the observation actually made. The reported case was AI-07:

> no run mechanism found — no Makefile, docker-compose, package.json start/dev script, build-tool wrapper (mvnw/gradlew), manage.py, or Procfile; **Claude Code cannot run the application without human involvement**

The detector had only established that a fixed list of root files was absent. The bolded clause is a capability claim, and it is false for a Maven repo whose CLAUDE.md documents `scripts/backend-start.sh` — the case raised in Slack. Because these strings ship with `reliability_default = "maximal"`, every divergence between a detector's proxy and the capability it stands for became a confidently stated falsehood in a client-facing report.

## The change

Evidence now states what was checked and what was found: the patterns and globs searched, the paths matched, counts and denominators. The capability claim stays in `references/standards.toml`'s `definition` field, which the report already surfaces as the hint and tooltip beside the evidence — so "why it matters" is still there, stated as *what the standard requires* rather than *what is true of this repo*. That is the correct owner of the sentence, precisely because the proxy can diverge from the definition.

AI-07 now reads:

> no run mechanism found at repo root — checked for Makefile, docker-compose.yml, docker-compose.yaml, run.sh, start.sh, justfile, Justfile, Taskfile.yml, Taskfile.yaml, mvnw, gradlew, manage.py, Procfile, and package.json start/dev script; none present

Interpolated from `ROOT_RUN_FILES`, so the sentence cannot drift from the list the code actually searches. The same treatment is applied to MCP config paths, README names, TLS config globs, and env-example globs, and the two `.env` pattern lists now share one constant — that drift is what produced an enumeration missing `.env*.local`, the Next.js default, while the regex accepted it.

Statements about the engine's own measurement are deliberately kept: incomputability ("cannot compute a pass rate"), applicability and SKIP-gate reasons, scoring-rule notes, and threshold citations. Those describe the measurement, not the audited project.

## Claims the code never checked

The sweep surfaced several strings asserting things no branch had established. These are corrected in wording only — no detection logic was added, so no score moves:

- `documentation.ts` reported "missing setup instructions" on a branch that only tested content length.
- `security.ts` claimed a Dockerfile `COPY .` would leak a file, having only checked that a Dockerfile exists.
- `supply_chain_security.ts` reported that dependency updates required review without checking branch protection or CODEOWNERS.
- `application_security.ts` named pbkdf2/sha256/sha512 as found on a branch reachable when only a CSPRNG signal is present — a repo with no password-hashing code at all was told weaker hashing algorithms were found. That branch is now split, with a regression test pinning it.
- A coverage gate claimed a threshold "is enforced" when only a pattern in CI or config had been found.
- QA-01 rendered one ratio two ways, so adjacent evidence lines read 59.5% and 60%, the second next to "below the 60% pass threshold".

## Verification

**No audit number moves.** The engine was run from the pre-change and post-change sources over two repositories and three synthetic fixtures, and every field except `evidence` was compared: per-check status, value, score, confidence and weight; per-dimension totals; audit totals and coverage ratios. All identical, to full float precision. The fixtures were built specifically to reach the two branches that were split, since neither real repository exercises them.

`npm test` is green across all four layers (1568 tests). `dist/` is rebuilt and committed. Prettier is clean.

## Regression protection

`tests/evidence-phrasing.test.ts` scans every string literal under `detectors/`, `metrics/` and `audit_core.ts` for banned conclusion phrasings. Its positive controls are derived mechanically — every string literal in the changed detectors is tokenized before and after the sweep and set-diffed — so the contract is "this diff cannot be reverted silently" rather than "these sentences won't come back". The scanner tracks comment, string, template and regex state rather than grepping raw source, and merges both `+` concatenation and template interpolation so a banned phrase split across a seam is still caught. The escape hatch requires a stated reason and is pinned at zero uses, so introducing the first one is a visible review decision.

Nine co-located regression tests pin the individual fixes — constant interpolation for AI-07/AI-04/DOC-01/AS-01, one-decimal rendering for QA-01/SBP-03, threshold-citation ordering for AS-05, code-point ranges for AIS-01, and the AS-07 branch split. Each was verified by reproducing the defect it guards and confirming the test fails.

**Known limit, stated plainly:** a lexical lint cannot judge whether a string is *true of the branch it sits on*. Every falsehood found during review of this work was banned-word-free and was caught by reading code, not by the lint. One residue class — inferring functional status from the presence of a directory — is documented in the test as out of reach.

## Also here

- `detectors/README.md` documents the evidence-phrasing contract, cross-referenced from `metrics/README.md`.
- `dimensions/application-security.md` drops advice and verdict wording its bullets had mirrored from the old evidence.
- Plugin version 2.4.3 → 2.4.4, since report output changes.
- CLAUDE.md and a lint assertion message are corrected to the four-file version-bump discipline the lint actually enforces — `plugins/awos/commands/flow.md` carries a generator-version constant that the three-file description omitted. Note this means the bump marks previously generated flow artifacts as stale on the next re-run; that is the pre-existing design, not new behavior.

## Summary by CodeRabbit

* **New Features**
  * Updated the AWOS plugin to version 2.4.4.
  * Improved AI readiness audit results with clearer evidence, thresholds, searched locations, and measured percentages.
  * Added more precise reporting across security, documentation, testing, delivery, and development checks.

* **Documentation**
  * Documented guidelines for consistent, fact-based audit evidence.

* **Bug Fixes**
  * Corrected misleading or incomplete audit messages and improved reporting consistency.

* **Tests**
  * Added regression coverage for evidence accuracy, thresholds, and detector output.


plugins/awos/skills/ai-readiness-audit/detectors/security.ts
```diff
@@ -21,14 +21,19 @@ import { ALL_HOOK_PATHS } from '../agent_tools.ts';
 const ENV_GITIGNORE_RX =
   /^\s*(\.env(\.\*|\*(\.local)?)?|\*\.env|\*\*\/\.env|\/\.env)\s*(?:#.*)?$/m;
 
+// Human-readable form of the patterns checked in ENV_GITIGNORE_RX, shared by
+// the PASS and FAIL evidence lines so they can't drift apart.
+const ENV_GITIGNORE_PATTERNS =
```
AndreyNenashev: `ENV_GITIGNORE_PATTERNS` is still a hand-typed twin of `ENV_GITIGNORE_RX`. I expanded the regex and the seven patterns do line up today, but nothing keeps them lined up — `det-security.test.ts` exercises the regex, not the sentence. That's the exact seam that produced the `.env*.local` omission you describe in the PR body, and this ended up as the one hand-written rendering without a pin (AIS-01 got its range test, everything else interpolates the constant it searches). What I'd do: flip it so the list is the source of truth — `['.env', '.env.*', '.env*', '.env*.local', '*.env', '**/.env', '/.env']` — join it for the evidence string, and add a small test that each entry, written as a `.gitignore` line, matches the regex (plus something like `.envrc` that shouldn't). Then the sentence can't lie.

plugins/awos/skills/ai-readiness-audit/detectors/quality_assurance.ts
```diff
@@ -108,12 +108,19 @@ export function detectTestInfrastructure(
...
   const ratio = testCount / sourceCount;
-  const pct = Math.round(ratio * 100);
+  // One decimal for the measured value everywhere it's rendered in this
+  // check: at integer precision a measured ratio can round to the same
+  // display value as the (also rounded) threshold it's being compared
+  // against, producing a self-contradictory sentence like "70% — below the
+  // 70% pass threshold". Using one decimal consistently, rather than only
+  // in the threshold-comparison lines, keeps every evidence line for the
+  // same ratio agreeing with the others.
+  const pctDisplay = (ratio * 100).toFixed(1);
```
AndreyNenashev: One decimal shrinks the collision window from 0.5 pp to 0.05 pp, but it's still there (same story in `detectTypeSafety` and `detectAuthOnMutations`). Take 149 test files over 497 source modules: that's 29.98%, which renders as `"30.0% — below the 30% warn threshold"` (line 159). And the guard regex `/(\d+)% — below the \1%/` can't see it, since `30.0` ≠ `30` as text — so the "must never render as the same number" assertion holds for the fixture, not in general. Cheapest fix is to admit the residual in the comment and widen the guard to `/(\d+)(?:\.\d+)?% — below the \1(?:\.\d+)?%/` with a numeric compare. If you want to actually close it, bump precision only when the rendered value collides with the threshold that branch cites as "below":

```ts
let pctDisplay = (ratio * 100).toFixed(1);
if (Number(pctDisplay) === citedBelowThresholdPct)
  pctDisplay = (ratio * 100).toFixed(2);
```

Either way it's evidence-only — status and score don't move.

plugins/awos/skills/ai-readiness-audit/detectors/documentation.ts
```diff
@@ -75,12 +75,12 @@ export function detectRootReadme(
...
   if (!HEADING_RX.test(content)) {
     return makeResult('WARN', content.length, [
-      `${relPath} lacks a Markdown heading structure — may not be well-organised`,
+      `${relPath} has no Markdown heading (# …) or RST underline heading`,
     ]);
   }
 
   return makeResult('PASS', content.length, [
-    `${relPath} present with headings and setup instructions (${content.length} bytes)`,
+    `${relPath} present with headings and setup instructions (${content.length} characters)`,
```
AndreyNenashev: This line and the WARN at line 72 are two survivors of the class you fixed one branch up in this same function. `SETUP_CONTENT_RX` matches ten keywords, but both strings claim something stronger — that setup instructions exist or don't. A README that says "Bootstrapping: `./bootstrap.sh`" has setup instructions and matches nothing, so it gets told it "contains no setup/install/usage instructions." They slipped past the lint because there's no banned token in them. One-word-ish fixes:

- line 72: `` `${relPath} has no setup/install/usage/run/build keyword` ``
- line 83: `` `${relPath} present with heading(s) and setup-keyword match (${content.length} characters)` ``

Could also pull the keywords into a constant and interpolate, same as `ROOT_RUN_FILES`.

plugins/awos/skills/ai-readiness-audit/tests/evidence-phrasing.test.ts
```diff
@@ -0,0 +1,849 @@
...
+ * scanned. A banned phrase split across a `'a' + 'b'` concatenation seam or
+ * a `` `a ${x} b` `` template interpolation seam is still matched as one
+ * clause: adjacent `+`-joined literals are merged, and a template literal's
+ * static parts are read as ONE combined string with interpolations
+ * contributing nothing to the text (patterns tolerate the resulting
+ * variable whitespace via `\s+`), rather than being split into disconnected
+ * fragments at each `${...}`.
+ *
+ * The positive-control corpus below is derived mechanically from
+ * `git diff f7293e3 4f619bb` — every evidence string issue #156 actually
```
AndreyNenashev: Tiny one: this cites `git diff f7293e3 4f619bb` (also at lines 546 and 574), but `4f619bb` is a branch commit — after the squash-merge it won't resolve, and the "mechanically derived" claim becomes unverifiable. Point at issue #156 / this PR instead, or say "this PR's base..head".

AndreyNenashev (approved):
Ran everything locally on the PR head before reading in depth: all 8 evidence-phrasing tests and the 60 co-located detector tests pass, `npm run build:audit-engine` gives an empty diff against the committed `dist/`, prettier is clean, and the version says 2.4.4 in all four places the lint looks. I also traced the AS-07 split by hand — the fallthrough WARN really is only reachable when CSPRNG is the lone signal, and both WARN paths keep the same status and score.

Approving. I have four comments — all small, and honestly three of them are just your own patterns from this PR applied to spots it didn't reach. Nothing here touches detection logic or moves a score.

plugins/awos/skills/ai-readiness-audit/detectors/prevention_coverage.ts
```diff
@@ -586,13 +586,13 @@ export function detectTestCoverageGate(
 
   if (coverageHits.length > 0) {
     return makeResult('PASS', testHits.length + coverageHits.length, [
-      'CI runs the test suite and a coverage threshold is enforced',
+      'CI runs the test suite, and a coverage-threshold pattern found (in the CI workflow or a coverage config file)',
```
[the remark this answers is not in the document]
AlexanderMakarov: Fixed in c9de28e. Added the missing `was` to the PASS branch, and to the paired WARN branch on the next line so the two read alike — "no coverage-threshold pattern found" was idiomatic on its own, but reads inconsistent sitting next to the corrected PASS text.

plugins/awos/skills/ai-readiness-audit/detectors/software_best_practices.ts
```diff
@@ -238,18 +238,21 @@ export function detectTypeSafety(
...
     }
     if (ratio >= warnAt) {
       return makeResult('WARN', pct, [
-        `${pct}% of Python function signatures carry return-type annotations — some typing present but not enforced by a type checker`,
+        `${pctDisplay}% of Python function signatures carry return-type annotations in a sample of up to 20 .py files (single-line signatures) — below the ${passAtPct}% pass threshold, at or above the ${warnAtPct}% warn threshold (no mypy/pyright config found)`,
       ]);
     }
     return makeResult('FAIL', pct, [
-      `${pct}% of Python function signatures carry return-type annotations — project appears essentially untyped`,
+      `${pctDisplay}% of Python function signatures carry return-type annotations in a sample of up to 20 .py files (single-line signatures) — below the ${warnAtPct}% warn threshold (no mypy/pyright config found)`,
```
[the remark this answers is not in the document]
AlexanderMakarov: Leaving this as-is, deliberately — the asymmetry is the fix, not an oversight.

`software_best_practices_sbp03.test.ts`, added in this PR, pins exactly this split: the measured value renders at one decimal while the threshold citations stay integral. Rendering both at the same precision is what produced the self-contradictory sentence issue #156 reported ("60% — below the 60% pass threshold"); giving the measurement more precision than the threshold is what stops the two colliding.

The example also is not reachable from configuration. Every `pass_at`/`warn_at` in `references/standards.toml` is a whole percentage — SBP-03 is `0.60`, the others `0.50` and `0.70` — so `pass_at: 0.605` would have to be introduced first. Meanwhile one-decimal thresholds would print "at or above the 60.0% pass threshold" on every run.

If a fractional threshold is ever configured, the fix I would take is trailing-zero trimming rather than blanket `toFixed(1)`: `60%` stays `60%`, and `0.605` would render `60.5%`.

plugins/awos/skills/ai-readiness-audit/detectors/supply_chain_security.ts
```diff
@@ -578,13 +577,13 @@ export function detectDependencyAutomationReview(
...
     return makeResult('WARN', foundFiles.length, [
-      'dependency automation configured but automerge is enabled — updates may merge without human review',
+      'dependency automation config found with automerge enabled',
       ...foundFiles.map((f) => `config: ${f}`),
     ]);
   }
 
   return makeResult('PASS', foundFiles.length, [
-    `dependency automation configured with review required: ${foundFiles.join(', ')}`,
+    `dependency automation configured without automerge enabled: ${foundFiles.join(', ')}`,
```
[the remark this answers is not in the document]
AlexanderMakarov: Fixed in 9ae37cb. An unreadable config was pushed into `foundFiles` and then skipped past the content check, so it counted toward the PASS while its contents were never inspected — and the evidence went on to assert "without automerge enabled" about a file nobody opened. Same observation-vs-conclusion problem this PR is closing, one level down.

Unreadable paths are now tracked separately from read ones:

- some readable, none matched → PASS, evidence scoped to "no `automerge: true` pattern was found in the config file(s) read", with the unreadable paths listed separately rather than folded into the claim;
- no config readable at all → WARN, since automerge is unknown rather than absent;
- `automerge: true` found → WARN, and no config at all → FAIL, both unchanged.

Added tests for the mixed and wholly-unreadable cases, neither of which had coverage.

plugins/awos/skills/ai-readiness-audit/detectors/security.ts
AlexanderMakarov: Done in aca7b1e, taking the flip you describe.

`ENV_GITIGNORE_PATTERNS` is now the array, the evidence joins it, and the old comment claiming the two lines "can't drift apart" is gone — it was true of the PASS/FAIL pair and false of the thing that matters, since neither was tied to the regex that decides.

The test walks every entry as the literal `.gitignore` line and asserts the detector accepts it, with `.envrc` as the negative so it cannot pass by the regex matching everything.

plugins/awos/skills/ai-readiness-audit/detectors/documentation.ts
AlexanderMakarov: Fixed in 3ba6202, with your wording.

- line 72 → `${relPath} has no setup/install/usage/run/build keyword`
- line 83 → `${relPath} present with heading(s) and setup/install/usage/run/build keyword match (N characters)`

Also pulled the keywords into a `SETUP_KEYWORDS` constant the sentences interpolate, mirroring `ROOT_RUN_FILES`, so the wording stays tied to what is searched rather than being retyped next to it — same failure mode as the AS-12 thread.

plugins/awos/skills/ai-readiness-audit/tests/evidence-phrasing.test.ts
AlexanderMakarov: Fixed in 43a724c — all three sites (35, 546, 574) now cite issue #156 and this PR's base..head instead of `4f619bb`.

plugins/awos/skills/ai-readiness-audit/detectors/quality_assurance.ts
AlexanderMakarov: You are right, and this also corrects a reply I posted earlier on this PR. CodeRabbit raised the threshold-rendering question from the other direction (a fractional `pass_at`), I pushed back because nothing in `standards.toml` configures one — and then framed the collision as closed. It was not. The residual comes from the measurement side, which I did not consider.

Verified your case directly before fixing: 149/497 is 29.9799%, renders `"30.0% — below the 30% warn threshold"`, and `/(\d+)% — below the \1%/` returns **false** on that string. So the guard was holding for its fixture, exactly as you said.

Took the second option, in 945242c. `formatMeasuredPct` in `_base.ts` renders at one decimal and adds precision only when the result would read as equal to a threshold the same sentence calls it "below". One refinement on your snippet: it only guards against thresholds the value is genuinely *below*, so a ratio sitting exactly on a threshold still renders `30.0%` rather than escalating to `30.000%` — "at or above the 30%" is accurate there, so there is nothing to disambiguate.

Applied to all three sites you named — QA-01, SBP-03, AS-06 — since it was the third copy of the shape. The guard in `software_best_practices_sbp03.test.ts` now extracts both numbers and compares them numerically instead of by backreference. New test pins your exact ratio, and I confirmed it reproduces `"30.0% — below the 30%"` when the formatting is reverted.

Evidence-only, as you said: status and score do not move.

PR #169, opened 2026-07-31
fix(prompts): stop unanswerable gates from costing the deliverable

AlexanderMakarov:
Three ways an AWOS command could stop partway through and leave the user with nothing, found by the awos-qa e2e suite. Each is a behavior fix; the tests are how they surfaced, not what they are for.

## `/awos:hire` produced no coverage report

`/awos:hire` could run for a while and then stop without ever writing `context/product/hired-agents.md` — its one durable output. Two separate places did it.

Step 2 ended with "show the user a table … and confirm before proceeding". Written as prose rather than a question, it is a pause with nothing to resume it. Worse, it pauses on information the user has no particular reason to act on: at that point nothing has been installed or written, and the same roles come back in Step 3 with their coverage status and again in the Step 8 roster. The step now presents the proposals and carries on, and says where the information reappears.

The second was the install-consent gate in Step 4. That gate is right and stays — it installs shell scripts that run automatically on lifecycle events, and it must never be auto-approved. What was wrong was treating a gate that got no answer as a reason to abandon everything after it. Consent covers *installing*; it does not cover reporting. An unanswered gate now means nothing from that gate is installed, the command continues, and Step 8 records what is actually on disk, listing the skipped components under Gaps for a later run.

Step 6 also showed each generated agent file "for approval before saving". Agent files are reversible, so they are written and then shown for review, like the other commands' artifacts.

## `/awos:product` wrote a file after the user declined

Answering **No** to "do you have external documentation to import?" still produced `context/sources/sources.md`. Both branches of the question had been written as one run-on sentence, with the Yes-branch's fallback write trailing the No-branch's "skip to substep 4".

The file is not inert: the same substep reads an existing `## Status: none` back as "previously declined" and stops asking. So a decline recorded as a file quietly became permanent — the user said "not this time" and the command heard "never ask again". The branches are now separate, and the fallback write is scoped to the case it was meant for: the user said Yes and the skill could not be invoked.

## `/awos:spec` produced acceptance criteria that were not testable

Step 3.3 described Given/When/Then as three ingredients — a precondition, an action, an outcome — which is guidance about content, not about form. Specs came out with most criteria in the When/Then shape and the occasional one written as a flat statement ("The user submits X. They see Y."), which reads fine but is not the shape QA and downstream tooling rely on. The rule now says the shape explicitly, with a conforming example and the two ways it slips.

## Tests

Layer-1 lints pin all three behaviors, each verified to fail against the original wording. They match the contract rather than a fixed phrasing — an earlier revision pinned verbatim sentences, which pushed the prompts toward prose written for the linter instead of for the reader.

Full suite: 197 / 1324 / 20 / 11.

## Verification

Validated live against the awos-qa suite, serially, and re-validated after the wording rework:

- `hire-discovers-plugin-agents` — was 0/1 with no report written
- `product-external-docs-declined`
- `spec-uses-gwt-acceptance-criteria`

`spec-writes-functional-spec-unattended` was additionally run as a control on unmodified `main` and on this branch to confirm the spec change does not regress the write-first guarantee.

## Summary by CodeRabbit

* **New Features**
  * Hiring workflows continue through coverage analysis when consent cannot be obtained, skipping installations and listing proposed components as gaps.
  * Generated or updated agent files are written automatically and displayed for review afterward.
  * Acceptance criteria now use a consistent single-sentence When/Then format, with optional Given conditions.

* **Bug Fixes**
  * Declining external documentation import no longer creates an unnecessary status artifact.
  * Status reporting distinguishes unavailable integrations from other configuration failures.


commands/spec.md
```diff
@@ -122,7 +122,10 @@ Your first goal is to determine the **topic** - the single, specific feature or
...
     - Acceptance criteria must read as manual QA test scripts that a non-developer could execute. Describe only what is visible on screen and what the user does — never reference internal system behavior.
-    - Each acceptance criterion follows the same three-part shape as the example below: a precondition (Given), a user action (When), and a visible outcome (Then). Include Given only when the precondition affects the outcome.
+    - Each acceptance criterion follows the same three-part shape as the example below: a precondition (Given), a user action (When), and a visible outcome (Then). Include Given only when the precondition affects the outcome. The words **when** and **then** must both appear, in that order, within a **single sentence**, in every criterion — including short or seemingly obvious ones. Two ways it slips: splitting the pair across two sentences, and replacing it with a declarative retelling.
+      - ✅ "When the user submits the form with an empty name, then they see: 'Name is required.'"
+      - ❌ "The user submits the form with an empty name. They see: 'Name is required.'" — no `when`/`then` pair.
```
dustyo-O: what is the difference between this line and previous?

dustyo-O (approved):

commands/spec.md
AlexanderMakarov: Same content, different sentence form — that's the entire difference, and the fact that it takes a diff to see it is a fair hit on the wording.

The ✅ carries both literal keywords inside one sentence ("When …, then …"). The ❌ says exactly the same thing, split into two declarative sentences with neither keyword present. The rule here is lexical, not semantic: the behavioral check in awos-qa (`spec-uses-gwt-acceptance-criteria`) tests every acceptance-criterion bullet against `/\bwhen\b[^.!?]*?\bthen\b/i`, so a criterion that reads perfectly but drops the keywords — or splits them across a sentence boundary — fails it. The examples are deliberately a minimal pair so the model sees that only the wording changed, not the meaning — but if the contrast has to be inferred, it isn't teaching.

Fixed by annotating both sides instead of only the ❌:

- ✅ "When the user submits the form with an empty name, then they see: 'Name is required.'" — one sentence carrying both `when` and `then`.
- ❌ "The user submits the form with an empty name. They see: 'Name is required.'" — the same content split into two declarative sentences, with neither keyword.

PR #170, opened 2026-07-31
feat(audit): incidents source collector for a measured MTTR (DF-07)

dustyo-O:
## What

Makes the audit's **`incidents` source real** so **MTTR (DF-07 / category 1103)** is measured from actual incidents instead of only a git proxy. Git recovery signals were already covered (`change_failure_rate` + the existing git-proxy MTTR); the gap was ingesting a real incident source — the `sources = ["tracker", "incident"]` slot in `standards.toml` had no collector, so 1103 never scored.

Roadmap item: **Audit deterministic incidents collector** (Improvement).

## How

- **`collectors/incidents.ts`** *(new)* — connector-passed, mirroring `tracker`/`ci`/`docs`. The orchestrator fetches incidents from a real source — PagerDuty/OpsGenie/incident.io, Statuspage/Atlassian, or code-host incident labels — normalizes them to `IncidentRecord[]`, and passes them; the collector computes per-incident recovery spans and their median. No connector → `available:false` (SKIP), so **git-only repos are unaffected**.
- **`metrics/mttr.ts`** — a real-source branch takes precedence (median recovery time, reliability `maximal`, awards 1103 when ≥1 incident is resolved). **The git-proxy path is preserved unchanged as the fallback.**
- **`references/connector-shapes.md`** — the Incidents recipe: `IncidentRecord`/`IncidentsConnector` shapes + per-platform mapping. Maps **semantically from the live tool response** (open time → `started_at`, restore → `resolved_at`); vendor field names are typical-as-of-writing hints only. `SKILL.md` already lists `incident` among fetchable connectors.
- Wiring: `audit_core` base-pass collection + `has_incident_source` (a resolved incident, or a tracker incident source — the `.raw` read is also corrected); `artifact_types` registers the `incidents` source; `standards.toml` + `dimensions/delivery-flow.md` update DF-07.

## Design notes for review

- **Additive / low-risk.** A git-only repo scores exactly as before (incidents `available:false` → git proxy → DF-07 stays SKIP). Only a repo with a connected incident source changes — which is the point.
- **Drive-by fix.** `has_incident_source` previously read `trackerArt.incident_source` (top-level, always `undefined`); corrected to `.raw.incident_source`. A tracker-declared incident source now actually flips the flag.

## Testing

- **`tests/incidents-collector.test.ts`** — connector shaping (spans, median, open-incident exclusion, SKIP-without-connector).
- **`tests/met-mttr-incidents.test.ts`** — the real-source metric branch + git-proxy fallback preserved.
- **`tests/det-incidents-mttr.test.ts`** — end-to-end: `audit-core` (DF-07 SKIP) → write connector-shaped `incidents.json` → `enrich` → DF-07 becomes measured 2h / awarded / maximal.
- **Doc-drift lint** — `connector-shapes.md` pinned to the collector's TS interfaces.
- The existing git-proxy MTTR tests are untouched and pass. Full engine suite green; `dist/` rebuilt & committed.

## Follow-up (separate)

Extending the onboarding `configure-external-sources` skill with an "incident management" category (to wire up a PagerDuty/OpsGenie MCP the audit then auto-detects) — intentionally out of scope here; the audit assesses reachable connectors, it doesn't configure them.

## Summary by CodeRabbit

- **New Features**
  - Added incident-based MTTR measurement using resolved recovery spans and configurable lookback windows.
  - MTTR now reports median recovery time, DORA classification, sample counts, and reliability details.
  - Added support for incident data from recognized sources, including Statuspage and code-host incident labels.
  - Git branch-lifetime values remain available as clearly labeled, non-reliable fallbacks.

- **Bug Fixes**
  - Improved handling and warnings for open, invalid, malformed, or out-of-window incidents.
  - Reports and rollups now prefer measured MTTR values.
  - Incident source labels no longer incorrectly increase MTTR reliability or scoring.


plugins/awos/skills/ai-readiness-audit/references/connector-shapes.md
```diff
@@ -503,3 +503,42 @@ gh issue list --state all --limit 500 --json number,title,state,createdAt,closed
...
+  "started_at": "2026-07-01T09:12:00Z", // ISO 8601 — opened/detected
+  "resolved_at": "2026-07-01T11:42:00Z", // ISO 8601 — service restored; omit/null if still open
+  "severity": "SEV1", // optional, verbatim
+  "source": "pagerduty", // optional provenance, e.g. "github-label:incident"
+}
+```
+
+### IncidentsConnector
+
+The object the orchestrator assembles and writes to `collected/incidents.json`:
```
AlexanderMakarov: This recipe and the engine disagree about who computes the median, and the gap is silent in the worst case. The metric's new branch trusts only the derived fields — `raw.resolved_count` / `raw.median_duration_hours` (metrics/mttr.ts:81), and the topology gate reads `raw.resolved_count` too — but nothing in the shipped flow computes them: `audit-core` calls the collector with no connector, `enrich` skips collection entirely, and no CLI verb takes one, so `collect()`'s median logic is unreachable outside the unit tests. An orchestrator following this section literally writes `{incidents, source_label}`: as a bare object it at least trips the stranded-payload warning, but wrapped in the proper `{available: true, raw: {…}}` envelope with no derived fields, nothing fires — mttr silently falls back to the git proxy and DF-07 reads "Not applicable" while valid PagerDuty data sits on disk. And when the orchestrator does hand-write `median_duration_hours`, the engine republishes that number verbatim as maximal-reliability "measured from N resolved incidents" — an arithmetic slip in the orchestrator becomes a measured DORA number. The tracker already solved this shape problem: `collectors/tracker.ts` exports `buildTypeCounts`/`countResolved` so the metrics derive aggregates themselves, and its recipe says the orchestrator "does not need to produce this directly". This is also the roadmap item's central requirement — MTTR "computed deterministically by the engine from actual incident records instead of being an orchestrator-transcribed display string" — and hand-written `median_duration_hours` is exactly an orchestrator-transcribed number, just relocated into an artifact. I'd do what the tracker does — export `durationHours`/`median` from the collector, derive from `raw.incidents[]` in mttr.ts and the topology gate when the aggregates are absent (treating orchestrator-supplied ones as advisory), and warn when an available incidents artifact has records but no measurable span. That also makes this recipe truthful as written, and gives the e2e a variant worth adding: feed the envelope with only `incidents[]` and assert it still scores. (Small tell: det-incidents-mttr.test.ts:4 calls the hand-written full-envelope artifact "connector-shaped" — it isn't; it's the derived shape the doc never mentions.)

plugins/awos/skills/ai-readiness-audit/audit_core.ts
```diff
@@ -597,11 +602,22 @@ export async function auditCore(
...
     has_tracker: Boolean(trackerArt?.available),
     has_docs_connector: Boolean(docsArt?.available),
+    // A real, measurable incident source: the connector-passed incidents
+    // artifact with at least one resolved incident, or a tracker that names an
+    // incident source. Gates DF-07 (category 1103); the git proxy stays SKIP.
     has_incident_source: Boolean(
-      trackerArt?.available && trackerArt?.incident_source
+      (incidentsArt?.available && (incidentsRaw?.resolved_count ?? 0) > 0) ||
+      (trackerArt?.available && trackerRaw?.incident_source)
     ),
```
AlexanderMakarov, line 769: The `.raw` fix is right — the old top-level read meant this branch never fired in any audit to date — but making it live changes scoring in a way the PR's own docs disavow: a tracker that merely *names* an incident system now flips `has_incident_source`, and the git-proxy fallback still calls `awardCategories`, so 1103's full weight 5 is granted on a not-reliable proxy with zero incident data. That contradicts the new standards.toml definition ("awarded when at least one incident has a resolved recovery span"), the DF-07 What-line, and the comment two lines up. There's also no test pinning either leg of this expression — reverting the `.raw` fix fails nothing. I'd either drop the tracker disjunct (award only on measured spans) or keep it deliberately and align the three doc strings; either way a couple of pinning tests would guard it (tracker-declared path → flag true; incidents available with zero resolved → flag false).

plugins/awos/skills/ai-readiness-audit/dimensions/delivery-flow.md
```diff
@@ -73,7 +73,7 @@ Measures how fast and how safely change flows from commit to the default branch,
 
 ### DF-07: Mean time to recovery
 
-- **What:** Mean time to recovery from incidents; computed from git as a proxy by default (merge/revert/hotfix cadence), upgraded when a real incident source is present in the tracker artifact. Always included — never omitted from the artifact. SKIP only if even git is unavailable.
+- **What:** Median time to recovery from incidents. Measured from a real incident source — the incidents collector (`collectors/incidents.ts`), fed by the orchestrator from PagerDuty/OpsGenie/incident.io, Statuspage, or code-host incident labels (see `references/connector-shapes.md`). Category 1103 is awarded when at least one incident has a resolved recovery span. Without a source it falls back to a git proxy (merge branch-lifetime) and stays SKIP. Always included — never omitted from the artifact.
```
AlexanderMakarov: The new What-line and the retained Skip bullet now contradict each other: here, no source → git proxy "and stays SKIP"; three lines down, SKIP is only valid when git is unavailable, and Pass explicitly includes the git-proxy value. Two different SKIPs are conflated — the DF-07 check (SKIP whenever `has_incident_source` is false; your own e2e asserts that baseline) and the metric (never returns SKIP). I'd write "…and category 1103 stays SKIP" here and rewrite the Skip bullet to name the check, not the metric. Same wording slip in the audit_core.ts:617 comment ("the git proxy stays SKIP").

plugins/awos/skills/ai-readiness-audit/metrics/mttr.ts
```diff
@@ -66,6 +66,49 @@ export function compute(
...
+  //     median of those spans — a first-class measurement, reliability maximal.
+  const incidentsRead = readArtifact(collectedDir, 'incidents');
+  if (!('error' in incidentsRead) && incidentsRead.artifact?.available) {
+    const iraw = (incidentsRead.artifact.raw ?? {}) as {
+      resolved_count?: number;
+      median_duration_hours?: number | null;
+      source_label?: string | null;
+    };
+    const resolved = iraw.resolved_count ?? 0;
+    if (typeof iraw.median_duration_hours === 'number' && resolved > 0) {
```
AlexanderMakarov: The module header above still states the opposite of this branch — "no MTTR is ever computed from incident data here … reliability/confidence stay at the proxy's minimal level" — and its "Source shapes" list omits `incidents.json`, now the primary source. Worth rewriting for the two-path contract while you're in here. The header also carries two pre-existing self-contradictions worth fixing in the same pass: the proxy is uniform branch-lifetime over all first-parent merges, not "intervals between consecutive revert/hotfix/rollback merge commits", and "(incident_source field upgrades reliability)" is denied by the body itself.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/** Recovery time in hours for a resolved incident, or null when unmeasurable. */
+function durationHours(inc: IncidentRecord): number | null {
+  if (!inc.resolved_at) return null;
+  const start = Date.parse(inc.started_at);
+  const end = Date.parse(inc.resolved_at);
+  if (Number.isNaN(start) || Number.isNaN(end) || end < start) return null;
+  return (end - start) / 3_600_000;
+}
+
```
AlexanderMakarov: Every malformed record lands here as null and silently becomes "unresolved": a mis-mapped `started_at`, epoch-millis instead of ISO, or swapped fields all vanish without a trace, so a whole-batch mapping mistake reads as "no measurable incidents → Not applicable" with zero signal. `code_host` got a dedicated "N of M records lack a parseable timestamp" warning for the identical failure shape — incidents deserve the same: an `invalid_count` in the raw plus an artifact warning, and a "measured from N of M" note when only part of the sample is usable. Related edge: `end === start` counts a 0h span as measured, so auto-resolved alert flaps can band "elite". No test currently covers reversed spans or garbage dates.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
```diff
@@ -0,0 +1,91 @@
...
+    return makeArtifact(
+      'incidents',
+      false,
+      'no incident source connected; supply a PagerDuty/OpsGenie/incident.io, Statuspage, or code-host incident-label connector to measure MTTR from real incidents',
+      period,
+      {} as IncidentsRaw
+    );
+  }
+
+  const incidents = connector.incidents ?? [];
```
AlexanderMakarov: Nothing clamps these to the audit window: the recipe asks the orchestrator to query the same window, but unlike CI runs and tracker tickets — which the engine clamps defensively regardless of what it's sent — incidents are accepted as-is, including future-dated or all-time history. An over-fetched source silently shifts the "audit-window" median at maximal reliability and makes the number non-reproducible across differently-scoped fetches. A `started_at` within `window_anchor − lookback_days` filter wherever the derivation ends up (see the connector-shapes comment) would close it.

tests/lint-prompts.test.js
```diff
@@ -4058,3 +4058,39 @@ test('spec.md captures boundary/error behavior as rules in items 2 and 3', () =>
...
+      new RegExp(`export interface ${iface} \\{([\\s\\S]*?)\\n\\}`)
+    );
+    assert.ok(body, `collectors/incidents.ts must define interface ${iface}`);
+    return [...body[1].matchAll(/^\s*(\w+)\??:/gm)].map((m) => m[1]);
+  };
+  const fields = [
+    ...fieldsOf('IncidentRecord'),
+    ...fieldsOf('IncidentsConnector'),
+  ];
+  const missing = fields.filter((f) => !doc.includes(f));
```
AlexanderMakarov: Building on CodeRabbit's word-boundary point — it's a bit worse than a nitpick: `doc.includes('id')` is satisfied by the word "incident", and `source`/`incidents`/`count` by ordinary prose, so about half the guarded fields are checked vacuously. And the interfaces it checks are the input shapes; `IncidentsRaw` — whose `resolved_count`/`median_duration_hours` are what the engine actually reads — isn't covered, which is exactly the contract the recipe currently omits (see the connector-shapes comment). Matching identifiers inside the jsonc fences (or `\b`-bounded) and including `IncidentsRaw` would make the guard catch the drift that matters.

AlexanderMakarov (changes requested):
This lands the right architecture for a measured MTTR — a connector-passed collector mirroring tracker/docs, the git proxy preserved as fallback, and an e2e through the shipped bundle; the mechanics the roadmap item lists (new collector, `COLLECTOR_SOURCES`, the metric, standards.toml, `dist/` rebuilt) are all here. But the item's core ask is MTTR 
> computed deterministically by the engine from actual incident records instead of being an orchestrator-transcribed display string

and as it stands, that's the one part that didn't land - the engine only trusts derived aggregates that nothing in the shipped flow computes, so the median is still orchestrator-computed and republished verbatim at maximal reliability (inline on connector-shapes.md). Add to that the newly-live tracker gate awarding the category against the PR's own definition (inline on audit_core.ts), and the measured number isn't yet trustworthy end-to-end. Three things that don't map to changed lines:

**The executive headline still denies the feature.** `computeDerivedDelivery` (audit_core.ts:79-140) only reads the tracker, and `DerivedDelivery.mttr` has no value slot (unlike its sibling `cycle_time`), so after a successful enrich the headline row still renders "— (needs incident connector)" — or the now-false "incident source declared — no incident data mapped" — while the DF-07 detail right below shows a measured, maximal-reliability median. That's the headline/detail contradiction the 5bca168 fix was for, quoted in `computeDerivedDelivery`'s own docstring ("deriving both the value and the honest gated note from the SAME artifact makes that contradiction impossible"). SKILL.md and output-format.md also still say the MTTR headline is computed from `collected/tracker.json`. If threading the incidents artifact through the three call sites is out of scope here, a tracking note would help — as shipped, the report's most visible row contradicts its own detail.

**The plugin version isn't bumped.** `marketplace.json`, `plugin.json`, and the `EXPECTED_PLUGIN_VERSION` pin are all still 2.4.3, identical to main. This PR changes scoring behavior, and the engine stamps the plugin version into `audit.json` provenance and the report footer, so audits run after merge get mislabeled — the exact scenario the three-file rule exists for (#145/#150/#154 all bumped in-PR). If a separate release commit is the plan, worth saying so on the PR.

**`CONNECTABLE_SOURCES` / `COLLECTED_ARTIFACT_SOURCES` (audit_core.ts:1049-1066) never gained `'incidents'`.** Masked today — 1103 also lists `tracker` and is topology-gated, so neither `buildSkipReason` nor `canReuse` misbehaves yet — but the first category that lists `sources = ["incidents"]` alone gets a wrong skip reason and is wrongly treated as reusable on enrich, silently ignoring freshly-fetched connector data. The `'incident'` singular in those two sets is dead and can go (render.ts's `gated: 'incident'` union is a different, live concept — leave that alone).

plugins/awos/skills/ai-readiness-audit/references/connector-shapes.md
dustyo-O: Done — this is the main fix. The engine now derives the median itself from raw.incidents[] (deriveIncidentAggregates, shared by the metric, the gate, and the collector); any hand-written median_duration_hours is advisory and ignored. Reworked the recipe to the records-only envelope, and the e2e now feeds incidents[] only and asserts it still scores — the 'connector-shaped' mislabel is gone too.

plugins/awos/skills/ai-readiness-audit/audit_core.ts
dustyo-O: Dropped the tracker disjunct — 1103 is awarded only on a measured recovery span. standards.toml sources are incidents-only now, and I added the pinning tests: tracker-declared-only leaves DF-07 SKIP (e2e), plus hasMeasurableIncidents true/false legs (unit).

plugins/awos/skills/ai-readiness-audit/dimensions/delivery-flow.md
dustyo-O: Rewrote it — Pass/Skip now describe the DF-07 check via has_incident_source, and spell out that the mttr metric itself never SKIPs (it still reports the git proxy for context). Fixed the same slip in the audit_core comment.

plugins/awos/skills/ai-readiness-audit/metrics/mttr.ts
dustyo-O: Rewrote the header for the two-path contract, and fixed the two pre-existing contradictions: the proxy is a uniform branch-lifetime over all first-parent merges, and the incident_source line no longer claims a reliability upgrade. incidents.json is listed as the primary source now.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
dustyo-O: Added it — invalid_count in the raw plus a reliability note, and a 'measured from N of M' style note when only part of the sample is usable. Also fixed the 0h edge: end <= start is now invalid, so an auto-resolved flap can't band elite. Tests cover reversed spans, garbage dates, and the flap.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
dustyo-O: Clamped — deriveIncidentAggregates drops incidents whose started_at falls outside window_anchor - lookback_days, anchored to the newest record the same way clampToWindow does for CI/tracker. Test covers an over-fetched all-time history being dropped to the window.

tests/lint-prompts.test.js
dustyo-O: Fixed — the guard now matches each field as a standalone identifier (id is no longer satisfied by 'incident', nor count by 'resolved_count') and covers IncidentsRaw, so the derived fields the engine actually reads are both guarded and documented in the recipe.

dustyo-O:
Thanks — right call to push on this. All landed in 593e7c0, and I bumped the plugin to 2.4.4 (marketplace + plugin.json + the flow.md generator stamp + pin) as you flagged.

The core one: the engine now derives the MTTR median itself from `raw.incidents[]` — orchestrator-written aggregates are advisory and ignored — so it's a computed number, not a transcribed one. Everything else followed from that: the gate awards 1103 only on a measured span (tracker disjunct dropped), incidents are clamped to the window, unparseable/zero spans are counted and flagged, the executive headline reads the incidents artifact independently of the tracker (`DerivedDelivery.mttr` gained a value slot), and the dead `'incident'` source key is now `'incidents'`.

CodeRabbit's three nitpicks folded in too: the shared measurable-incident predicate (`hasMeasurableIncidents`), the word-boundary lint match, and the `Array.isArray` guard.

Engine 1343/0, lint 148/0, dist rebuilt.

plugins/awos/skills/ai-readiness-audit/render.ts
```diff
@@ -1313,6 +1313,7 @@ function normalizedDelivery(audit: AuditJson): DeliveryMetric[] {
   rows.push({
     label: 'MTTR',
     gated: 'incident',
+    ...(dd.mttr.display_value ? { display_value: dd.mttr.display_value } : {}),
```
AlexanderMakarov: The value lands here but the band doesn't. `DerivedDelivery.mttr` has no `band` field and the row goes in without a `check_id`, so a successful incidents connector renders the one row whose band the engine computes and then drops (`mttr.ts:60-65` already has `mttrBand`, and DF-07's own hint carries it).

I ran this against the shipped `dist/` bundle rather than guessing — fed a records-only `incidents.json` (spans 1h + 3h) through `enrich`, which correctly took DF-07 from SKIP to `PARTIAL`, weight 3.5/5, `median 2.0h MTTR over 2 incidents (high)`, maximal/HIGH. Rendering that audit gives:

```
| Metric | Value | Band |
| ------ | ----- | ---- |
| Deployment frequency | 4.8 / wk | high |
| Lead time for change | 1.1 h | elite |
| Change-failure rate | 4.8% | elite |
| Rework rate (DORA) | 6.5% | good |
| MTTR | 2 h | — |
```

Every sibling DORA row carries its band; the measured MTTR is the only one blank, in a block headed "Delivery (vs DORA bands)".

There's a second casualty on the same line: the `(of N resolved; M lacked a parseable span)` caveat that `mttr.ts:93-96` builds can't reach the headline either, because both renderers only print `note` when the value is absent (`:675` and `:1442`), and with no `check_id` there's no tooltip to carry it instead. So a median over 3 measurable spans out of 60 resolved incidents prints as a bare `0.5 h`. Adding `band` + `check_id: "DF-07"` to `DerivedDelivery.mttr`, setting them in `computeDerivedDelivery`, and spreading them here covers both; folding the sample size into the value (`2.4 h (3 of 60)`) would be the belt-and-braces version.

Worth knowing this line is entirely unpinned — deleting the `display_value` spread leaves the render and derived-delivery suites green, because the only `derived_delivery` fixture in the tree has `mttr: {}`. A fixture with a measured MTTR would pin the row end to end, and it's the assertion that would have caught the rollup gap in the summary too.

(Small one while you're here: `computeDerivedDelivery`'s docstring at `audit_core.ts:76-82` still says both gated rows are computed "from the tracker artifact".)

plugins/awos/skills/ai-readiness-audit/metrics/mttr.ts
```diff
@@ -173,6 +236,9 @@ export function compute(
   if (incidentSource) {
     reliability = appendReliabilityNote(reliability, trackerPartialNote);
   }
+  // An incidents artifact that carried records but no measurable span is worth
+  // saying out loud — otherwise a whole-batch mapping mistake reads as "no data".
+  reliability = appendReliabilityNote(reliability, incidentsNoSpanNote);
```
AlexanderMakarov, line 308: This append sits after both early returns in the `'error' in gitRead` branch above, so when `git.json` is missing or unreadable the note never fires — and that's the one path with no other information to fall back on. An incidents artifact carrying records but no usable span, on a repo where git also failed, reports only `git-proxy, true value may differ; git.json not found`: a connector that is present and unusable, and the report says nothing about it. Setting it in the two reliability objects at `:157-164` and `:178-182` would cover it. Narrow case, but it's exactly the one the note exists for.

plugins/awos/skills/ai-readiness-audit/references/standards.toml
```diff
@@ -212,9 +212,9 @@ metric = "mttr"
 dimension = "delivery-flow"
 weight = 5
 method = "computed"
-definition = "Mean time to recovery from incidents; SKIP unless a real incident source (PagerDuty/OpsGenie/incident.io) is explicitly provided"
+definition = "Median time to recovery from incidents, measured from a real incident source (PagerDuty/OpsGenie/incident.io, Statuspage, or code-host incident labels) via the incidents collector; awarded when at least one incident has a resolved recovery span. Without a source, MTTR reports a git proxy and this category stays SKIP."
```
AlexanderMakarov: One word: "Without a source, MTTR reports a git proxy and this category stays SKIP" — but a source with zero measurable spans stays SKIP too, which is what the clause just before it says and what the gate actually does. "Without a measurable recovery span" would match both.

plugins/awos/skills/ai-readiness-audit/references/connector-shapes.md
AlexanderMakarov: The records-only recipe reads right, and the e2e feeding `incidents[]` alone is exactly the variant that makes it honest. Three residuals in the new section:

1. The envelope documents only `raw.source_label`, but `period.source_label` is what the Sources column reads (`deriveSourceWindows`). Tracker, docs, code host and CI all set it (`:175`, `:273`, `:321`, `:452`); without it, Connections & Sources shows the generic "incident source" while the MTTR note says "PagerDuty" — the contradiction `computeDerivedDelivery` exists to prevent.
2. `:553` describes the clamp as `started_at` within `window_anchor − lookback_days`, but `clampIncidentsToWindow` anchors to the newest incident's own `started_at`. That matches `clampToWindow`'s convention, so it's the sentence that's off — and worth fixing, because as written it implies an all-stale incident stream would be excluded, when in fact it's kept in full.
3. `:550` lists `invalid_count` as "missing, reversed, or zero-length timestamps" — a missing `resolved_at` is classified still-open, not invalid.

plugins/awos/skills/ai-readiness-audit/audit_core.ts
AlexanderMakarov: Gate reads right now and the pinning tests are there. Two follow-ons.

The stale contract outlives the code in several places — I've put the list in the summary since none of it is on a changed line, but the two nearest this thread: `connector-shapes.md:73` and `collectors/tracker.ts:26-28`/`:90-93` both still say declaring `incident_source` upgrades MTTR reliability "to first-class".

And the new pin is weaker than it looks: `tests/det-incidents-mttr.test.ts:174` asserts `assert.ok(!((after!.weight_awarded as number) > 0))`, which also passes when the field is `undefined`, missing, or `NaN`. `assert.equal(after!.weight_awarded, 0)` says what you mean — and it's largely redundant with the `status === 'SKIP'` assertion two lines above anyway.

plugins/awos/skills/ai-readiness-audit/dimensions/delivery-flow.md
AlexanderMakarov: 76-79 reads right now. Line 16 of the same file still carries both of the claims you just deleted from `mttr.ts`'s header, though: MTTR "from git as a proxy (merge/revert/hotfix cadence)" — it's branch-lifetime over all first-parent merges — and reliability that "upgrades automatically when a real incident source artifact is present", where presence isn't enough any more, it needs a measurable resolved span. `dimensions/ai-sdlc-adoption.md:13` carries the same sentence.

plugins/awos/skills/ai-readiness-audit/metrics/mttr.ts
AlexanderMakarov: Header reads accurately now, including the two pre-existing ones — thanks for taking those in the same pass. One residual in the Source shapes block: the tracker line says an `incident_source` label "adds an informational note on the proxy path only", but it also puts `tracker` into `sources_used` (`:232`), so DF-07's Sources column credits the tracker for a pure git-proxy number — and in the git-missing branch tracker becomes the *sole* entry, while `standards.toml` now declares `sources = ["incidents"]` for 1103. Either widen the sentence or drop tracker from `sources_used` on the proxy path; the second reads truer now that the label changes nothing about the value.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
AlexanderMakarov: `invalid_count`, the `end <= start` rejection and the tests all landed — the flap-bands-elite hole is closed.

One case the fix doesn't reach, because it fails a step earlier than where invalidity is measured: a whole-batch `resolved_at` mis-mapping. The recipe deliberately tells the orchestrator to map semantically ("the names below are only the _typical_ ones"), so a source returning `resolvedAt` / `closed_at` / `resolution.timestamp` mapped wrong leaves every record with `resolved_at === undefined` — which `:131` classifies as still-open, not invalid. You get count 40, resolved 0, invalid 0, and the note reads "incident source connected — no incident with a resolved recovery span in the window": coherent, alarming, and false, since it says 40 incidents are open. An `open_count === count && count > 0` callout naming the field mapping would catch it, same shape as the code-host thin-fetch guard at `audit_core.ts:607-619`. Fine as a follow-up if you'd rather not widen this one.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
AlexanderMakarov: The clamp is here and the collector-level test pins it, so the shape is right. Two things I'd still look at, both about where the window comes from rather than whether it's applied.

The day count is `period.lookback_days` off the artifact — which the recipe has the orchestrator author. Every sibling takes it from standards instead: `ci_pass_rate.ts:91` and `pipeline_duration.ts:74` both call `lookbackDays(standards)` (`[meta].max_lookback_days`, default 90), and `clampToWindow`'s own docstring says records "must measure the last `[meta].max_lookback_days` only". Here all three call sites read the artifact — `mttr.ts:84-88`, `audit_core.ts:127`, and the gate at `:652-657` — and `clampIncidentsToWindow` short-circuits on a falsy value, so an artifact written without a `period` gets no clamp at all, and an all-time PagerDuty export becomes the "audit-window median" at maximal/HIGH. The irony is that `collect()` is the one call site that does get an engine-controlled period, and it's the one that never sees real data, since the orchestrator overwrites the file wholesale. `mttr.ts` already has `standards` in hand.

Related: nothing reports what the clamp dropped. `deriveIncidentAggregates` computes `dropped_out_of_window` and it has exactly one write and no reads — not in `IncidentsRaw`, not in `collect()`'s raw block, not in the reliability note. So 480 incidents clamping to 11 reads exactly like a source that only ever had 11. `_base.ts:89`'s `windowDropNote` is the existing shape for this (it hardcodes the noun "run", so it'd want parameterising), and `invalid_count` got precisely this treatment last round. Nice side effect: adding it to `IncidentsRaw` makes your new lint guard force it into the recipe.

Last thing — the clamp is unpinned everywhere it matters. Deleting the `lookbackDays` argument at `mttr.ts:86`, `audit_core.ts:116` and `:654` leaves every test green, because `tests/helpers.ts:208` writes `lookback_days: 730` into every `writeCollected` envelope, so no fixture can fall outside a window. Only the collector-level clamp is covered, and incidents isn't in `window-clamp.test.ts`.

tests/lint-prompts.test.js
AlexanderMakarov: The word-boundary fix and `IncidentsRaw` both landed — that was the vacuous part. One thing left: the regex still tests against the whole of `connector-shapes.md`, and six of the eleven guarded fields (`id`, `resolved_at`, `source`, `source_label`, `count`, `resolved_count`) appear as standalone identifiers in the tracker/CI/code-host recipes, so they're still satisfied from elsewhere in the file. Deleting both the `count` and `resolved_count` bullets from the incidents recipe leaves the test passing. Slicing `doc` from the `## Incidents` heading to EOF before matching closes it.

plugins/awos/skills/ai-readiness-audit/SKILL.md
```diff
@@ -167,14 +167,14 @@ node "${CLAUDE_SKILL_DIR}/dist/cli.js" patch-judgment "context/audits/YYYY-MM-DD
...
+   - `headline` — the executive band. Transcribe values verbatim from the dimension checks (cite the `check_id`); never invent numbers. Row 1 of the headline (capability Points + Coverage cap-score block) is emitted by the renderer directly from `audit_total`/`coverage` — do not add it as a `delivery[]` entry, and the two connector-gated rows (Cycle time In-Progress→Done from the tracker artifact, MTTR from the incidents artifact) are computed by the engine and appended by the renderer — do not author them either (an authored gated row is ignored). `delivery[]` carries only rows 2–7, each a `DeliveryMetric` object `{label, display_value?, band?, check_id?}`. Author them in this order, reading DORA bands from each check's `hint` field ("DORA-banded (high)"), and transcribing all values verbatim — never invent numbers:
```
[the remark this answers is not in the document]
AlexanderMakarov: Half of this is right and worth doing here. `output-format.md:77` does say "SKILL.md Step 6.4"; SKILL.md has no Step 6.4 — Step 6 is "What's Next?" and report authoring is Step 5 item 4, which `SKILL.md:98` itself writes as "(Step 5.4)". This PR edits that exact line, so it's a one-word fix in passing, and it's the only occurrence left in the repo.

The `SKILL.md:170` half is a mis-anchor — that line carries no step reference.

plugins/awos/skills/ai-readiness-audit/SKILL.md
```diff
@@ -167,14 +167,14 @@ node "${CLAUDE_SKILL_DIR}/dist/cli.js" patch-judgment "context/audits/YYYY-MM-DD
...
      1. **Merges** — put the unit in the value, not the label: `label: "Merges"`, `display_value` from `report-context` → `window_stats.merges_per_active_per_week` rendered as a per-week rate `"<n> / week (per active contributor)"` (e.g. `"1.5 / week (per active contributor)"`); no `band`; no `check_id`; source: git artifact. If the value is null (zero active contributors), omit `display_value`.
```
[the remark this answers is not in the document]
AlexanderMakarov: The mismatch is real: `SKILL.md:171` asks for `"<n> / week (per active contributor)"` while the `output-format.md` example still shows `"3.2 / active contributor"`. Both lines are pre-existing and untouched by this PR, though, so I'd leave it out of this one rather than widen the diff — worth its own small commit.

AlexanderMakarov (commented):
Everything from the last round is addressed — thanks for the turnaround, and particularly for taking the derivation into the engine instead of patching around it. `deriveIncidentAggregates` as the single place the median is computed, shared by the collector, the metric and the topology gate, is the right shape: awarded and measured can't drift apart now.

I ran the audit from this branch against a real 8-repo org (provectus-barhopping), and the low-risk claim holds exactly as you described: every repo produced `incidents.json` with `available: false`, DF-07 SKIP at weight 0, and `derived_delivery.mttr: {}` — byte-for-byte the old behavior on git-only repos. Driving a records-only `incidents.json` through `enrich` on the shipped `dist/` then took DF-07 from SKIP to `PARTIAL`, weight 3.5/5, `median 2.0h MTTR over 2 incidents (high)`, maximal/HIGH. The engine half does what it says.

The two things I'd want resolved before merge are both downstream of that, in the parts that show the number to a human. I reproduced both rather than reading them off the diff.

**The org rollup silently drops the measured MTTR.** `metrics/rollup_input.ts:166` is `delivery.mttr = gatedDisplay('incident')`, which reads the authored headline row — the one the orchestrator is told never to author (confirmed on the real run: every authored row came back `gated: null`) and the renderer drops. Line 163 directly above gives cycle time the `audit.derived_delivery?.cycle_time?.display_value ?? gatedDisplay('tracker')` precedence, under a comment saying engine-derived rows win, "same precedence as the renderer, same artifact truth". Giving one repo both gated rows in identical shape and running `rollup` over all 8:

```
{"repo":"barley","cycle_time":"3.4 d","mttr":null}
```

Cycle time carries through; MTTR doesn't. So in org mode a repo with a real incident connector shows the em-dash in the Repositories table while its own per-repo report shows `2 h` — the headline-contradicts-detail failure this feature exists to kill, one level up. That line was correct before this PR, since there was no derived MTTR to prefer. One line, plus a test mirroring the cycle-time one.

**Doc debt radiating from the disjunct removal.** Dropping the tracker `incident_source` disjunct was the right call, but a fair number of untouched sentences still describe the old contract, and several are reader-facing. Worth one sweep commit: `connector-shapes.md:73` and `collectors/tracker.ts:26-28`/`:90-93` all still promise that declaring `incident_source` upgrades MTTR reliability "from git-proxy to first-class"; `SKILL.md:155` still tells the orchestrator to fetch it "for the gated rows" and `audit_patch.ts:490` still hands it to the narrative author unlabelled; `references/standards.md:40`'s `sources` enum still lists `"incident"` after this PR renamed the token to `"incidents"`, so a category written to that doc would fall out of both `CONNECTABLE_SOURCES` and `COLLECTED_ARTIFACT_SOURCES`; `render.ts:1174-1178`'s `MTTR_TIP` tells the reader the git proxy "scores DF-07 separately" when it never awards 1103; and `dimensions/project-topology.md:107` still frames `has_incident_source` as repo-detected from "a PagerDuty reference" — precisely the disjunct you removed. The loudest is `references/ai-sdlc-metrics-catalog.md:58,93`: "MTTR (mttr) is SKIP-by-default. Do not compute MTTR unless…", against `mttr.ts`'s "SKIP: never", with the collector column still pointing at `collectors/tracker.ts`. The same-file one (`dimensions/delivery-flow.md:16`) is on its thread.

Smaller points are inline or on the existing threads. Two non-code notes: the branch is conflicting with main (~11 hunks — #168 and #169 landed), and Quality Check hasn't run against `593e7c0d` at all; its last run on this branch is against `7cbf6b6f`, so CodeQL is the only check on the current head. The suite does pass at head locally with `dist/` clean, so I read this as the conflict blocking the merge-ref build rather than anything wrong — a rebase should bring CI back.

plugins/awos/skills/ai-readiness-audit/references/connector-shapes.md
AlexanderMakarov: Landed in ebf52aa.

1. `source_label` is now in the `period` block of the envelope example alongside `raw.source_label`, with a line saying which one feeds where — `period.source_label` for the Connections & Sources column, `raw.source_label` for the MTTR note. Removes the case where the Sources column says "incident source" while the DF-07 note says "PagerDuty".
2. Corrected: the clamp is anchored to the newest incident's own `started_at`, matching `clampToWindow`, and the doc now spells out the consequence — an all-stale stream is kept in full and measured against its own newest record, not excluded. The window length comes from `[meta].max_lookback_days`, not the `period` the orchestrator writes (see the thread on the artifact-period read).
3. Fixed — a record with no `resolved_at` is still-open and never counted invalid. Matches `deriveIncidentAggregates` and the `IncidentsRaw` docstring, which already had it right.

plugins/awos/skills/ai-readiness-audit/audit_core.ts
AlexanderMakarov: Both follow-ons are in.

The stale contract is swept in ebf52aa: `connector-shapes.md:73` and both `collectors/tracker.ts` docstrings now describe `incident_source` as a provenance label that neither upgrades MTTR reliability nor awards 1103 — only a measurable recovery span in `collected/incidents.json` does. The rest of the list from your summary went with it.

The weak pin is tightened in d182f13: `assert.equal(after!.weight_awarded, 0)`. You were right that the `assert.ok(!(x > 0))` form also passed on `undefined`, a missing field, and `NaN`.

plugins/awos/skills/ai-readiness-audit/dimensions/delivery-flow.md
AlexanderMakarov: Both lines corrected in ebf52aa — `dimensions/delivery-flow.md:16` and `dimensions/ai-sdlc-adoption.md:13`. The proxy is described as branch-lifetime over all first-parent merges rather than revert/hotfix cadence, and the reliability upgrade now requires at least one measurable resolved span instead of the artifact merely being present. Same wording in both so they cannot drift apart again.

plugins/awos/skills/ai-readiness-audit/metrics/mttr.ts
AlexanderMakarov: Took the second option, in e7cf4b5 — the tracker is out of `sources_used` on the proxy path, including the git-missing branch where it was the sole entry behind an absent value. The `incident_source` label still adds its informational note, but it no longer credits a source for a number it did not contribute to, which lines up with `sources = ["incidents"]` for 1103. The Source-shapes header block says so explicitly now.

`tests/met-mttr.test.ts` had pinned the old behaviour (`sources_used must include tracker when incident_source declared`); it now asserts the tracker is not credited.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
AlexanderMakarov: Folded it in rather than deferring — e86f9a8.

Two parts: an audit-core artifact warning when every in-window incident is still open (same shape as the code-host thin-fetch guard you pointed at), and the MTTR reliability note now names the likely `resolved_at` mapping miss instead of asserting that N incidents are open. The old generic branch turned out to be unreachable — `resolved_count` is always 0 by the time that block runs — so it is a clean two-way split now.

Pinned in `tests/audit_core.test.ts` and `tests/met-mttr-incidents.test.ts`. Say the word if you would rather this moved out to keep the diff tight.

plugins/awos/skills/ai-readiness-audit/collectors/incidents.ts
AlexanderMakarov: All three, across 6ee9f9f and 55a66d0.

**Window source.** All three call sites now use `lookbackDays(standards)` — `mttr.ts`, `computeDerivedDelivery`, and the topology gate — matching `ci_pass_rate` and `pipeline_duration`. Standards is threaded into `computeDerivedDelivery`, and through `aggregate`/`patchJudgments` from the CLI so the repair verbs use the same window. The period-less envelope case you identified was the sharp one: the clamp short-circuits on a falsy lookback, so that path had no clamp at all.

**Reporting the drop.** `dropped_out_of_window` is on `IncidentsRaw`, set in `collect()`, and surfaced in the reliability note via `windowDropNote`, which now takes the record noun instead of hardcoding "run". And yes — the lint guard forced it into the recipe, exactly as you predicted.

**Pinning.** `window-clamp.test.ts` covers incidents and mttr now, with an explicit narrow lookback rather than the 730-day default `helpers.ts` bakes into every envelope. I checked the cases actually bite: reverting any of the three call sites to the artifact period turns them red. One of my first drafts passed either way — median of `[2, 2, 200]` is still 2 — so the fixture now uses three ancient incidents to move the median.

tests/lint-prompts.test.js
AlexanderMakarov: Fixed in d182f13 — `doc` is sliced from the `## Incidents` heading to EOF before matching, so the guard can only be satisfied from within the incidents recipe. Verified it now fails when the `count` or `resolved_count` bullet is deleted, which was the case that stayed green before.

plugins/awos/skills/ai-readiness-audit/SKILL.md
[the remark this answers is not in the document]
AlexanderMakarov: Fixed the valid half in ebf52aa — `output-format.md:77` now reads "SKILL.md Step 5.4". The `SKILL.md:170` anchor carried no step reference and needed no change, as confirmed above.

plugins/awos/skills/ai-readiness-audit/render.ts
AlexanderMakarov: Both halves landed in c8e86e1, plus the docstring.

`DerivedDelivery.mttr` carries `band` and `check_id: "DF-07"` now, set in `computeDerivedDelivery` and spread in `normalizedDelivery`, so the row bands like its siblings instead of being the one blank cell in a block headed "vs DORA bands". Took the belt-and-braces option for the caveat: the sample size is folded into the value (`1 h (1 of 3)`) when part of the sample was unusable, so a median over 3 of 60 resolved incidents no longer prints bare.

You were right that the line was entirely unpinned — the only `derived_delivery` fixture had `mttr: {}`. There is a measured-MTTR fixture now asserting value and band end to end; removing the band spread turns it red.

`computeDerivedDelivery`'s docstring is rewritten to name both artifacts (Cycle time from the tracker, MTTR from the incidents artifact, computed independently so it fills with no tracker at all), and `normalizedDelivery`'s header carried the same slip — fixed in the same pass.

plugins/awos/skills/ai-readiness-audit/metrics/mttr.ts
AlexanderMakarov: Fixed in bb94640 — the note is set into both reliability objects in the git-missing branch. Narrow path, but as you say it is the one the note exists for: with git gone there is nothing else in the report to explain why MTTR is empty. Two tests cover it, one per leg of that branch.

plugins/awos/skills/ai-readiness-audit/references/standards.toml
AlexanderMakarov: Changed in ebf52aa — "Without a measurable recovery span", which covers both the no-source case and the source-with-nothing-measurable case the gate treats alike. The longer definition pushed the report inline lead past the 320-char cap, so `[category.mttr]` gained a short `summary` to front it.

AlexanderMakarov:
Picked up the two remaining blockers from my last review plus the residuals on the open threads, pushed to this branch (`593e7c0..67ad62c`). Per-thread replies are inline; the summary-level items:

**Org rollup drops the measured MTTR** — b8e08e9. `delivery.mttr` now takes `audit.derived_delivery?.mttr?.display_value` first, matching the cycle-time precedence directly above it and the renderer. Pinned by a test on `readPerRepoAudit` mirroring the cycle-time one; it fails against the old line.

**Doc debt from the disjunct removal** — ebf52aa, one sweep. `connector-shapes.md:73`, `collectors/tracker.ts:26-28`/`:90-93`, `SKILL.md:155`, `audit_patch.ts:490` (the report-context key is `incident_source_label` now, so the narrative author sees what it is), `references/standards.md:40` (`"incident"` → `"incidents"`, and the enum was missing `code_host` too), `render.ts` `MTTR_TIP`, `dimensions/project-topology.md:107`, `references/ai-sdlc-metrics-catalog.md:58,93`, and `dimensions/delivery-flow.md:16` + `ai-sdlc-adoption.md:13`. The catalog was the loudest, as you said — it read as a direct instruction not to compute the thing this PR computes.

**Branch/CI** — merged `origin/main` in (e51b3d2). Only two files actually conflicted: `dist/cli.js`, settled by the rebuild, and `tests/lint-prompts.test.js`, where both sides had added a test at the same spot — kept both. Merged rather than rebased to avoid force-pushing a shared branch; say if you would rather have the linear history and I will rebase.

Also folded in the `open_count === count` callout you offered as a follow-up (e86f9a8) — it was small. Happy to split it out if you would rather keep the diff tight.

Verified locally at head: `npm test` green across all four layers (engine 1376/0, lint 151/0), `prettier --check` clean, and `dist/` reproducible — a fresh `npm run build:audit-engine` leaves `git diff --exit-code` empty.

plugins/awos/skills/ai-readiness-audit/audit_patch.ts
```diff
@@ -165,7 +170,8 @@ export function aggregate(outDir: string): void {
   // artifacts are gone.
   const derivedDelivery = computeDerivedDelivery(
     collectedDirAgg,
-    collectedAgg.get('tracker')!.art
+    collectedAgg.get('tracker')!.art,
+    standards
```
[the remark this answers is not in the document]
AlexanderMakarov: Holds, and it went further than "not reachable today" — fixed in 76fb79b.

The guard enumerated the fields that count as "the fresh derivation produced something": `cycle_time.display_value`, `cycle_time.note`, `mttr.note`. That enumeration never learned about the measured MTTR path, which sets `display_value`/`band`/`check_id` *instead of* `note` — so on a repo with an incidents connector and no tracker every listed field was absent and the fresh measurement was dropped for the stored one. The asymmetry was in the expression itself: cycle time's value was checked, MTTR's was not.

It is masked in the documented `enrich` → `patch-judgment` order, where the stored block is already the enriched one. It bites the standalone `aggregate` repair verb — the command CLAUDE.md points at after fixing a broken collector artifact, where the stored block is by definition pre-fix — and `patch-judgment`'s own `aggregate` call when `incidents.json` lands without a following `enrich`.

Rather than extend the enumeration, the guard now asks whether either row is empty, which cannot go stale as fields are added. A regression test in `tests/audit_core.test.ts` runs `aggregate` over a stale `derived_delivery` plus a measurable incidents artifact; it fails against the old guard. Separately, 2eb3f5b collapses the measured fields into one indivisible `mttr.measured` block so the renderer can no longer print a value beside an em-dash band.

plugins/awos/skills/ai-readiness-audit/collectors/tracker.ts
```diff
@@ -23,8 +23,10 @@ export interface TicketRecord {
 }
 
 /** Connector passed in by the caller when a project tracker integration is
- *  available. `incident_source` is optional — when present it upgrades MTTR
- *  reliability from a git-proxy estimate to a first-class measurement. */
+ *  available. `incident_source` is optional and purely a provenance label: it
+ *  names the incident system, but does not upgrade MTTR reliability and does
+ *  not award category 1103. Only a measurable recovery span in the incidents
+ *  artifact does that (see collectors/incidents.ts). */
```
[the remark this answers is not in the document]
AlexanderMakarov: Correct, and I had marked this done prematurely on another thread. The earlier sweep fixed the two *block-level* docstrings; both *field-level* ones survived. Fixed in e6f6d2c.

`:31-33` said the identifier "feeds MTTR" — it never feeds the value — and `:50-51` called it the source "for MTTR reliability upgrade", contradicted by `mttr.ts`'s own comment that `incident_source` presence does not upgrade reliability. Both now describe it as a provenance label that is reported, never scored; 1103 gates on `topology.has_incident_source`, which reads the incidents artifact.

Worth noting the structural gap you implied: the field-drift lint guard only diffs `connector-shapes.md` against field *names* in `collectors/incidents.ts`, so prose drift in `tracker.ts` has nothing watching it.

plugins/awos/skills/ai-readiness-audit/references/ai-sdlc-metrics-catalog.md
```diff
@@ -51,11 +51,11 @@ Tier C activates only when a CI connector or accessible CI logs can be resolved
...
+| work_mix_allocation | Work-mix allocation                            | Freed capacity shifts toward Growth, away from KTLO/Support — in **team-FTE share, never money** (DX Core 4 FTE-allocation model)                                            | `metrics/work_mix_allocation.ts` | `collectors/tracker.ts`   | `work-mix` (standards.toml) | banded |
+| issue_throughput    | Delivered-issue throughput & backlog burn-down | Output and debt-clearing                                                                                                                                                     | `metrics/issue_throughput.ts`    | `collectors/tracker.ts`   | `delivery-flow`             | raw    |
+| mttr                | MTTR                                           | Recovery speed — the metric never SKIPs (git proxy fallback); **category 1103 is awarded only on a measured recovery span**; do not infer from re-fixes of the same issue ID | `metrics/mttr.ts`                | `collectors/incidents.ts` | `delivery-flow`             | banded |
```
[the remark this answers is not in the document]
AlexanderMakarov: Agreed, and it is a real consequence of the change rather than a formatting point. Fixed in e6f6d2c.

The `mttr` row sat under "Tier I — issue tracker", whose activation line reads "Tier I activates when a Jira, Linear, or equivalent issue-tracker connector is resolvable", while this PR repoints its Collector column at `collectors/incidents.ts`. So a team with PagerDuty and no tracker measured MTTR while the catalog said their tier needed a tracker, and a tracker-only team read that they would get it and would not.

It now has its own "Tier N — incident source", mirroring Tier D, whose activation line states the independence explicitly: naming an incident system in the tracker is a provenance label, not incident data.

plugins/awos/skills/ai-readiness-audit/audit_patch.ts
```diff
@@ -487,7 +496,10 @@ export function reportContext(outDir: string): Record<string, unknown> {
       (git?.raw as Record<string, unknown> | undefined)?.window_stats ?? null,
     tracker_fetch_meta:
       (tracker?.raw as Record<string, unknown> | undefined)?.fetch_meta ?? null,
-    incident_source:
+    // Provenance label only — names the incident system the tracker declares.
+    // It does not produce the MTTR value (derived_delivery.mttr does, from the
+    // incidents artifact) and does not award category 1103.
+    incident_source_label:
       (tracker?.raw as Record<string, unknown> | undefined)?.incident_source ??
```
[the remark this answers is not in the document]
AlexanderMakarov: Right on the key name, and there was a larger half underneath it. Both fixed in 511abd6.

The rename stopped one keystroke short: `SKILL.md:155` still said `incident_source` while `reportContext()` emits `incident_source_label`. The sibling abbreviation in the same sentence drops a *prefix* (prose "tracker `fetch_meta`" for the key `tracker_fetch_meta`), so this one read as an error rather than shorthand.

The larger half: `reportContext()` sourced that label from `tracker.raw.incident_source` and never opened `collected/incidents.json` — it only reads git and tracker. Post-PR that inverts the signal. A project with a real PagerDuty connector and no tracker declaration handed the narrative author `null` while MTTR scored maximal; a project whose tracker merely names "opsgenie" with no incident data handed over `"opsgenie"` while 1103 stayed SKIP. The label the report actually shows comes from the incidents artifact's `source_label`, so that is now the first source, with the tracker's declaration as fallback. Pinned by a test that gives the two artifacts different labels.

plugins/awos/skills/ai-readiness-audit/dimensions/project-topology.md
```diff
@@ -104,7 +104,7 @@ The engine computes these topology flags deterministically (`topology.ts`) from
 - `has_docs_connector`: true|false (e.g. Confluence/Coda CLI on PATH or MCP server in session)
 - `has_hooks`: true|false (e.g. .claude/settings.json hooks or pre-commit hooks detected)
 - `has_http_api`: true|false (e.g. FastAPI, Express, Django REST, Spring MVC, or similar found)
-- `has_incident_source`: true|false (e.g. PagerDuty/OpsGenie reference or incident tracker detected)
+- `has_incident_source`: true|false (NOT repo-detected — true only when `collected/incidents.json` is available and carries at least one measurable resolved recovery span in the window)
```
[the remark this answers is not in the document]
AlexanderMakarov: The diagnosis holds and is now fixed, but not the way you suggested — e6f6d2c.

You are right that `ai-sdlc-adoption.md:19` tells the reader to extract `has_incident_source`, that the flag now means "`collected/incidents.json` is available and carries a measurable span", and that the setup block had no way to produce that file.

Adding a `collect incidents` line would have documented a command that does not exist: there is no `collect` verb in `cli.ts` at all, on this branch or on main — the dispatcher is `progress`/`render`/`rollup`/`audit-core`/`enrich`/`aggregate`/`patch-*`/`report-context`. The whole block was already describing the retired per-dimension fan-out. So the block is deleted rather than extended; scoring runs through one `audit-core` pass.

One thing worth surfacing: the lint suite *required* that reference ("body must reference the collect engine command"), so the fiction was being enforced. It now forbids it instead. The per-check `How:` lines in the same file still cite an equally nonexistent `metric` verb — ten occurrences, pre-existing and unrelated to DF-07, so left for their own change rather than widened into this one.

tests/lint-prompts.test.js
```diff
@@ -4077,6 +4077,61 @@ test('spec.md captures boundary/error behavior as rules in items 2 and 3', () =>
...
+    new RegExp(`(?<![A-Za-z0-9_])${f}(?![A-Za-z0-9_])`).test(section);
+  const missing = [...new Set(fields)].filter((f) => !documented(f));
+  assert.deepEqual(
+    missing,
+    [],
+    `connector-shapes.md must document these incidents fields (drifted from incidents.ts): ${missing.join(', ')}`
+  );
+});
+
 test('every artifact-producing command writes before review, never gating the write on a reply', () => {
```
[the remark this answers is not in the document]
AlexanderMakarov: Skipping this one. Both lines are pre-existing and only moved down by the block inserted above them, and Prettier already accepts them as written — it hugs a last-argument callback rather than breaking an unbreakable string literal, which is why the untouched siblings nearby (113 and 115 characters) stand too. `npx prettier . --check` is CI-enforced and green at the current head, so there is nothing to change here.

AlexanderMakarov (approved):
Both real-life checks pass. Full picture for PR 170's round-5 fixes:

**1. Audit harness** (full `/awos:ai-readiness-audit`, live `claude -p`, PR-170 worktree code, target = real awos repo):
- Compliant: model ran the engine (`audit_core_calls=1`, zero fan-out spawns), judgments patched, `audit_total=255.4`, 8m52s, $5.18.
- Incident behavior on a repo with no incident source: `incidents.json` `available: false`, DF-07 SKIP with the honest git-proxy note — unchanged, as it should be.
- Archive: `/Users/aleksandrmakarov/code/awos-pr-170/tmp/audit-runs/awos/20260824T141849Z__awos-5a3b666/audit-output/report.html`

**2. awos-qa e2e** (`audit-connectors-incident` — mock Jira incident MCP, `AWOS_REPO` = PR-170 worktree): **PASS 8/8**, 4m47s, $2.64 — engine compliance, reports produced, connectors fetched, scores within baseline tolerance, and the incident-fed measured path fired for real: `derived_delivery.mttr.measured = {median_hours: 4.5, incidents_used: 3, band: high}` with `MTTR | 4.5 h` in the rendered report.
- Archive: `/Users/aleksandrmakarov/code/awos-qa/tmp/audit-runs/audit-connectors-incident/2026-08-24T14-33-12-040Z/report.html`

**3. Degraded-state wording** (the strings this round changed) was proven earlier against the real renderer via crafted artifacts — all-open/mapping-miss, quiet-quarter, shape warning, period-only label, 90-day window clamp — in `…/scratchpad/pr170-validation/`.

PR #122, opened 2026-05-22
feat: add change-request skill

tigra:
## What

Adds an opt-in `/awos:change-request` command that captures **why** a previously-agreed requirement changed — the trigger, the driver, and its business-level impact on already-implemented work — and saves a numbered record under `context/change-requests/`.

## Files

- `commands/change-request.md` + `claude/commands/change-request.md` wrapper
- `templates/change-request-template.md`
- Integration hooks (opt-in):
  - `commands/product.md`, `commands/roadmap.md` — offer a CR before writing the artifact
  - `commands/spec.md` — offer a CR when a new spec revises a prior requirement

## Backward compatibility

All hooks are **opt-in**: the skill opens with an `AskUserQuestion` skip option, so existing `product` / `roadmap` / `spec` flows are unchanged unless the user chooses to log a CR.

## Validation

- `npm test` → 69/69 pass (static prompt lint + installer + fixtures)
- `npx prettier --check .` clean

Suggested release-drafter label: **minor** (new feature skill).

## Summary by CodeRabbit

* **New Features**
  * New Change Request workflow for documenting and tracking revisions to agreed requirements.
  * Optional prompts to log a change request during product definition, roadmap finalization, and spec drafting.
  * Guided, single-question interview flow with skip/defer options to create or stub a request.
  * Structured change-request template capturing trigger, rationale, nature, impacts on existing requirements, decisions, and follow-ups.

PR #171, opened 2026-07-31
feat: /awos:quick — ad-hoc task shortcut outside the spec cycle

kdementiev-provectus:
## Why

AWOS gives AI agents full project context so they produce correct code — but the full workflow (`spec → tech → tasks → implement → verify`) takes five steps. That's the right investment for a roadmap feature, but overkill for everyday work: fixing a bug, tweaking a config, or doing a small refactor.

Without a lighter option, teams face an awkward choice: run the full ceremony for a 20-minute fix, or bypass AWOS entirely and lose the quality guarantees (a written plan, a matched specialist, alignment with architecture).

`/awos:quick` is the middle path. One command that still plans before coding, still picks the right specialist agent, and still reads the project's product definition and architecture — but skips the layers a small task doesn't need.

## Summary

- Adds `/awos:quick` — a single command for contained, one-off work that doesn't justify a full spec cycle.
- Plans the task, delegates coding to a specialist subagent (same pool as `/awos:implement`), and records `PLAN.md` + `SUMMARY.md` under `context/quick/`.
- Reads existing AWOS context files (`product-definition.md`, `architecture.md`, `roadmap.md`) before execution — so even quick tasks stay aligned with the project's design decisions.
- Includes `list`, `status <slug>`, and `resume <slug>` subcommands for task management.
- Separate from `context/spec/` and the roadmap — quick tasks never pollute the feature backlog.

## Usage examples

```bash
# Execute a contained task (default depth — plan and execute):
/awos:quick Fix the Docker rate-limit failure by using the Artifactory proxy

# Investigate before planning (choose "Research first" when asked):
/awos:q Add request retry with backoff to the API client

# Run with full depth — discuss, research, execute, validate:
/awos:q Migrate user sessions from Redis to PostgreSQL

# List all quick tasks with their status:
/awos:q list

# Check status of a specific task:
/awos:q status fix-docker-rate-limit

# Resume an incomplete task:
/awos:q resume add-request-retry
```

## What's included

| File | Purpose |
|------|---------|
| `commands/quick.md` | Full prompt — orchestrator logic |
| `claude/commands/quick.md` | Thin wrapper (user-editable customization layer) |
| `claude/commands/q.md` | `/awos:q` shortcut — points to the same prompt |
| `docs/commands/quick.md` | User-facing command documentation |
| `docs/rationale.md` | Philosophy section: when to use quick vs full cycle |
| `README.md` | Quick Tasks section added to the usage guide |
| `CLAUDE.md` | Internal docs updated: quick-task shortcut + delegation rule |

## Design decisions

- **Orchestrator-only contract** — like `/awos:implement`, it never writes code itself. All changes are delegated to specialist subagents.
- **Context-aware execution** — when AWOS context files exist, the subagent reads them before starting. If the project has no context yet, the step is skipped gracefully.
- **Depth is a question, not a flag** — asks the user up front how thorough to be (Plan and execute / Discuss first / Research first / Full) instead of requiring memorized CLI flags.
- **DRY validation** — before planning, checks whether the task parallels existing structure and defaults to reuse over duplication.
- **Explicit commit control** — never commits without the user's explicit choice; skips the question entirely in non-git projects.

## Summary by CodeRabbit

* **New Features**
  * Added `/awos:quick` for handling focused, one-off tasks through planning, specialist execution, optional validation, and commit confirmation.
  * Added `/awos:q` and `/quick` shortcuts.
  * Added `list`, `status`, and `resume` task-management commands.
  * Quick tasks now retain lightweight plans and summaries for later review or continuation.

* **Documentation**
  * Added usage guidance, workflow details, examples, and recommendations for choosing quick tasks versus full workflows.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -102,7 +102,9 @@ The canonical flow (each command is a markdown prompt under `commands/`):
 
 The first four are run once at project setup; the last five iterate per feature. Each command reads/writes a specific document under `context/` (e.g. `context/product/product-definition.md`, `context/spec/NNN-feature/tasks.md`). The numeric prefix on spec directories is allocated by `scripts/create-spec-directory.sh`.
 
-**Implementation delegation rule:** `/awos:implement` is an orchestrator only — it reads `tasks.md`, extracts the `**[Agent: name]**` marker from each task, and delegates to a subagent. The orchestrator is explicitly prohibited from editing code itself. Preserve this contract when editing `commands/implement.md`.
+**Quick-task shortcut:** `/awos:quick` sits alongside the canonical flow for contained, one-off work that doesn't justify a full spec. It plans a single task, delegates to a specialist subagent, and writes `PLAN.md`/`SUMMARY.md` under `context/quick/NNNNNNNN-slug/` — separate from `context/spec/` and never tracked in the roadmap. It shares `/awos:implement`'s orchestrator-only contract (delegates all code changes) and offers `list`/`status`/`resume` subcommands.
+
+**Implementation delegation rule:** `/awos:implement` is an orchestrator only — it reads `tasks.md`, extracts the `**[Agent: name]**` marker from each task, and delegates to a subagent. The orchestrator is explicitly prohibited from editing code itself. Preserve this contract when editing `commands/implement.md`. `/awos:quick` follows the same rule.
 
 ## Architecture: Installer Pipeline
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ For contained, one-off work that needs a plan and a specialist but not a full sp
 | `/awos:quick` | Executes a small ad-hoc task end-to-end — plans, delegates, summarizes. | [Details](docs/commands/quick.md) |
 | `/awos:q`     | Shortcut — identical to `/awos:quick`.                                  | [Details](docs/commands/quick.md) |
 
-`/awos:quick` plans the task, delegates the coding to a specialist subagent (the same specialists `/awos:implement` uses), and records a `PLAN.md` and `SUMMARY.md` under `context/quick/` — separate from the roadmap. When AWOS context files exist (product definition, architecture, roadmap), the subagent reads them before starting — so even quick tasks stay aligned with the project's design. It asks how thorough to be (plan-only, discuss, research, or full) and asks before committing. Use `/awos:quick list`, `status <slug>`, and `resume <slug>` to manage accumulated tasks.
+`/awos:quick` plans the task, delegates the coding to a specialist subagent (the same specialists `/awos:implement` uses), and records a `PLAN.md` and `SUMMARY.md` under `context/quick/` — separate from the roadmap. When AWOS context files exist (product definition, architecture, roadmap), the subagent reads them before starting — so even quick tasks stay aligned with the project's design. It asks how thorough to be (Plan and execute, Discuss first, Research first, or Full) and asks before committing. Use `/awos:quick list`, `status <slug>`, and `resume <slug>` to manage accumulated tasks.
 
 ### Step 4: You're Awesome
 
--- a/README.md
+++ b/README.md
@@ -60,8 +60,9 @@ For contained, one-off work that needs a plan and a specialist but not a full sp
 | Command       | What it does                                                            | Docs                              |
 | ------------- | ----------------------------------------------------------------------- | --------------------------------- |
 | `/awos:quick` | Executes a small ad-hoc task end-to-end — plans, delegates, summarizes. | [Details](docs/commands/quick.md) |
+| `/awos:q`     | Shortcut — identical to `/awos:quick`.                                  | [Details](docs/commands/quick.md) |
 
-`/awos:quick` plans the task, delegates the coding to a specialist subagent (the same specialists `/awos:implement` uses), and records a `PLAN.md` and `SUMMARY.md` under `context/quick/` — separate from the roadmap. It asks how thorough to be (plan-only, discuss, research, or full) and asks before committing. Use `/awos:quick list`, `status <slug>`, and `resume <slug>` to manage accumulated tasks.
+`/awos:quick` plans the task, delegates the coding to a specialist subagent (the same specialists `/awos:implement` uses), and records a `PLAN.md` and `SUMMARY.md` under `context/quick/` — separate from the roadmap. When AWOS context files exist (product definition, architecture, roadmap), the subagent reads them before starting — so even quick tasks stay aligned with the project's design. It asks how thorough to be (plan-only, discuss, research, or full) and asks before committing. Use `/awos:quick list`, `status <slug>`, and `resume <slug>` to manage accumulated tasks.
 
 ### Step 4: You're Awesome
 
--- a/README.md
+++ b/README.md
@@ -53,6 +53,16 @@ Once your foundation is set, iterate through this cycle for each feature on your
 
 > **When to skip the cycle**: Not every change needs a spec. Hotfixes, simple bugfixes, and small edits don't require the full spec workflow — Claude Code's built-in plan mode handles those just fine.
 
+### Quick Tasks
+
+For contained, one-off work that needs a plan and a specialist but not a full spec — a focused bugfix, a small refactor, a config or dependency change — use the shorter path:
+
+| Command       | What it does                                                            | Docs                              |
+| ------------- | ----------------------------------------------------------------------- | --------------------------------- |
+| `/awos:quick` | Executes a small ad-hoc task end-to-end — plans, delegates, summarizes. | [Details](docs/commands/quick.md) |
+
+`/awos:quick` plans the task, delegates the coding to a specialist subagent (the same specialists `/awos:implement` uses), and records a `PLAN.md` and `SUMMARY.md` under `context/quick/` — separate from the roadmap. It asks how thorough to be (plan-only, discuss, research, or full) and asks before committing. Use `/awos:quick list`, `status <slug>`, and `resume <slug>` to manage accumulated tasks.
+
 ### Step 4: You're Awesome
 
 That's it! By following these steps, you can systematically turn your vision into a well-defined and fully implemented product.
--- a/commands/q.md
+++ b/commands/q.md
@@ -1,0 +1,21 @@
+---
+description: Shortcut for /awos:quick — executes a small ad-hoc task end-to-end.
+---
+
+# ROLE
+
+This is an alias for `/awos:quick`. All behavior, rules, and process steps are identical.
+
+---
+
+# INTERACTION
+
+- Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+
+---
+
+# PROCESS
+
+Follow the full process defined in `.awos/commands/quick.md` — this command is identical in every way.
+
+Read the file `.awos/commands/quick.md` and execute it exactly as written, passing through the user's prompt unchanged.
--- a/commands/quick.md
+++ b/commands/quick.md
@@ -94,7 +94,16 @@ If depth is "Research first" or "Full":
 - The research prompt should ask: what is the best approach for this task in _this_ codebase, which existing files/patterns to follow, and what pitfalls to avoid.
 - Fold the findings into `PLAN.md`.
 
-### Step 5: Write the Plan
+### Step 5: Choose the Specialist Subagent
+
+Determine which subagent should do the coding, reusing the same specialists as `/awos:implement`:
+
+- Enumerate the available subagents by inspecting the `Agent` tool's description block in your own system prompt (introspection — no tool call needed). Both project-local agents (declared as files under `.claude/agents/*.md`) and plugin-provided agents (recognized by the `plugin-name:` prefix on `subagent_type`) are listed there.
+- Match the task to a subagent by technology and intent (e.g. FastAPI/Python → a Python specialist, React/UI → a frontend specialist, DB → a database specialist, infra/CI → an infrastructure specialist).
+- If no specialist clearly matches — plain config edits, documentation, shell scripts, and other generic one-off work — use `general-purpose`. Don't force a stack specialist onto a task outside its domain just because the verification happens to use its language.
+- You'll record the choice in `PLAN.md` in the next step as `**[Agent: agent-name]**`.
+
+### Step 6: Write the Plan
 
 Write `context/quick/[YYYYMMDD]-[slug]/PLAN.md` with:
 
@@ -102,18 +111,9 @@ Write `context/quick/[YYYYMMDD]-[slug]/PLAN.md` with:
 - **Approach** — how it will be done (incorporating any discussion/research findings).
 - **Steps** — the concrete edits/actions, in order.
 - **Definition of Done** — how success is verified (tests, lint, typecheck, curl, manual check).
-- **Agent** — the specialist chosen in Step 6, recorded as `**[Agent: agent-name]**`.
+- **Agent** — the specialist chosen in Step 5, recorded as `**[Agent: agent-name]**`.
 
 Keep it short — a quick task's plan is a checklist, not a document.
-
-### Step 6: Choose the Specialist Subagent
-
-Determine which subagent should do the coding, reusing the same specialists as `/awos:implement`:
-
-- Enumerate the available subagents by inspecting the `Agent` tool's description block in your own system prompt (introspection — no tool call needed). Both project-local agents (declared as files under `.claude/agents/*.md`) and plugin-provided agents (recognized by the `plugin-name:` prefix on `subagent_type`) are listed there.
-- Match the task to a subagent by technology and intent (e.g. FastAPI/Python → a Python specialist, React/UI → a frontend specialist, DB → a database specialist, infra/CI → an infrastructure specialist).
-- If no specialist matches, use `general-purpose`.
-- Record the choice in `PLAN.md` as `**[Agent: agent-name]**`.
 
 ### Step 7: Delegate Execution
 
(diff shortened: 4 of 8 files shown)


commands/quick.md
```diff
@@ -0,0 +1,241 @@
...
+
+# PROCESS
+
+## Step 0: Parse Arguments and Route
+
+Inspect `<user_prompt>` and route:
+
+- Starts with `list` → go to **LIST**.
+- Starts with `status ` → the remainder is a slug → go to **STATUS**.
+- Starts with `resume ` → the remainder is a slug → go to **RESUME**.
```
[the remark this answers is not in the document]
kdementiev-provectus: Intentional — the trailing space inside the code span is the literal prefix the agent matches against (e.g. the input "status fix-docker" starts with "status " including the space). Moving the space outside would change the 
matching semantics. This is a prompt instruction, not rendered documentation — MD038 doesn't apply here.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+- **Agent** — the specialist chosen in Step 5, recorded as `**[Agent: agent-name]**`.
+
+Keep it short — a quick task's plan is a checklist, not a document.
+
+### Step 6.5: Gather Project Context
+
+Check which AWOS context files exist in the project:
+
+- `context/product/product-definition.md`
+- `context/architecture/architecture.md`
```
AlexanderMakarov, line 143: This path never resolves — the architecture doc lives at `context/product/architecture.md` (that's what `commands/architecture.md`, `tech.md`, and `hire.md` all read and write; nothing in the repo creates `context/architecture/`). As written, the existence check silently fails and the subagent never sees the architecture, which quietly defeats the "stays aligned with the project's design" promise the README makes for this command. RESUME step 5 reuses this list, so it inherits the same miss.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+    - **Discuss first** — clarify the approach and surface assumptions before planning.
+    - **Research first** — investigate approaches, libraries, and pitfalls before planning.
+    - **Full** — discuss, then research, then plan, execute, and validate afterward.
+3.  Record the chosen depth. It enables: Discussion (Step 3), Research (Step 4), and post-execution Validation (Step 8). "Plan and execute" enables none of them.
+
+### Step 2: Create the Task Directory
+
+1.  Generate a slug from the task description (per Slug rules).
+2.  Compute today's date: `date +%Y%m%d`.
+3.  Check for collisions: `ls -d context/quick/*-[slug]/ 2>/dev/null`. If a match exists, ask the user whether to resume it (go to **RESUME**) or append a numeric disambiguator (`-2`, `-3`, etc.) to the slug until unique.
```
AlexanderMakarov, line 80: Two things on this collision check:

1. The glob `context/quick/*-[slug]/` isn't anchored to the date prefix, so `*` can eat across slug boundaries — with an existing `20260101-fix-auth-bug`, creating `auth-bug` false-fires a collision, and `status auth-bug` / `resume auth-bug` (lines 230 and 240 use the same pattern) silently return the wrong task instead of "not found". `context/quick/????????-[slug]/` (exactly eight date characters) fixes all three sites.
2. The resume-vs-disambiguate choice here is a fixed two-option ask written as prose — elsewhere in this file (depth, commit, re-run) you route those through `AskUserQuestion`; this one could match.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+
+Given a sanitized slug:
+
+1.  Find the directory: `ls -d context/quick/*-[slug]/ 2>/dev/null | tail -1`. If none, print `No quick task found with slug: [slug]` and stop. (`tail -1` picks the most recent match.)
+2.  Read `PLAN.md` (goal, approach, steps, agent) and `SUMMARY.md` (if present, for what's done). If `PLAN.md` is missing, inform the user that this task has no plan to resume and suggest starting fresh with `/awos:quick [task description]`. Stop.
+3.  If `SUMMARY.md` exists with `status: complete`, warn the user the task is already done and ask (via `AskUserQuestion`) whether to re-run it or stop.
+4.  Announce what you're resuming and the current status.
+5.  Run Step 6.5 (Gather Project Context) to check for AWOS context files — resumed tasks still need project alignment.
+6.  Extract the agent name from `PLAN.md`'s `**[Agent: agent-name]**` field. If the field is missing, fall back to Step 5 (Choose the Specialist Subagent) before delegating.
+7.  Continue from where the task left off: re-enter the RUN flow at Step 7 (Delegate Execution) using the existing `PLAN.md`, passing the subagent the plan plus a note of what remains. Then finish with Steps 8–11.
```
AlexanderMakarov, line 246: Step 8 is gated on "If depth is Full", but depth only exists in Step 1's interactive choice — the PLAN.md schema doesn't persist it, so when RESUME re-enters here and "finishes with Steps 8–11" there's nothing to evaluate, and a "Full" run that gets resumed silently loses its validation pass. Persisting a Depth field in PLAN.md next to the Agent field (and reading it back in RESUME), or an explicit resume default, would close it.

commands/q.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+
+---
+
+# PROCESS
+
+Follow the full process defined in `.awos/commands/quick.md` — this command is identical in every way.
+
+Read the file `.awos/commands/quick.md` and execute it exactly as written, passing through the user's prompt unchanged.
```
AlexanderMakarov, line 21: This file never binds `$ARGUMENTS` — every other root command anchors user input with `<user_prompt>$ARGUMENTS</user_prompt>` (`implement.md:19`, `spec.md:29`, `tasks.md:21`…). Without the placeholder, the host appends the typed text to the end of the prompt, and the runtime Read of quick.md then surfaces a literal, unsubstituted `$ARGUMENTS` for Step 0 to route on — it mostly works by model goodwill. A one-line `- **User Prompt:** <user_prompt>$ARGUMENTS</user_prompt>` makes it mechanical.

Separate question, genuinely asking: is the runtime-Read alias (vs the `@.awos/commands/…` import the wrappers use) deliberate — nested `@`-imports not resolving? If so, a one-line note here would keep the next editor from "fixing" it.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+  - `PLAN.md` — the task goal, approach, steps, and definition of done.
+  - `SUMMARY.md` — the outcome, files changed, and status (written after execution).
+- **Action:** A call to a specialist subagent (via the `Agent` tool) to perform the coding.
+
+---
+
+# INTERACTION
+
+- Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+- An unanswered or skipped question is never a stop signal — in an unattended run, fall back to the documented default and continue, including writing the task's `PLAN.md` and `SUMMARY.md`. The default depth is "Plan and execute"; the default commit choice is "Don't commit".
```
AlexanderMakarov, line 43: This sentence deliberately mirrors the deliverable-command contract the linter enforces ("never a stop signal" + "including writing"), but `quick.md` isn't added to the `deliverableCommands` array in `tests/lint-prompts.test.js` (~line 421) — so the contract is unenforced and free to drift. Adding `'quick.md'` there passes immediately. (The other new contracts worth pinning are in the summary.)

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+
+---
+
+## RUN (default)
+
+### Step 1: Restate the Task and Choose Depth
+
+1.  Restate the task in one or two sentences so the user can confirm you understood it: "Quick task: [restatement]."
+2.  Ask how thorough the run should be with `AskUserQuestion` (header `Depth`, single-select):
+    - **Plan and execute** (Recommended) — go straight to a brief plan, then delegate.
```
AlexanderMakarov, line 70: I'd drop "(Recommended)" — flow.md road-tested exactly this and retired it ("a default is just a default", flow.md:41, and its history explicitly stopped pre-marking the lowest-friction option). No other command tags an option this way, and INTERACTION above already names "Plan and execute" as the unattended default, so the label adds steering without adding information.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+
+You do not write or edit code yourself. Construct a delegation prompt that includes:
+
+- The full `PLAN.md` content (goal, approach, steps, definition of done).
+- Any discussion decisions and research findings.
+- A `<project_context>` block (from Step 6.5): the list of existing context files with the instruction to read them before starting and align changes with the product definition and architecture.
+- A `<scope_discipline>` block: "Only make changes the task requires. Don't add features, refactor unrelated code, or add validation for scenarios outside the task. If something is unclear, ask rather than guessing."
+- An `<investigate_before_answering>` block: "Don't speculate about code you haven't opened. Read relevant files before editing. Issue independent reads in parallel."
+- A `<use_available_skills>` block: "Apply any skills declared in your frontmatter `skills:` list, and any project, user, or plugin skills whose description matches this work."
+- A `<completion_evidence>` block: "Before reporting done, run the verification commands from the Definition of Done and cite their fresh output. Do not claim success from belief — show the evidence."
```
AlexanderMakarov, line 160: These two blocks drifted from `implement.md`'s current versions while their neighbors are verbatim copies: `<use_available_skills>` drops the second sentence ("Skills carry project-specific patterns — they should shape your implementation."), and `<completion_evidence>` omits the conditional RED-validation instruction from PR #152 (when the task writes a test: revert → watch it fail → restore → watch it pass, quote both outputs). Since the Definition of Done here explicitly anticipates tests, mirroring `implement.md:70-71` keeps the two delegation paths held to the same standard.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+
+---
+
+## LIST
+
+1.  List task directories: `ls -d context/quick/*/ 2>/dev/null`. If none, print `No quick tasks found.` and stop.
+2.  For each directory, derive:
+    - **slug** — the directory name with the `YYYYMMDD-` prefix stripped. Before displaying, sanitize: strip non-printable characters and path separators. Never interpolate a raw directory name into a shell command.
+    - **date** — from the `YYYYMMDD-` prefix in the directory name.
+    - **status** (determined by comparing the `YYYYMMDD` prefix to today's date):
```
AlexanderMakarov, line 217: The lead-in says status is "determined by comparing the `YYYYMMDD` prefix to today's date", but the first two branches read `SUMMARY.md` and never look at the date — only the two missing-SUMMARY branches do. Worth rewording to "from `SUMMARY.md` when present, else by age of the date prefix" so a model following the parenthetical doesn't date-classify completed tasks.

CLAUDE.md
```diff
@@ -102,7 +102,9 @@ The canonical flow (each command is a markdown prompt under `commands/`):
 
 The first four are run once at project setup; the last five iterate per feature. Each command reads/writes a specific document under `context/` (e.g. `context/product/product-definition.md`, `context/spec/NNN-feature/tasks.md`). The numeric prefix on spec directories is allocated by `scripts/create-spec-directory.sh`.
 
-**Implementation delegation rule:** `/awos:implement` is an orchestrator only — it reads `tasks.md`, extracts the `**[Agent: name]**` marker from each task, and delegates to a subagent. The orchestrator is explicitly prohibited from editing code itself. Preserve this contract when editing `commands/implement.md`.
+**Quick-task shortcut:** `/awos:quick` sits alongside the canonical flow for contained, one-off work that doesn't justify a full spec. It plans a single task, delegates to a specialist subagent, and writes `PLAN.md`/`SUMMARY.md` under `context/quick/NNNNNNNN-slug/` — separate from `context/spec/` and never tracked in the roadmap. It shares `/awos:implement`'s orchestrator-only contract (delegates all code changes) and offers `list`/`status`/`resume` subcommands.
```
AlexanderMakarov, line 105: `NNNNNNNN-slug` reads as an eight-digit sequence number — especially two paragraphs after the spec `NNN-` prefix convention — but it's a date. `context/quick/YYYYMMDD-slug/`, as the other docs write it, avoids the misread.

README.md
```diff
@@ -53,6 +53,17 @@ Once your foundation is set, iterate through this cycle for each feature on your
 
 > **When to skip the cycle**: Not every change needs a spec. Hotfixes, simple bugfixes, and small edits don't require the full spec workflow — Claude Code's built-in plan mode handles those just fine.
 
+### Quick Tasks
+
+For contained, one-off work that needs a plan and a specialist but not a full spec — a focused bugfix, a small refactor, a config or dependency change — use the shorter path:
```
AlexanderMakarov, line 58: The tip a few lines up still says "Hotfixes, simple bugfixes, and small edits" belong to built-in plan mode, and this section now routes "a focused bugfix" here — opposite advice in adjacent paragraphs. `commands/quick.md`'s WHEN TO USE already draws the three-way line (trivial one-step edit → plan mode; needs a plan and a specialist → quick; roadmap feature → full cycle); narrowing the tip to that framing would make the README agree with itself.

docs/commands/quick.md
```diff
@@ -0,0 +1,86 @@
...
+This command is a shorter path for one-off work that doesn't justify a full `spec → tech → tasks → implement → verify` cycle. Like `/awos:implement`, it acts as an orchestrator — it plans the task, delegates the coding to a specialist subagent, and tracks the outcome. It does **not** write code itself.
+
+Quick tasks live in their own space, separate from planned features:
+
+- `context/quick/[YYYYMMDD]-[slug]/PLAN.md` — the task goal, approach, steps, and definition of done.
+- `context/quick/[YYYYMMDD]-[slug]/SUMMARY.md` — the outcome, files changed, and status (written after execution).
+
+Nothing in `context/spec/` or the roadmap is touched.
+
+## When to use it
```
AlexanderMakarov, line 18: All nine existing `docs/commands/*.md` pages carry `## Prerequisites` right after "What it does". It's genuinely informative here too — quick is the one command that works with zero prerequisites, using the context files only when they exist.

commands/quick.md
```diff
@@ -0,0 +1,246 @@
...
+
+# PROCESS
+
+## Step 0: Parse Arguments and Route
+
+Inspect `<user_prompt>` and route:
+
+- Starts with `list` → go to **LIST**.
+- Starts with `status ` → the remainder is a slug → go to **STATUS**.
+- Starts with `resume ` → the remainder is a slug → go to **RESUME**.
```
AlexanderMakarov, line 56: These prefixes have no guard: `/awos:quick resume the paused ingestion job` routes to RESUME, slug-sanitizes the sentence into garbage, and stops with "No quick task found" instead of running the task; `list stale feature flags and remove them` likewise disappears into LIST. A model will often save this with common sense, but a cheap guard exists: only route to STATUS/RESUME when the remainder already matches the slug rules, and treat `list` as reserved only when it's the entire prompt — otherwise fall through to RUN.

AlexanderMakarov (changes requested):
Before the inline details — I'd push back on the premise itself, on three grounds.

First, this grows the core surface. A new core command is cognitive load for every AWOS user — the README now offers built-in plan mode, `/awos:quick`, and the full cycle for overlapping classes of work, with the choosing burden on the user — and that cost needs a proportionally strong reason to exist.

Second, the repo already has an answer aimed at exactly this middle: `/awos:flow` generates a project-tailored `/fix-bug` command (diagnose → fix → scoped re-verify → **targeted spec amendment**) — lighter than the spec cycle, delegating to the same hired specialists, and, unlike `/awos:quick`, ending by amending the specs the change touched. The PR's rationale (full ceremony vs bypassing AWOS is an awkward choice) doesn't engage with that existing path; I'd want it to argue why the flow-generated `/fix-bug` doesn't cover this, and if a real gap remains (non-bug one-off tasks? projects that haven't run `/awos:flow`?), whether it's better filled by extending flow's output than by a parallel core command.

Third — and this is the deeper one — `/awos:quick` writes its `PLAN.md`/`SUMMARY.md` to `context/quick/`, deliberately outside the roadmap and specs. That means the changes it makes are invisible to the product context afterward: the spec documents no longer describe the code, and the SDD premise this whole framework rests on — an agent can rehydrate the project from the documents alone — erodes a little with every quick task. The flow-generated `/fix-bug` solved this with its spec-amendment step; `/awos:quick` has no equivalent, and "separate from the roadmap" is presented as a feature. Ad-hoc changes untied to product context are the thing spec-driven development exists to prevent, so if this command ships, it needs a story for how its changes flow back into the specs.

If there's a settled decision behind this I'm not seeing, point me to it and I'll take the premise objection back.

On the implementation itself (details inline):

- Step 6.5 checks a nonexistent path — quick tasks silently never see the architecture doc.
- The slug-lookup glob matches across slug boundaries — `status`/`resume` can act on the wrong task.
- Depth isn't persisted — a resumed "Full" run loses its validation pass.
- The `/awos:q` alias never binds `$ARGUMENTS`.
- The PR's new structural contracts (directory layout, `status:` frontmatter, delegation-block parity, the quick↔q relation) ship with no Layer-1 pins — nothing fails today, but nothing protects them either.

The rest are small consistency points. The CodeRabbit round is fully addressed — nothing from it re-raised here.

AlexanderMakarov:
@kdementiev-provectus - first of all let me thank you for the idea and a feedback provided by the PR itself.

I want to share with you https://github.com/provectus/awos/pull/185 - this is an alternative solution for the same problem directly in core commands. If you would be able to check/review this PR then it would be a good contribution to AWOS and I will appreciate it.

dashagaranina:
Closing this on philosophy grounds, not on craft — the depth question instead of flags, the orchestrator-only contract, and the unattended defaults are all careful work, and the annoyance is real: five commands for a twenty-minute fix was a bad deal. The philosophy that landed since (docs/philosophy.md in #194, principle 11) draws the line differently, though: awos is for work worth specifying, and for hotfixes and small edits the escape hatch is the host's own plan mode — not a second awos lane, which is the "two ways to do the same step" regression the direction doc names. context/quick/ compounds it by recording changes where no spec or product document will ever look, which is the erosion Alexander's review flagged and the same reason #122's parallel ledger closed. The kernel — small work shouldn't pay full spec ceremony — is being answered on the one path instead: roadmap Phase 6A collapses tasks and verify into implement, so the cycle becomes spec → tech → implement, and the README's "when to skip the cycle" note keeps plan mode as the documented boundary below that threshold. This is also where the #185 thread landed ("a shorter spec is still a spec"); closing here closes that loop.

PR #181, opened 2026-08-17
feat(better): experimental research-backed spec plugin (/better:spec)

kmakarychev-dev:
## What

Adds a second, experimental plugin `plugins/better/` — the `better` namespace for higher-rigor forks of core commands. The first command, `/better:spec`, forks the core `/awos:spec` flow and adds four phases:

1. **Research fan-out before drafting** — dispatched as parallel `Agent` calls in a single message: codebase (built-in `Explore`, always), web (`general-purpose` with WebSearch/WebFetch, always, self-reports `SKIPPED` without web access), and internal KB (gated on `context/sources/sources.md` existing with `## Status: configured` and a documentation source; recorded transports are passed into the agent prompt).
2. **Synthesis round** — origin-labeled findings (`[code]`/`[web]`/`[kb]`) become sharper interview questions (conflicts, discovered decisions) or fold directly into requirements and acceptance criteria; technical findings are translated to user-facing language per the Language Rules.
3. **Blind verification after the write** — the bundled `better:spec-verifier` agent (`tools: Read`, the repo's first tools-restricted agent) is dispatched with only the spec file path, reads it cold, and reports ambiguities and untestable criteria. Runs exactly once; fixes folded in, residuals become `[NEEDS CLARIFICATION]` markers.
4. **A human-facing review page** — the command's final step renders `functional-spec.html`: an at-a-glance summary, a decisions-to-approve table, per-lane research findings linked to the requirement each one shaped, and every requirement as a card with named, collapsible acceptance criteria and provenance badges (Interview / Codebase / Web / KB). The command authors a view-model in-session and passes it as an ephemeral temp file (never stored in the project) to `scripts/render-spec.mjs` — a no-dependency, syntax-driven Node renderer that extracts contract text verbatim from the markdown (composes, never rewrites), computes verification counts from checkboxes, and degrades to a generic render without a view-model. Unconditional but never fatal; the page is a stamped point-in-time snapshot. No separate render command by design.

5. **An optional share link** — after rendering, the command offers to publish the review page as a Claude artifact (a URL beats a file for reviewers; GitHub PRs can't render HTML). Gated on the `Artifact` tool being present in the session, and a strict consent gate: unanswered means no, unattended runs never publish (the hire.md withheld-consent posture). Renderer support: `--artifact` emits a skeleton-free fragment and the stylesheet honors the artifact viewer's `data-theme` toggle.

## Backward compatibility

The deliverable contract is deliberately identical to `/awos:spec`: same template, same `context/spec/NNN-name/functional-spec.md`, directory allocated by the same core script. `/awos:tech` and everything downstream work unchanged. `research-notes.md` and `functional-spec.html` are purely additive artifacts nothing depends on. **No core command is modified.** Update Mode is deliberately out of scope (amendment-shaped prompts route to `/awos:spec`).

## Registration & versioning

- One new entry in `.claude-plugin/marketplace.json` — the installer registers the marketplace as a whole, so **zero installer changes**.
- Independent version line starting at `0.1.0`; its discipline is three files moving together (marketplace entry, its `plugin.json`, the `EXPECTED_BETTER_PLUGIN_VERSION` lint pin).

## Tests

- New lint blocks: version-pin (three-file discipline), structural grep checks on the command (single-message fan-out, `AWOS_UNATTENDED` handling, marker discipline, core-contract references, sources gate idiom, verifier once/path-only, render step present and non-fatal), and verifier blindness (`tools: Read` + no-other-file prohibition).
- `tests/better-render.test.js`: renderer contracts — verbatim criteria extraction, checkbox-computed counts, view-model layering (one-liners, decisions, anchored findings, criteria names), and graceful no-view-model degradation.
- All local layers pass (203 markdown/installer/fixture/renderer, 1348 engine, 20 harness, 11 tools); `prettier --check` clean.
- Behavioral contracts (fan-out actually fires headless, verifier stays blind) belong in the awos-qa repo — follow-up.

## Field test

Run A/B against a real project (same roadmap item, core `/awos:spec` vs `/better:spec`): the researched spec caught three requirements the interview-only spec got wrong or untestable against the underlying tech's real behavior (an indefinite-wait trap after a laptop reboot, a silent-hang fallback path with no timing bound, and a UI recovery assertion the codebase contradicts). The renderer was validated end-to-end against the same project's spec (8 requirement cards, 28 criteria parsed with correct per-requirement counts). Known tune-up for a follow-up: technical test-setup details can leak into acceptance criteria; they belong in `research-notes.md`.

## Follow-ups (deliberately not in this PR)

- Root README exposure once the experiment graduates.
- Optional-input line in core `tech.md` pointing at `research-notes.md`.
- awos-qa behavioral coverage.
- CONTRIBUTING.md project tree omits `plugins/` entirely (pre-existing staleness).

## Summary by CodeRabbit

* **New Features**
  * Added the experimental `better` plugin and `/better:spec` command.
  * Generate research-backed functional specifications, research notes, and HTML review pages.
  * Supports unattended workflows, clarification markers, source-status reporting, optional artifact publishing, and compatibility with `/awos:tech`.
  * Added responsive light/dark review pages with requirements, findings, progress, and diagrams.
  * Added `/awos:flow` to the available plugin commands.

* **Documentation**
  * Updated plugin installation, usage, workflow, output, and experimental-status guidance.

* **Tests**
  * Added coverage for specification workflows, rendering, validation, and safe output handling.


plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,615 @@
...
+    /requirement/i.test(section.title) ||
+    section.subs.some((sub) => parseBody(sub.body).criteria.length > 0);
+
+  const requirementSubs = [];
+  const genericSections = [];
+  let scopeSection = null;
+  for (const section of spec.sections) {
+    if (/scope/i.test(section.title) && section.subs.length) {
+      scopeSection = section;
+    } else if (isRequirementSection(section)) {
```
AlexanderMakarov, line 732: The review page loses its requirements for every spec written in the template's own format.

`isRequirementSection()` matches on the heading text, so `## 2. Functional Requirements (The "What")` always enters this branch. The branch then pushes `section.subs` — but `templates/functional-spec-template.md` puts requirements as flat bullets with a nested `**Acceptance Criteria:**` checklist, with no `###` subsections at all. Neither core `spec.md` nor this command's Step 6 tells the model to add them. So `subs` is empty, nothing is pushed, and because the `else if` already consumed the section it never falls through to `genericSections` either.

I ran the shipped renderer against a fixture in the template's shape. The output HTML contains Overview, Scope and Boundaries, and Change Log — no requirements heading, no requirement text, neither acceptance criterion. The status bar still reads "1 of 2 criteria verified", because the count comes from a separate pass over `spec.sections`. The script exits 0, so Step 10 reads it as a clean render and Step 10.3 will offer to publish that page.

Guarding the branch on having subsections fixes it — `else if (isRequirementSection(section) && section.subs.length)` sends the section to `renderGenericSection`, and I confirmed the requirement and both criteria come back. Worth pairing with a fixture built from the real template.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Step 5: Synthesis — Interview Round Two
+
+1. Merge the findings, keeping their origin labels (`[code]` / `[web]` / `[kb]`). Set aside the raw technical form of every finding for `research-notes.md` (Step 7).
+2. Sort the findings into three piles:
+   - **Conflicts** — a finding contradicts the user's answers, the roadmap, or another finding. Each conflict becomes an `AskUserQuestion` option set: present the alternatives and let the user decide.
+   - **Discovered decisions** — a finding surfaces a choice the user has not made (a convention comparable products follow, an edge case the codebase already handles a particular way). Ask via `AskUserQuestion`, offering the finding-backed option as the recommended first choice.
+   - **Uncontroversial detail** — edge cases, boundary behavior, and error paths no reasonable user would dispute. Fold these directly into the draft as requirements and acceptance criteria; do not spend a question on them.
+3. Batch the questions — up to four per `AskUserQuestion` call, at most two calls in this round. Anything unresolved beyond that budget, and any question that times out, becomes a `[NEEDS CLARIFICATION: …]` marker in the draft.
```
AlexanderMakarov: Step 5 can cost you both deliverables, and line 52 of this same file says it shouldn't.

The conflicts and discovered-decisions piles each dispatch a real `AskUserQuestion`, and Step 7 — where `functional-spec.md` and `research-notes.md` are actually written — comes after. Under `claude -p` a dismissed question ends the turn, so nothing after it runs and both files are lost. That is exactly the failure the editor note in core `commands/spec.md:43` describes, and its stated fix is structural: keep the write ahead of any dismissable question, then refine afterward.

The `AWOS_UNATTENDED` branch on line 51 doesn't cover this. The instruction to "take the marker path immediately and continue" is itself an instruction after the dismissed question, so it never executes.

The shape that works is the one you already use in Step 9: let Step 5's conflicts and decisions become `[NEEDS CLARIFICATION: …]` markers during drafting, and move the interactive asks to after the write.

CLAUDE.md
```diff
@@ -122,6 +122,10 @@ Each audit dimension is a standalone `.md` file in `plugins/awos/skills/ai-readi
 
 The repo has **two independent version lines**. The npm installer version is managed by release-drafter via PR labels — never bump it manually. The **plugin version** is manual: when plugin behavior changes, bump it as one deliberate commit moving four files together — `.claude-plugin/marketplace.json`, `plugins/awos/.claude-plugin/plugin.json`, the `EXPECTED_PLUGIN_VERSION` pin in `tests/lint-prompts.test.js`, and the generator-version constant in `plugins/awos/commands/flow.md` (the lint exists to force exactly this four-file discipline). `flow.md` carries its own copy of the version because it stamps the generator version into every generated artifact's footer and uses it to detect stale re-runs. The engine stamps the plugin version into `audit.json` provenance and the report footer, so a behavior change shipped without a bump mislabels its audits.
 
+## The Better Plugin (experimental)
```
AlexanderMakarov: This section lands between `## The Audit Plugin` and its own subsections, so it adopts them.

`### Measurement engine (TypeScript)` (129) and `### Running the engine tests` (150) document the audit engine, and they now sit under the better plugin's heading. Anything reading CLAUDE.md by heading — a person or an agent — comes away with `npm run build:audit-engine`, the committed `dist/`, and the tree-sitter grammars attached to `better`. Moving this section to after line 161 fixes it.

Two smaller things in the same paragraph:

1. `research-notes.md` is described as "raw findings for the tech phase". Nothing reads it — `commands/tech.md` takes `functional-spec.md` and `architecture.md`, and this PR modifies no core command. `plugins/better/README.md` hedges this correctly with "can draw on it"; CLAUDE.md states it as fact.
2. The KB gate is given as `## Status: configured` alone, but Step 4.3 also requires at least one source with `Category: documentation`. A project configured with only tickets or chat sources won't dispatch the lane. Both READMEs get this right — only CLAUDE.md is loose.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,184 @@
...
+2. **Functional Requirements (The "What"):** capture what the user can do, including the boundary and error behavior surfaced by research — what error message appears, what limits exist, what happens when the action fails. Mark every unresolved detail with `[NEEDS CLARIFICATION: your specific question]` directly in the draft.
+3. **Acceptance Criteria:** every requirement gets at least one criterion in the When/Then shape — the words **when** and **then** both appear, in that order, within a single sentence (Given optional, only when the precondition affects the outcome). Requirements with boundary or error behavior get at least one failure-path criterion.
+4. **Scope and Boundaries:** ask the user what to exclude within the topic; add all other roadmap items to Out-of-Scope automatically and say you did so.
+
+Then self-review the full draft end to end: replace any developer-facing language that slipped in; make vague or unmeasurable wording concrete in user-perceivable terms or convert it to a `[NEEDS CLARIFICATION: …]` marker; confirm every requirement carries at least one criterion with `when` and `then` in a single sentence.
```
AlexanderMakarov: This self-review check passes a requirement whose other criteria are non-compliant.

"At least one criterion with `when` and `then`" is satisfied by the first good bullet, and the rest go unread. Core `spec.md:148` carries the stronger form — "re-read the acceptance criteria one bullet at a time — not the set as a whole" — with the commit that added it saying why: checking the set as a whole is how a single offender survives.

The lint that pins that phrase reads `commands/spec.md` only, so nothing catches the fork losing it. Worth restoring the wording and adding the assertion to the new `better` block.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,184 @@
...
+2. Fold its findings back into the spec:
+   - Mechanical fixes — vague wording it flagged, a criterion missing its when/then pair, a term used before it is defined — apply directly.
+   - Judgment calls — a genuine ambiguity only the user can resolve — ask via `AskUserQuestion` (the files are already written, so an unanswered question just leaves a `[NEEDS CLARIFICATION: …]` marker).
+3. Re-save the spec once. The verifier runs exactly once — do not re-dispatch it after folding fixes; findings that remain open become markers or reported residuals.
+
+### Step 9: Final Review
+
+Present the saved specification. Resolve each remaining `[NEEDS CLARIFICATION: …]` marker with the user via `AskUserQuestion`, offering the assumption you would otherwise make as the recommended first option; fold each answer back into the relevant requirement and its acceptance criteria, then re-save. If no answer comes (an unattended run), leave the markers in place; the user — or `/awos:tech` — can resolve them later.
```
AlexanderMakarov: Open-ended markers get forced into a multiple choice here.

Core `spec.md:158` has "(with a free-text option for open-ended markers)" in this sentence and the fork dropped it. Not every `[NEEDS CLARIFICATION]` reduces to an option set, and line 49 of this file already warns against inventing filler options — so without the clause the two rules push in opposite directions. Restoring the parenthetical is the whole fix.

tests/lint-prompts.test.js
```diff
@@ -3430,6 +3430,128 @@ test(`plugin.json version matches the awos marketplace entry and equals ${EXPECT
...
+    [
+      '`## Status: configured`',
+      'the internal-KB lane must gate on the exact sources.md status idiom used across AWOS commands',
+    ],
+    [
+      'The verifier runs exactly once',
+      'the blind-verification cycle must be bounded to a single verifier dispatch',
+    ],
+    [
+      'render-spec.mjs',
```
AlexanderMakarov: Two needles in this block are loose enough to miss the failures they're there for.

1. Nothing pins `Agent(subagent_type="better:spec-verifier")`. Plugin agents only resolve through the `plugin-name:` prefix, so dropping or mistyping it means the verifier silently never runs — and the blind verification pass is the headline feature. The suite already pins `subagent_type="Explore"` verbatim for the same reason (`tests/lint-prompts.test.js:2154`).
2. This needle is the bare string `render-spec.mjs`, not `${CLAUDE_PLUGIN_ROOT}/scripts/render-spec.mjs`. The `flow.md` block pins the full `${CLAUDE_PLUGIN_ROOT}/…` form (`1455-1456`) precisely because a relative path breaks across install locations.

plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,615 @@
+#!/usr/bin/env node
+/**
+ * render-spec.mjs — deterministic review-page renderer for /better:spec.
+ *
+ * Usage:
+ *   node render-spec.mjs <spec-dir> <view-model.json>
```
AlexanderMakarov: The usage line in the header doesn't match the one the script prints.

Here it reads `<spec-dir> <view-model.json>`, implying both are required. `main()`'s own usage string is `<spec-dir> [view-model.json] [--artifact <fragment-out>]` — the view-model is optional and `--artifact` is a real mode, both used from Step 10.3 and covered by tests.

The schema block just below has the same problem: it documents `diagram` as `{ svg, note }`, but `renderDiagram` also reads `d.title` and renders it as the section heading (`:347`). Since Step 10.1 tells the model to author the view-model "following the schema documented at the top of the renderer script", this header is the contract — a field missing from it doesn't exist as far as the author is concerned.

tests/better-render.test.js
```diff
@@ -0,0 +1,193 @@
...
+    html.includes('CSV only for now'),
+    'decision rows from the view-model must render in the Decisions table'
+  );
+  assert.ok(
+    html.includes('href="#r21"'),
+    'a finding whose anchor matches a rendered requirement must link to it'
+  );
+  assert.ok(
+    html.includes('Happy path downloads a file'),
+    'criteria micro-names from the view-model must label the criteria in order'
```
AlexanderMakarov: This assertion promises more than it checks.

The message says the names "must label the criteria in order", but the check is a single `includes` on the first name. The second name isn't asserted at all, so a reversal or an off-by-one in `criteria_names` passes. Asserting both and comparing their `indexOf` would close it — and given the index bug on the thread at `render-spec.mjs:200`, ordering in exactly this field is worth pinning.

plugins/README.md
```diff
@@ -18,6 +18,7 @@ To register the marketplace manually:
...
-| **awos** | `/awos:ai-readiness-audit` | Comprehensive AI readiness audit across 8 dimensions — security, architecture, documentation, AI tooling, and more. Produces a scored report with actionable recommendations. |
+| Plugin                    | Commands                                 | Description                                                                                                                                                                                                                            |
+| ------------------------- | ---------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **awos**                  | `/awos:ai-readiness-audit`, `/awos:flow` | Comprehensive AI readiness audit across 8 dimensions — security, architecture, documentation, AI tooling, and more — plus end-to-end delivery-flow setup tailored to the team.                                                         |
```
AlexanderMakarov: [nit] "8 dimensions" is stale — `standards.toml` lists 13, of which one is unscored, so 12. Pre-existing drift, but this row is rewritten in the diff so it's cheap to fix here.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,184 @@
...
+
+---
+
+# INTERACTION
+
+- Use the `AskUserQuestion` tool for multiple-choice questions instead of plain text or numbered lists.
+- The tool accepts two to four listed options per question — pair the natural answer with the genuinely different behavior, never with a "Yes — I'll type it" filler (free text already covers that). One question per axis; combinable answers use `multiSelect`.
+- A skipped or unanswered question is never a stop signal. Mark the unresolved detail with a `[NEEDS CLARIFICATION: …]` marker and continue through the remaining steps, including writing both output files.
+- Interactive and unattended runs handle a no-answer differently. Read whether this run is unattended from the `AWOS_UNATTENDED` environment variable. **Unattended (set):** take the marker path immediately and continue. **Interactive (unset):** a 60-second timeout usually means the user is thinking — re-ask the question once; if it times out again, proceed with the marker and name the assumption you would otherwise make.
```
AlexanderMakarov: This line copies the sentence that already drew a reviewer question on `flow.md`, without the half that answers it.

1. `flow.md` carries "a 60s timeout usually means the user is thinking or briefly away" — and immediately before it, the grounding: `AskUserQuestion` returns a `No response after 60s` result, which is the harness's guard so headless runs never hang, not a signal the command sets or can extend. This fork takes the first half and leaves the grounding behind, so the number arrives unexplained. That's the question @dustyo-O asked about the same sentence in #134 ([discussion r3559960303](https://github.com/provectus/awos/pull/134#discussion_r3559960303)) — worth carrying the answer across with the text.
2. The second objection in that thread applies to this line word for word: "re-ask the question once — that doesn't seem reliable. When I do two projects in parallel, timeouts can be huge." That thread is still open, and this file inherits the behavior it questions. If the re-ask is deliberate here, it'd help to say why it holds for a spec interview when it was doubted for the flow interview.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,176 @@
...
+   - `requirements` — per requirement (matched by its number, e.g. `"2.1"`): a one-line plain-language statement, source badges, and a short scannable name for each acceptance criterion, in order.
+   - `diagram` (optional) — when the behavior is stateful or flow-heavy, an inline SVG overview with user-language labels, plus a note that the criteria remain the contract.
+
+   Before rendering, self-check the view-model against the saved spec: it must claim nothing the spec does not support — this page is a review lens, and a lens that embellishes misleads the person approving.
+
+2. Run the renderer and clean up:
+
+   ```bash
+   node "${CLAUDE_PLUGIN_ROOT}/scripts/render-spec.mjs" "context/spec/[index]-[short-name]" "$TMPDIR/spec-view-[index].json" && rm "$TMPDIR/spec-view-[index].json"
+   ```
```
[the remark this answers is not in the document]
AlexanderMakarov: Agreed, and it goes one step further than the read path. Line 179's *output* target `"$TMPDIR/spec-artifact-[index].html"` has the same missing fallback, and that write (`render-spec.mjs:610`) isn't guarded — so with `TMPDIR` unset the publish render throws rather than degrading, immediately after the user has said yes to publishing. The cleanup steps at 174 and 182 then remove paths that were never written. Resolving one variable and using it for the write, both renders, and the cleanup covers all of it.

plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,575 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Composition
+// ---------------------------------------------------------------------------
+
+function findVmRequirement(vm, heading) {
+  const reqs = (vm && vm.requirements) || [];
+  return reqs.find((r) => r.match && heading.includes(r.match)) || null;
+}
```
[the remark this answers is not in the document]
AlexanderMakarov: Confirmed by running the renderer: with requirements `2.1` and `2.10` and a distinct view-model entry for each, `2.10` rendered `2.1`'s one-liner *and* `2.1`'s source badge.

The badge case is worth calling out in the fix. A mis-attributed provenance badge is worse than a mis-attributed one-liner, because it's the field a reviewer uses to decide how much to trust the requirement in front of them.

plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,575 @@
...
+${panels}
+</div>`;
+}
+
+function renderDiagram(vm) {
+  const d = vm && vm.diagram;
+  if (!d || !d.svg || !/^\s*<svg[\s>]/i.test(d.svg)) return '';
+  return `<h2 id="lifecycle">${inline(d.title || 'Overview diagram')}</h2>
+<div class="diagram-wrap">${d.svg}</div>
+${d.note ? `<div class="diagram-note">${inline(d.note)}</div>` : ''}`;
```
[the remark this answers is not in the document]
AlexanderMakarov: Confirmed by running it. An `<svg>` carrying `onload="alert(1)"` with a nested `<script>alert(2)</script>` passes the opening-tag gate, and both the handler and the script tag land verbatim in `functional-spec.html` and in the `--artifact` fragment — which is what gets published to claude.ai, where inline script runs.

Two things that raise the priority. The view-model is authored immediately after the Step 4 web lane has ingested arbitrary pages, so "LLM-authored" here means "influenced by whatever the web returned". And this repo already has the pattern for pinning it: `plugins/awos/skills/ai-readiness-audit/tests/render-escaping.test.ts` does the hostile-payload pass for the audit renderer. `tests/better-render.test.js` never exercises `renderDiagram`, `esc()`, or `inline()` with hostile input at all.

plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,615 @@
...
+  }
+
+  const out = render(md, vm, pluginVersion);
+  const outPath = path.join(specDir, 'functional-spec.html');
+  fs.writeFileSync(outPath, out.page);
+  console.log(outPath);
+  if (artifactPath) {
+    fs.writeFileSync(artifactPath, out.fragment);
+    console.log(artifactPath);
+  }
```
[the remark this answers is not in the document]
AlexanderMakarov: Confirmed with three shapes against the shipped script: `{"at_a_glance":"str"}` exits 1 at `:496`, `{"findings":{"code":"oops"}}` exits 1 at `:318`, and `{"requirements":{}}` exits 1 on the same `:496` path. Normalizing every optional collection after the parse is the right fix.

One thing to add to it. Even when validation works, the degrade is invisible: the `catch` at 583-588 collapses a missing file, a permissions error, a truncated write, and a syntax error into one stderr line, then exits 0 with a real file at the reported path. Step 10 only checks "node is missing or the script errors", so it can't tell a stripped page from a complete one — and Step 10.3 will still offer to publish it. A greppable marker on stdout that Step 10 reads, and that suppresses the publish offer, would close that.

AlexanderMakarov (commented):
Thank you for the idea and quite suitable implementation! I guess we should show this "alternative" to our folks and gather feedbacks. While I still consider it as temporal solution - see 1st point in "Architectural notes" below.

Two things need fixing before merge, both in the renderer.

The review page drops the Functional Requirements section entirely for specs written in the shape `templates/functional-spec-template.md` actually produces. I ran the shipped renderer against a template-shaped fixture: the requirement and both its acceptance criteria are absent from the HTML, while the status bar still reports "1 of 2 criteria verified" — so the page counts criteria it never shows. The render exits 0, which means the command treats it as a success and can go on to offer publishing it as the shareable artifact. That is the review page's whole job, and it fails on the default path.

The other is the raw `d.svg` injection CodeRabbit flagged. It executes: an `<svg>` carrying `onload=` and a nested `<script>` passes the opening-tag gate and lands verbatim in both the page and the published fragment. What makes it worth blocking on rather than deferring is where the view-model comes from — the model authors it right after the Step 4 web lane has ingested arbitrary pages.

The rest are smaller. Three of them are places where the fork quietly dropped a hardening that core `spec.md` carries, which is the pattern I'd most want to get ahead of before a second `better` command forks anything.

## Architectural notes

**The fork inherits core's contracts by copy, and three have already drifted.** Step 6 lost the per-bullet `when`/`then` re-check, Step 9 lost the free-text-option clause, and Step 5 breaks the write-before-any-dismissable-question rule that this file's own line 52 restates. Each is individually small; together they're a maintenance shape worth naming now, because the drift is invisible — the existing lint that pins these phrases reads `commands/spec.md` only, so the fork can lose any of them silently. https://github.com/provectus/awos-qa may help with it but better to end up with the single implementation.

**The test fixture invents a spec shape the template never produces.** `SAMPLE_MD` uses `### 2.1.` per-requirement subsections; the real template emits flat bullets. That single mismatch is why the requirements-dropping bug got through a test suite that otherwise covers this renderer well. A fixture derived from `templates/functional-spec-template.md` would have caught it, and would keep catching it as the template evolves.

plugins/better/scripts/render-spec.mjs
kmakarychev-dev: Fixed in ebd3fa0. Guarded the branch as you suggested — `isRequirementSection(section) && section.subs.length` — so a template-shaped section falls through to `renderGenericSection` with its prose and both criteria intact.

Also took the fixture point, which was the more useful half: `tests/better-render.test.js` now carries `TEMPLATE_SHAPED_MD`, built from `templates/functional-spec-template.md`'s flat-bullet shape, alongside the existing subsection fixture. The new test asserts the requirement text, both criteria, and that the status-bar count and the rendered list describe the same criteria — the mismatch was the tell that the section had been dropped.

plugins/better/scripts/render-spec.mjs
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in 33cb2f9, by removing the raw field rather than sanitizing it.

`diagram` is now declarative: `items` of `box`/`arrow`/`text`, and the renderer draws the SVG itself. Geometry goes through a numeric coercion (a non-numeric coordinate drops that item), every label is escaped via `esc()`, and the legacy `svg` field is ignored outright — not parsed, not filtered. The opening-tag check is gone.

Added the hostile-payload pass `tests/better-render.test.js` was missing, modelled on `render-escaping.test.ts`: a view-model carrying `<script>`, `onload=`, an ``, a `javascript:` coordinate and a legacy raw-`svg` payload, asserted against both `functional-spec.html` and the `--artifact` fragment. It also asserts the escaped form is present, so the test fails if a field ever stops being rendered at all rather than passing by absence.

plugins/better/commands/spec.md
kmakarychev-dev: Deferred deliberately, and I'd like your read before changing it.

You're right about the mechanism — a dismissed `AskUserQuestion` in Step 5 ends the turn before Step 7 writes anything, the `AWOS_UNATTENDED` branch can't save it because that instruction is itself downstream of the dismissal, and line 52 of the same file says the write comes first. No argument on any of that.

I held it back because it's the one comment that changes the flow rather than patching it: moving the conflicts and discovered-decisions asks after the write means Step 6 drafts from markers instead of answers, which changes what the spec looks like on a normal interactive run, not just an unattended one. That felt like it belonged in a commit of its own with your sign-off, next to the fork-drift question below. The three mechanical fixes went in without it (64db412).

If you're happy with the Step 9 shape applied to Step 5 — conflicts and decisions become `[NEEDS CLARIFICATION: …]` during drafting, interactive asks move after Step 7 — I'll do exactly that.

CLAUDE.md
kmakarychev-dev: Fixed in 64db412, all three.

The section now sits after `### Running the engine tests`, so `### Measurement engine (TypeScript)` and the engine-test instructions stay under `## The Audit Plugin` where they belong.

`research-notes.md` now reads "raw findings the tech phase can draw on — nothing reads it automatically", matching the README's hedge. The KB gate now carries the documentation-source condition (and, per CodeRabbit's thread, the non-manual requirement that came with it).

plugins/better/commands/spec.md
kmakarychev-dev: Fixed in 64db412. Restored core's wording verbatim — "re-read the acceptance criteria one bullet at a time — not the set as a whole" — with the reason attached, and added `'one bullet at a time'` to the better block's needle list so the fork can't lose it silently again.

Your framing of this as the general problem is the part I'd want to keep discussing; replied on that under the review body.

plugins/better/commands/spec.md
kmakarychev-dev: Fixed in 64db412. Restored the parenthetical, and spelled out the tension you named so the two rules stop pulling against each other: "(with a free-text option for open-ended markers — not every marker reduces to an option set, and inventing filler options is its own failure)". Pinned as a needle too.

tests/lint-prompts.test.js
kmakarychev-dev: Both fixed in 64db412.

`Agent(subagent_type="better:spec-verifier")` is now pinned verbatim, with the failure it guards spelled out in the assertion message — without the prefix the agent doesn't resolve and the blind pass silently never runs.

The render needle is now the full `\${CLAUDE_PLUGIN_ROOT}/scripts/render-spec.mjs`, matching what the `flow.md` block does. The bare `render-spec.mjs` string would have passed on a relative path.

plugins/better/scripts/render-spec.mjs
kmakarychev-dev: Fixed in 64db412. The usage line now matches `main()` — `<spec-dir> [view-model.json] [--artifact <fragment-out>]` — with a sentence on what each optional part does.

The schema block is rewritten anyway, since `diagram` became declarative in 33cb2f9; `title` is documented there now along with `items`, `width`, `height`, and `note`. Your point about the header being the contract is why I documented it there rather than only in the command: Step 10.1 sends the author to that block and nowhere else.

tests/better-render.test.js
kmakarychev-dev: Fixed in 64db412. Now asserts both names are present and that `indexOf` of the first is less than the second, so a reversal or an off-by-one fails.

Worth pairing with the `2.10` fix on the other thread, which grew its own test asserting per-card one-liners *and* badges — between them, mislabeling within a card and mis-attribution across cards are both pinned.

plugins/README.md
kmakarychev-dev: Fixed in 64db412 — now reads "12 scored dimensions", the 13 in `dimension_order` minus `descriptors`.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,152 @@
...
+3. **Internal-KB lane (dispatched when configured).** Dispatch iff `context/sources/sources.md` exists with `## Status: configured` and lists at least one source with `Category: documentation`. Copy each such source's `Platform:`, `Access:`, `Tool:`, and `Scope:` lines into the prompt of one `Agent(subagent_type="general-purpose")` call, charged with: retrieve what the organization's knowledge base says about this topic — prior decisions, related designs, naming, constraints. Return a terse findings list, each item labeled `[kb]` with the source. If a recorded transport is unreachable at runtime, return `SKIPPED: <source> unreachable` for that source rather than failing. Sources with `Access: manual` are excluded from the agent — handle them in Step 5 via an `AskUserQuestion` invitation to paste relevant content (skip that invitation when `AWOS_UNATTENDED` is set). When `sources.md` is absent or not `configured`, note the lane as `NOT CONFIGURED` in the final summary and move on — never treat it as an error.
```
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in 64db412. The dispatch condition now requires at least one `Category: documentation` source whose `Access:` is not `manual`, with a sentence saying why — when every documentation source is manual there is nothing for the agent to retrieve. Manual sources keep going through the Step 5 `AskUserQuestion` invitation, unchanged.

plugins/better/commands/spec.md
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in 64db412, using the resolved-variable form both of you described.

Step 10.1 now resolves `view_model_path` and `artifact_path` once with `\${TMPDIR:-/tmp}`, and both renderer calls plus every cleanup use those variables. @AlexanderMakarov — your addition was the more serious half: the artifact *write* at `render-spec.mjs:610` is unguarded, so with `TMPDIR` unset the publish render threw immediately after the user consented, and the `rm` calls were removing paths that never existed. That path is now covered by the same two variables.

plugins/better/scripts/render-spec.mjs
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in 64db412. `findVmRequirement` now parses the leading number from the heading and compares it exactly, so `2.1` can no longer claim `2.10`'s card.

The regression test follows @AlexanderMakarov's note about badges being the worse half: it renders a spec containing both `2.1` and `2.10` with a distinct one-liner *and* a distinct source badge per entry, then asserts the `2.10` card carries its own of each. A reviewer reads the badge to decide how much to trust the requirement, so a mis-attributed badge is the failure worth pinning.

plugins/better/scripts/render-spec.mjs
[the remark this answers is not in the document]
kmakarychev-dev: Partly fixed in 64db412 — the crash is gone, the marker isn't in.

`normalizeViewModel()` now coerces every optional collection after the parse (`at_a_glance`, `decisions`, `requirements`, each findings lane, `criteria_names`, `sources`), so a wrong-typed field degrades its own block instead of failing the page. All three shapes @AlexanderMakarov reproduced are covered by a new test, plus `decisions: [null, 'nope']` and a string `criteria_names`.

I did not add the `VIEW_MODEL_STATUS=degraded` marker. It's a reasonable idea, but it changes the renderer's stdout contract and the command's publish gate together, and the publish gate is already being reworked on the Step 5 thread — I'd rather land those in one considered change than bolt a marker on now. Tracking it with the Step 5 discussion rather than closing it silently.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,184 @@
...
+
+3. Offer to publish the review page as a Claude artifact — a link is easier to share with reviewers than a file. Make the offer only when both hold: the render succeeded, and the `Artifact` tool is available in your toolset (introspection — when it is absent, skip this step silently rather than mentioning an option that cannot work). Ask via `AskUserQuestion`: publish as a private artifact, or skip. Publishing sends the spec's content to an external service, so this is a consent gate, not a preference: an unanswered or dismissed question means **no** — never publish by default, and never publish in an unattended run. This deliberately inverts this command's usual unanswered-question rule; outward-facing actions default to withheld consent. If the user opts in:
+
+   ```bash
+   node "${CLAUDE_PLUGIN_ROOT}/scripts/render-spec.mjs" "context/spec/[index]-[short-name]" "$TMPDIR/spec-view-[index].json" --artifact "$TMPDIR/spec-artifact-[index].html"
+   ```
+
+   Publish the fragment file with the `Artifact` tool (favicon `📋`), then delete both temp files. Report the artifact URL: it starts private, and the user shares it from claude.ai when ready.
```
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in 64db412, both halves.

Step 10.3 now does `rm -f "$artifact_path"` before rendering, and publishes only when the render succeeds and produced a file; on failure it reports, publishes nothing, and cleans up. Both temp files are removed either way.

Belt-and-braces on the renderer side too: `--artifact` now exits non-zero if its target resolves to `functional-spec.md` or `functional-spec.html`, with a test covering both collisions.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,152 @@
...
+2. Fold its findings back into the spec:
+   - Mechanical fixes — vague wording it flagged, a criterion missing its when/then pair, a term used before it is defined — apply directly.
+   - Judgment calls — a genuine ambiguity only the user can resolve — ask via `AskUserQuestion` (the files are already written, so an unanswered question just leaves a `[NEEDS CLARIFICATION: …]` marker).
+3. Re-save the spec once. The verifier runs exactly once — do not re-dispatch it after folding fixes; findings that remain open become markers or reported residuals.
+
+### Step 9: Final Review and Recommend Next Step
+
+1. Present the saved specification. Resolve each remaining `[NEEDS CLARIFICATION: …]` marker with the user via `AskUserQuestion`, offering the assumption you would otherwise make as the recommended first option; fold each answer back into the relevant requirement and its acceptance criteria, then re-save. If no answer comes (an unattended run), leave the markers in place; the user — or `/awos:tech` — can resolve them later.
```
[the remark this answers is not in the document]
kmakarychev-dev: Not changing this one — the ordering is deliberate, and the auto-marker above is wrong (41eff0a was the artifact-publish feature, not a fix for this).

Step 9 doesn't rewrite the spec. It resolves `[NEEDS CLARIFICATION: …]` markers with answers the user gives, then re-saves. Those markers exist precisely because nobody — the verifier included — could resolve them from the file alone, so a second verification pass wouldn't validate them either; it would re-read a document whose open questions had just been answered by the only party able to answer them. The exactly-once bound on the verifier is the contract that keeps the blind read blind and the run bounded.

The deterministic self-review you suggest as the alternative is already there: Step 6's per-bullet when/then re-check, restored from core in 64db412 and now lint-pinned, and Step 9 folds each answer back into the requirement *and* its acceptance criteria.

kmakarychev-dev:
@AlexanderMakarov — thanks for the review, and for running the renderer rather than reading it. Both blockers were real; three commits pushed:

| | |
|---|---|
| `ebd3fa0` | requirements dropped for template-shaped specs |
| `33cb2f9` | raw view-model SVG → declarative diagram schema |
| `64db412` | the batch: exact `2.x` matching, view-model normalization, `\$TMPDIR` resolution, publish-only-on-success, artifact-path collision guard, the two restored core contracts, the KB manual-source gate, tightened lint needles, docs |

209 markdown/installer/renderer tests pass, prettier clean. Per-thread replies are inline.

Two things I did **not** do, both by choice rather than oversight:

**Step 5's questions before the write.** Your diagnosis is right and I'm not disputing it. I left it out of this batch because it changes the flow's shape — drafting from markers instead of answers changes what an ordinary interactive run produces, not just an unattended one — and that deserves its own commit with your sign-off rather than riding along with mechanical fixes. If the Step 9 shape applied to Step 5 is what you want, say so on that thread and it's a small change.

**CodeRabbit's "verification boundary".** Declined as by-design; reasoning on that thread.

## On the fork drift

This is the note worth keeping. You found three drifted contracts; the mechanism you identified — the lint that pins those phrases reads `commands/spec.md` only — is the real finding, because it means the count of drifted contracts is unknowable rather than three. I've pinned the two restored ones in the better block, but that's treating instances.

The structural options, as I see them:

1. **Shared needle list.** Extract the core-contract phrases into one array both blocks assert against. Cheap, and it makes drift fail loudly the moment a fork drops a phrase. It does force every fork to carry every core contract verbatim, which is right for hardening rules and wrong for anything a fork deliberately diverges on (this fork drops Update Mode on purpose) — so it needs an explicit opt-out list, which is itself a place drift can hide.
2. **Fold back into core sooner.** Your "better to end up with the single implementation" — the fork exists to find out whether the research fan-out and blind verification are worth their cost, and the field test says they are. If the answer is already yes, the fork's remaining job is small and the drift window closes when it merges rather than being managed indefinitely.

I lean toward (1) now and (2) soon: the shared list is a couple of hours and pays off even if the fork merges next month, since the same problem returns the first time `/better:tech` forks anything. But the sequencing is your call as much as mine — if you'd rather not accumulate more hardening in a copy that's meant to be temporary, I'd take that as the argument for going straight to (2) and treating this PR as the last one that touches the fork in isolation.

Your fixture point is the same lesson one layer down, and I took it: `TEMPLATE_SHAPED_MD` is now derived from `templates/functional-spec-template.md`, so the renderer is tested against the shape the template actually emits — and will keep being tested against it as the template evolves.

plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,766 @@
...
+  return {
+    at_a_glance: arr(raw.at_a_glance),
+    decisions: arr(raw.decisions).filter((d) => d && typeof d === 'object'),
+    requirements: arr(raw.requirements).filter(
+      (r) => r && typeof r === 'object'
+    ),
+    findings: {
+      code: arr(f.code),
+      web: arr(f.web),
+      kb: arr(f.kb),
```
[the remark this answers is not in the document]
kmakarychev-dev: Confirmed and fixed in 863791d. Reproduced exactly as described — `{"findings":{"code":[null]}}` exited 1 at `renderFindings` and wrote no page at all.

This was my own incomplete fix: 64db412 normalized the *collections* but not their *entries*, which is the same bug one level down. `normalizeViewModel` now runs every collection whose entries get dereferenced — all three findings lanes, `decisions`, `requirements` — through one entry filter that keeps only objects.

Test extended past the payload you named: null and string entries across the findings lanes, `decisions`, and `requirements`, plus a non-array `sources`.

plugins/better/scripts/render-spec.mjs
```diff
@@ -0,0 +1,766 @@
...
+  // Declarative items only — the view-model never supplies markup (see the
+  // header note); anything but well-formed box/arrow/text items is skipped.
+  const d = vm && vm.diagram;
+  if (!d || !Array.isArray(d.items)) return '';
+  const body = d.items.map(renderDiagramItem).filter(Boolean).join('\n');
+  if (!body) return '';
+  const w = num(d.width, 860);
+  const h = num(d.height, 320);
+  return `<h2 id="lifecycle">${inline(d.title || 'Overview diagram')}</h2>
+<div class="diagram-wrap"><svg viewBox="0 0 ${w} ${h}" role="img" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
kmakarychev-dev: Confirmed and fixed in 863791d. `{"width":0,"height":-5}` produced `viewBox="0 0 0 -5"` — invalid, and the diagram silently did not draw despite valid items.

Dimensions now fall back to the default canvas unless the supplied value is positive. New test covers zero, negative, null, non-numeric, and absent dimensions, and also asserts valid positive dimensions are still honored — a fallback that swallowed good input would be the natural way to get this wrong.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,192 @@
...
+
+   ```bash
+   rm -f "$artifact_path"
+   node "${CLAUDE_PLUGIN_ROOT}/scripts/render-spec.mjs" "context/spec/[index]-[short-name]" "$view_model_path" --artifact "$artifact_path"
+   ```
+
+   Publish only a fragment this run produced: if that command fails or writes no file, report that the artifact render failed, publish nothing, and clean up. Otherwise publish the fragment file with the `Artifact` tool (favicon `📋`). Either way, delete both temp files afterwards (`rm -f "$view_model_path" "$artifact_path"`). Report the artifact URL: it starts private, and the user shares it from claude.ai when ready.
```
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in 863791d. You're right that the `rm -f` sat inside the opt-in branch, so declining the offer, an absent `Artifact` tool, an unattended run, or a failed render all left the view-model behind.

Cleanup is now its own unconditional Step 10.4 after the publish decision, with the paths it removes and the reason spelled out — the view-model holds the spec's content and shouldn't linger in shared temp.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,192 @@
...
+### Step 10: Render the Review Page and Report
+
+The spec content is now final for this run, so render the human-facing review page. This step is unconditional — every run produces the page — but its failure is never fatal: if `node` is missing (`command -v node`) or the script errors, report that the render was skipped and why, and continue to the report; the markdown remains canonical.
+
+1. Author the **view-model** — the re-shaped, human-friendly representation of what this session learned. Write it as JSON to a temporary file outside the project, following the schema documented at the top of the renderer script. Resolve the temp paths once, with a fallback for an unset `TMPDIR`, and reuse those exact paths for every write, render, and cleanup below:
+
+   ```bash
+   view_model_path="${TMPDIR:-/tmp}/spec-view-[index].json"
+   artifact_path="${TMPDIR:-/tmp}/spec-artifact-[index].html"
+   ```
```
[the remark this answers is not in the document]
kmakarychev-dev: Not taking this one, at least not in this form.

The collision needs two `/better:spec` runs, concurrent, at the same spec index, in different projects, on one machine. Spec indices are allocated per project by `create-spec-directory.sh`, so this is two people or two sessions on shared hardware hitting the same number in the same window. Real, but narrow.

What makes me hesitate on the fix rather than the diagnosis: `mktemp -d` in a prompt adds a shell step whose failure mode (unset variable, cleanup of a directory that was never created) lands in the same place the `\$TMPDIR` bug just did. The command is a prompt, not a script — every branch is a thing a model can get wrong. I'd rather carry a narrow, understood risk than a broader instruction with a new failure surface.

Reasonable to revisit if the flow ever writes more than two temp files, or if these get any longer-lived than one run. Leaving the thread open rather than closing it, in case @AlexanderMakarov reads the tradeoff differently.

plugins/better/commands/spec.md
kmakarychev-dev: Implemented in bcb8d64 — you were right, and going in I found the problem was bigger than Step 5.

**Why not the `AWOS_UNATTENDED` gate.** I checked before choosing: that variable is only ever *read*. Nothing sets it — it is documented purely as an operator prerequisite for `/awos:flow`-generated automation (`flow.md:113`, `delivery-flow-template.md:80`). So gating on it would protect only the runs someone had already configured correctly, which are the ones that were never at risk, while bare `claude -p` — the invocation we document — stayed broken.

**What landed.** Your shape, with one refinement. Step 5 no longer interviews: conflicts and discovered decisions are drafted on the research-backed answer and marked with the assumption *named in the marker* (`[NEEDS CLARIFICATION: <what you said> vs <what research found> — drafted on the second; confirm or correct.]`). The refinement matters for the case you would rightly worry about — a conflict can reshape a requirement, not just fill a gap — so the marker names both sides rather than quietly picking one. The questions move to Step 9, now ordered by what a wrong assumption costs (conflicts first), batched, and folded back into the requirement **and** its criteria.

The payoff is that an unattended run stops producing nothing and starts producing a complete, coherent spec with every assumption visible.

**Two more pre-write asks I found while doing it**, both the same hazard:

- Step 6 item 4 asked the user what to exclude from scope — before the write. Now derived from research and Round One, and marked.
- The Step 4 manual-KB invitation ("paste relevant content") was also pre-write. Moved to Step 9.

**Round One I kept**, since it is the interview and removing it would gut the interactive case. But it is now a single batched call — one dismissal opportunity instead of several — and is skipped outright when `AWOS_UNATTENDED` is set, drafting from the topic, roadmap, and product definition with markers for what the interview would have closed. So: exactly one dismissable question before the writes in the worst case, none in a configured unattended run. I would rather state that residual honestly than claim the hazard is gone.

Interactive runs are unchanged in substance — same questions, same research behind them, a few steps later, against a spec already on disk.

Pinned three needles so the fork cannot quietly lose this: `'This step asks the user nothing'`, the Round One skip, and `'Do not ask here: this step precedes the write'`.

One thing worth your call, tied to your single-implementation point: core `/awos:spec` has the same exposure — Step 3 interviews before the Step 5 write, and the editor note at `commands/spec.md:43` acknowledges the rule is "necessary but not sufficient" without the ordering being fixed. I left core alone deliberately (this PR touches no core command), but the fork now has the treatment core wants. Happy to port it in a separate PR, or leave it for the fold-back if that is closer.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,210 @@
...
... Dispatch iff `context/sources/sources.md` exists with `## Status: configured` and lists at least one source with `Category: documentation` whose `Access:` is not `manual`. When every documentation source is `Access: manual` there is nothing for an agent to retrieve — skip the dispatch and handle those sources through the Step 5 invitation below. Copy each such source's `Platform:`, `Access:`, `Tool:`, and `Scope:` lines into the prompt of one `Agent(subagent_type="general-purpose")` call, charged with: retrieve what the organization's knowledge base says about this topic — prior decisions, related designs, naming, constraints. Return a terse findings list, each item labeled `[kb]` with the source. If a recorded transport is unreachable at runtime, return `SKIPPED: <source> unreachable` for that source rather than failing. Sources with `Access: manual` are excluded from the agent — record them for Step 9, which invites the user to paste relevant content after the files are written; note in `research-notes.md` that they were not consulted. When `sources.md` is absent or not `configured`, note the lane as `NOT CONFIGURED` in the final summary and move on — never treat it as an error.
```
AlexanderMakarov: This sends manual-only KB sources to a step that no longer takes them. Line 101 says Step 5 asks the user nothing, and the invitation moved to Step 9.3 — which this same sentence gets right two clauses later ("record them for Step 9").

A project whose documentation sources are all `Access: manual` hits exactly this path, and gets two different answers about where the paste invitation lives. One word.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,210 @@
...
+- Interactive and unattended runs handle a no-answer differently. Read whether this run is unattended from the `AWOS_UNATTENDED` environment variable. **Unattended (set):** take the marker path immediately and continue. **Interactive (unset):** a 60-second timeout usually means the user is thinking — re-ask the question once; if it times out again, proceed with the marker and name the assumption you would otherwise make.
+- **Both markdown deliverables are written before any question that could end the run unanswered.** Steps 1–7 ask nothing that can end the turn: the only interview is Round One (Step 3), which establishes the topic and cannot be skipped past, and everything research raises becomes a `[NEEDS CLARIFICATION: …]` marker carrying its assumption. All remaining questions live in Step 9, where they refine files that already exist. Under `claude -p` a dismissed question ends the turn, so a deliverable written after one never happens — this ordering, not the marker rule above, is what makes an unattended run produce files. The HTML render (Step 10) runs last so the page reflects the final content.
```
AlexanderMakarov: Two claims on this line were overtaken by the rest of the same commit. "All remaining questions live in Step 9" — Step 8 asks its own at line 152. "Round One … cannot be skipped past" — line 86 skips it entirely under `AWOS_UNATTENDED`.

Neither changes the behavior, and the premise this line rests on is right. It matters because line 52 is the sentence someone editing Steps 8–10 will trust, and the `:152` thread is roughly what happens when it's trusted. "from Step 8 onward" and "asked whenever the run is interactive" would both do it. `CLAUDE.md:161` carries the first claim too.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,210 @@
...
+
+4. Delete the temp files, whatever the publish decision was:
+
+   ```bash
+   rm -f "$view_model_path" "$artifact_path"
+   ```
+
+   This runs on every path — offer declined, `Artifact` tool absent, unattended run, render failed, or published. The view-model holds the spec's content; it does not belong in shared temporary storage once the run is done.
+
+5. Report: all three saved paths, the artifact URL if one was published, any research lane that was `SKIPPED` or `NOT CONFIGURED`, any verifier findings left unresolved, and the next command: `/awos:tech`. Note that the HTML is a point-in-time snapshot stamped with its generation date — it does not update when the spec is later edited or verified.
```
AlexanderMakarov: There's no label here for a KB lane that's configured but manual-only. Line 97 created that outcome: `sources.md` is `configured`, there's a real `Category: documentation` source, but every one is `Access: manual`, so nothing is dispatched.

That lane can't self-report `SKIPPED`, because no agent ran to report it. `NOT CONFIGURED` is the other option and it's actively misleading — it tells the user to go run `/awos:configure-external-sources`, which they already did. A third label, something like `MANUAL SOURCES ONLY — invited in Step 9`, closes it.

tests/better-render.test.js
```diff
@@ -0,0 +1,451 @@
...
+    { requirements: {} },
+    { at_a_glance: 'not an array' },
+    { findings: { code: 'oops' } },
+    // Entries inside an otherwise well-formed collection, not just the
+    // collection itself — these are dereferenced by the renderers.
+    { findings: { code: [null] } },
+    { findings: { web: ['plain string'], kb: [null, { text: 'ok' }] } },
+    { decisions: [null, 'nope'] },
+    { requirements: [null] },
+    { requirements: [{ match: '2.1', criteria_names: 'not-an-array' }] },
```
AlexanderMakarov, line 400: This shape passes whether or not the guard it exists to pin is there.

Running the pre-fix renderer (`41eff0ad`) against it: `criteria_names: 'not-an-array'` doesn't throw, because `names[0]` and `names[1]` are ordinary string index reads. You get `<b>n</b>` and `<b>o</b>` as criteria labels, exit 0, and the page still contains "Export failed. Try again." — which is the loop's only assertion. So the one case covering `Array.isArray` at `render-spec.mjs:229-231` would survive a revert of it.

The other eight shapes do genuinely crash pre-fix and are caught properly. For this one, `assert.ok(!/<b>[a-z]<\/b>/.test(html), …)` or lifting it out of the shared loop would pin what it's here for.

tests/lint-prompts.test.js
```diff
@@ -3430,6 +3430,153 @@ test(`plugin.json version matches the awos marketplace entry and equals ${EXPECT
...
+    [
+      'one bullet at a time',
+      'the self-review must re-read acceptance criteria one bullet at a time (carried from core spec.md) — checking the set as a whole lets a single non-compliant bullet survive',
+    ],
+    [
+      'free-text option for open-ended markers',
+      'marker resolution must offer free text (carried from core spec.md) — not every [NEEDS CLARIFICATION] reduces to an option set',
+    ],
+    [
+      'functional-spec.html',
```
AlexanderMakarov, line 3536: `64db412`'s subject names four fixes — matching, temp paths, drift, docs. Three got pinned; temp paths didn't. There's no needle anywhere in this file for `TMPDIR`, `view_model_path`, or `artifact_path`.

That's the one of the four where a regression is silent. If the paths go back to being re-interpolated per step instead of resolved once, the file written and the file rendered can diverge, and the `rm -f` cleans a path that was never written — which was the original bug. A needle on `view_model_path="${TMPDIR:-/tmp}/spec-view-[index].json"` would hold it, in the same shape as the seven siblings in this block.

AlexanderMakarov (approved):
Everything from the last round is addressed. Thanks for the turnaround.

Two open CodeRabbit threads still hold, and I've replied on both. `:180` is the real one: per-requirement enrichment — one-liners, provenance badges, criteria names — never renders for flat-bullet specs. Your field test's 8 requirement cards say the model emits numbered subsections in practice, so it's narrower than it reads. `:152` costs `functional-spec.html`, but only headless, which isn't what this command is for — I'd fix it for awos-qa's sake, since a run that can stop before the render is hard to write a reliable scenario against.

Approving: the spec itself lands on every path, and what's left is a page, a report label and two test-hygiene items.

One carried-over note — the fork copies core's contracts and the lint reads only core `spec.md`, so drift stays invisible. Step 8's parenthetical is the second instance.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,192 @@
...
+   ```bash
+   view_model_path="${TMPDIR:-/tmp}/spec-view-[index].json"
+   artifact_path="${TMPDIR:-/tmp}/spec-artifact-[index].html"
+   ```
+
+   Its content, all in the spec's non-technical register:
+   - `at_a_glance` — the feature in about five plain sentences.
+   - `decisions` — every judgment call made in this spec (the things a reviewer might challenge): the decision, the choice, why, and its sources (`interview` / `code` / `web` / `kb`).
+   - `findings` — per research lane, what was discovered and what it changed in the spec (`impact`, with an `anchor` like `#r21` pointing at the requirement it shaped). A lane that did not run gets an explanatory note, not silence.
+   - `requirements` — per requirement (matched by its number, e.g. `"2.1"`): a one-line plain-language statement, source badges, and a short scannable name for each acceptance criterion, in order.
```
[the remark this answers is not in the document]
AlexanderMakarov: Running the head renderer against the suite's own `TEMPLATE_SHAPED_MD` with a complete view-model: the `one_liner`, both `criteria_names`, both `sources` badges, `class="req"` and `href="#r21"` are each absent — zero occurrences. Cause is the `section.subs.length` gate at `render-spec.mjs:603`; `requirementSubs` stays empty for flat bullets, so `renderRequirement` never runs and `knownAnchors` (`:620`) is empty, which drops the finding anchors to unlinked text.

Worth naming what this is: the surviving half of the bug `ebd3fa0` fixed. That guard correctly stopped the content being dropped, but it routes the flat-bullet shape to `renderGenericSection`, which has no view-model path — so the enrichment loss moved rather than closed, while `spec.md:180`'s authoring schema and `README.md:37`'s badge promise both still describe it working.

The correction: this bites less often than the analysis suggests, and the evidence is yours. The field test in the PR description reports 8 requirement cards — and a card (`<section class="req">`) is only ever produced from `requirementSubs`, so on a real spec the model did emit numbered `###` subsections. The template's flat-bullet example isn't what actually gets drafted.

That changes which fix I'd pick. Teaching the renderer to match flat bullets positionally handles a shape the model apparently doesn't produce; making Step 6 emit numbered `### N.M` subsections explicitly pins the shape it already produces by accident. The second is smaller, but it does diverge the fork from the core template the deliverable contract is pinned to — your call which side of that you'd rather be on. Either way `TEMPLATE_SHAPED_MD` should get an enrichment assertion, since today it pins only the content half, which is why this survived.

plugins/better/commands/spec.md
```diff
@@ -0,0 +1,210 @@
...
+   A lane that did not produce findings records `SKIPPED: <reason>` (web/kb) or `NOT CONFIGURED` (kb) under its heading. Technical language — file paths, component names, API shapes — is allowed and expected here.
+
+5. Write both files without waiting for approval — a spec is reversible (re-run this command or `/awos:spec` to revise), and open questions are already captured as `[NEEDS CLARIFICATION: …]` markers.
+
+### Step 8: Blind Verification
+
+1. Dispatch the verifier: `Agent(subagent_type="better:spec-verifier")` with a prompt containing **only** the absolute path of the saved `functional-spec.md` — no topic summary, no interview recap, no research findings. The agent's value is that it reads the spec cold; any context you pass contaminates the test.
+2. Fold its findings back into the spec:
+   - Mechanical fixes — vague wording it flagged, a criterion missing its when/then pair, a term used before it is defined — apply directly.
+   - Judgment calls — a genuine ambiguity only the user can resolve — ask via `AskUserQuestion` (the files are already written, so an unanswered question just leaves a `[NEEDS CLARIFICATION: …]` marker).
```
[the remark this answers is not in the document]
AlexanderMakarov: This holds, and I'd add the third site.

`:193-200` has the same exposure, and there it costs more than the page: a turn ending on the publish question also skips Step 10.4's `rm -f`, leaving the view-model JSON — which holds the spec's content — in shared temp storage. That's the thing `:208` explicitly says must not happen.

The specific line I'd change first is Step 8's parenthetical at `:152`: "the files are already written, so an unanswered question just leaves a `[NEEDS CLARIFICATION: …]` marker." The first half is true and the second isn't — if the turn ends there, nothing downstream runs, so no marker gets left and Step 10 never renders. That's the pre-fix assumption core's own editor note at `commands/spec.md:43` was written to warn against.

On the shape of the fix, there's a precedent worth copying rather than inventing: `/awos:tasks` in awos #136 wrote the file before the review and marked it `<!-- not-user-reviewed -->`. The equivalent here is rendering once right after Step 8's re-save, re-rendering after Step 9, and leaving 10.3 as publish-only. That does mean a page rendered twice, the first time from non-final content — the `<!-- not-user-reviewed -->` trick is what made that acceptable for tasks, and a stamp on the page saying the same would carry it here.

I'd steer away from extending Step 3's `AWOS_UNATTENDED` skip to cover these instead. Your argument against gating on it is right, and awos-qa hit the same wall from the other side — the `flow-adopts-sources-transports` scenario notes that `AWOS_UNATTENDED=1` alone "did not reliably cover it" and needed an answer map.

None of this is merge-blocking from where I sit — it only bites headless, which isn't what `/better:spec` is for. The reason I'd still fix it is testability: a run that can stop before the render is awkward to write a reliable awos-qa scenario against.

plugins/better/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Amendments to existing specs are out of scope by design — use `/awos:spec`, which handles in-place updates.
+
+## What it does differently
+
+1. **Research fan-out before drafting.** After a big-picture interview round, up to three agents run in parallel against the current understanding:
+   - **Codebase** (always) — adjacent features, current behavior, in-repo docs, contradictions.
+   - **Web** (always) — how comparable products solve the problem, domain conventions, UX patterns. Reports `SKIPPED` when web access is unavailable.
+   - **Internal KB** (when configured) — runs only when `context/sources/sources.md` exists with `## Status: configured` and lists a documentation source. That file is written by the `configure-external-sources` skill, which ships with the **awos plugin** (not this one and not AWOS core) — install it with `/plugin install awos@awos-marketplace`, then run its `configure-external-sources` skill (or let `/awos:product` trigger it on brownfield projects). Reports `SKIPPED` when a recorded transport is unreachable.
```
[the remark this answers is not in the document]
AlexanderMakarov: Agreed — the diff shows this is straightforward drift. `64db412` added the non-manual condition to `spec.md:97` and to `CLAUDE.md:161` ("with a non-manual `Category: documentation` source"), but left this line at "lists a documentation source" — even though the commit edited lines right around it. Two of three descriptions updated.

While you're there: that gate also creates a lane outcome neither the README nor `spec.md:210`'s report labels name — configured, real documentation source, nothing dispatched. Worth documenting the Step 9.3 paste fallback in both places rather than only the README.

plugins/better/commands/spec.md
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in e5f94ed, renderer-side. You were right that it was the surviving half of the bug, and the reproduction matched yours exactly — one_liner, criteria names, badges, `class="req"` and `href="#r21"` all at zero on a flat-bullet spec with a full view-model.

`splitFlatRequirements` now splits a flat-bullet section into the same per-requirement units the subsection shape yields, numbered positionally from the section heading (2.1, 2.2, …). That number is the join: a view-model entry matching `"2.1"` and a finding anchored to `#r21` resolve identically in both shapes, so nothing in the authoring schema or the command had to change. A section that yields no unit still falls through to the generic renderer — showing it unenriched beats dropping it, which is the guarantee ebd3fa0 added and I did not want to trade back.

On your fix-selection point: I went renderer-side rather than pinning `### N.M` in Step 6, because backward compatibility with `/awos:spec`'s template is the promise the whole plugin is built on — the deliverable contract is the reason this fork is allowed to exist. Teaching the renderer both shapes costs ~30 lines and keeps the fork honest about that; teaching the model one shape saves the lines but makes the fork's output diverge from the template it claims to fill. Your read that the model already emits subsections in practice is what made this safe to do without urgency, but it is not something I want the renderer to depend on.

`TEMPLATE_SHAPED_MD` now asserts the enrichment half too — card, one-liner, both criteria names, both badges, and the resolved anchor. I verified it fails when the split is reverted.

tests/better-render.test.js
kmakarychev-dev: Confirmed and fixed in e5f94ed. You're right about the mechanism: `criteria_names: 'not-an-array'` index-reads as characters, so it exits 0 pre-fix and the loop's only assertion still passes. I ran the pre-fix renderer to check rather than take it on faith.

Pinned by its visible effect instead — `assert.ok(!/<b>[a-z]<\/b>/.test(html))` — so single-letter criterion labels fail the test. Verified it fails with the `Array.isArray` guard reverted. The other eight shapes stay in the loop, where they do genuinely crash pre-fix.

tests/lint-prompts.test.js
kmakarychev-dev: Fixed in e5f94ed. Right that it was the silent one of the four — a regression there reintroduces the original bug rather than showing up as a visibly broken page.

Needle added on `view_model_path="\${TMPDIR:-/tmp}/spec-view-[index].json"`, in the same shape as the siblings, with the failure spelled out in the assertion message: re-interpolating a bare `\$TMPDIR` per step lets the file written and the file rendered diverge, and makes the cleanup `rm` a path that was never created.

plugins/better/commands/spec.md
kmakarychev-dev: Fixed in e5f94ed — one word, as you said. The sentence now says Step 9 in both places.

plugins/better/commands/spec.md
kmakarychev-dev: Both fixed in e5f94ed, and your reason for caring is the right one — line 52 is the sentence someone editing Steps 8–10 will trust, and the `:152` thread is what trusting it looked like.

Took your suggested wording: "Every question from Step 8 onward refines files that already exist" and "asked whenever the run is interactive and skipped outright when it is not". `CLAUDE.md:161` carried the first claim too and is updated to match.

plugins/better/commands/spec.md
kmakarychev-dev: Fixed in e5f94ed. Added `MANUAL SOURCES ONLY` as a third label, recorded where the lane is decided (Step 4.3) and listed in the Step 10 report alongside `SKIPPED` and `NOT CONFIGURED`, with the distinction spelled out: no agent ran to report it, and the user's setup was correct.

Pinned as a lint needle, and the plugin README now documents the manual-only fallback rather than only the gate.

plugins/better/README.md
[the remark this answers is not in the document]
kmakarychev-dev: Fixed in e5f94ed. Straightforward drift, as you say — the README now carries the non-manual condition, and documents where a manual-only source actually gets used (the Step 9 paste invitation) rather than leaving it as an unexplained gate. The missing lane label from your other thread is covered there too.

plugins/better/commands/spec.md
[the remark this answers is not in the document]
kmakarychev-dev: Not fixing in this PR — agreeing with your own bottom line rather than overriding it. You said it only bites headless, isn't merge-blocking, and that your reason for wanting it is testability against awos-qa. That last part is exactly why I'd rather it be its own change than a rider on a batch of review fixes.

Three things I did take from the thread now, in e5f94ed:

The false half of `:152` is gone. It read "the files are already written, so an unanswered question just leaves a marker" — you're right that the second clause is the pre-fix assumption core's editor note warns against. It now says both files are saved and survive whatever the answer is, then: do not assume the run continues past an unanswered question, so write any marker you intend to leave into the file *before* you ask, not after. That doesn't move the render, but it stops the command instructing a model to rely on a step that may never run.

I'm not touching the `AWOS_UNATTENDED` skip to paper over it — noted, and the awos-qa `flow-adopts-sources-transports` data point (the flag alone "did not reliably cover it", needed an answer map) is a better argument against than the one I made myself.

The render-twice shape is the right one and I'd like to implement it as you described — render after Step 8's re-save, re-render after Step 9, leave 10.3 publish-only, stamp the first page. The `/awos:tasks` `<!-- not-user-reviewed -->` precedent from awos #136 is the part that makes it acceptable rather than just "a page that might be stale", and I'd want to follow that stamp's wording rather than invent one. Two open questions I'd want your call on before writing it: whether the stamp belongs in the rendered HTML only or also as a marker in the markdown, and whether a page rendered from non-final content should be publishable at all or only after the Step 9 render.

Happy to do it as the next PR on this branch's heels, or fold it into the awos-qa work if you'd rather the scenario and the fix land together. Leaving this thread open either way — it's the one substantive item on the PR that isn't closed.

PR #199, opened 2026-09-07
Update README.md

spushkarev-provectus:
## Summary by CodeRabbit

* **Documentation**
  * Updated the README to describe `awos` as a spec-driven development framework built on Claude Code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # **Agentic Workflow Operating System**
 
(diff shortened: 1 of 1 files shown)


dashagaranina:
Thanks — this is a real improvement. "Built on Claude Code" is accurate, and the "autonomous engineering department" framing it replaces oversold what the framework does. The intro line is untouched by the README restructuring we have in flight on the cleaning branch (#200), so this merges cleanly with no conflict. The remaining "ensuring results that are production-ready" claim in the same paragraph has a similar problem, but we'll handle that in the cleanup work rather than hold this up.

dashagaranina (approved):
Accurate de-hype — matches the philosophy's 'what awos is not' (not a Claude Code distribution, not an autonomous engineering department claim). Approved.

